PR #128, opened 2024-05-13
Feat project not found

LeoFranklin015:
Added page not found routing to the projects which aren't available.


pages/project/[projectId]/index.tsx
```diff
@@ -125,7 +125,12 @@ export const NestedLayout = ({ children }: Props) => {
           }
           setProject(fetchedProject);
         } catch (error: any) {
-          console.log(error);
+          console.log(error.message);
+          if (
+            error.message === "AxiosError: Request failed with status code 422"
```
Comment (amaurymagalhaes) [inline pages/project/[projectId]/index.tsx:130]: On this, i think it's more appropriate you use `error?.include` with the code you are looking for, because Axios can change that message.

PR #130, opened 2024-05-14
Enlist / Delist the community admins

LeoFranklin015:
Added a feature to enlist and delist the community admins in /admin/communities page


amaurymagalhaes (changes requested):
Suggestions:
- Let some spacing around the admin addresses, like: padding of 20px
- Make the trash icon red and turn it clickable(cursor: pointer)
- address * -> Address *
- Community admin -> Community Admin
- It closed instantly after i added an admin
- It's not refreshing the table after i added the address as admin
- Right now, this page is only bring infos, doesn't realize any action. But your PR brings some actions to it, so it's important to have permissions there, otherwise, any admin will see these "Add/Remove admin", Mahesh is aware of it? My suggestion is to allow these two columns to contract owner only
- "Remove admin >" Button, turn it red and put the trash icon instead of ">"

PR #190, opened 2024-07-17
fixed admin listing

LeoFranklin015:
Need to add types properly, apart from that it's working.Kindly test it once again. I have tested with OP sepolia , it works ,and since we dont have endpoints in base sepolia , it doesnt work


components/Pages/Admin/CommunityAdmin.tsx
```diff
@@ -124,6 +117,7 @@ export default function CommunitiesToAdminPage() {
                     const matchingCommunityAdmin = communityAdmins.find(
                       (admin: any) => admin.id === community.uid
                     );
+                    console.log(matchingCommunityAdmin);
```
amaurymagalhaes, line 120: Just remove these console spams and will be good then.

PR #243, opened 2024-08-16
feat : community stats

LeoFranklin015:
added community stats in separate page 

/admin/communities/stats


amaurymagalhaes (changes requested):
Just need to adjust the responsiveness thing, after that, will be good.

amaurymagalhaes (approved):

PR #249, opened 2024-08-21
fix: Request Intro

LeoFranklin015:
Added Contstaraints before calling the BE


amaurymagalhaes:
LGTM
Gonna deploy after backend push

PR #253, opened 2024-08-26
feat : Added total projects and increased page limit

LeoFranklin015:


amaurymagalhaes (changes requested):
Also, take a look on what i asked you on discord. You are using unique on that page of pagination, so backend should return all uniques for that route.

PR #267, opened 2024-09-03
feat : added mint to top donors 

LeoFranklin015:
In gitcoin airdrop , 
- added mints only for top minters 
- Added s.no for the donors


fabianferno (changes requested):
image

The input field must be rounded like the text area above it 

Also can you move the text input right next to the file upload? - like this 
image

fabianferno (approved):

PR #268, opened 2024-09-03
feat: Group by programs on community grants page

fabianferno:


amaurymagalhaes (changes requested):
image
- Dark mode have to be fixed
- Filter by program label should be in one line instead of two

image
- If the grant program name is too big, will have some issues, maybe a dropdown would be better?

PR #299, opened 2024-09-11
Added telegram to program registry

LeoFranklin015:


amaurymagalhaes (changes requested):
image
There is couple places that telegram social media is missing, i approved a program so you can test it properly.

PR #313, opened 2024-09-13
feat: max 30 characters on titles

amaurymagalhaes:
- [x] Adding 30 characters as maximum on titles


fabianferno (changes requested):
image

All good except the "edit contact info on project" - it doesn't have that condition. I'm able to add more than 30 characters.

PR #328, opened 2024-09-24
fix: update typing of setActiveBadgeIds

FrancoAguzzi:


amaurymagalhaes:
As this code isn't really adding something useful, we are going to close for now

PR #388, opened 2025-01-22
Fix project_id redirects (swap uuid for slug) missing trailing paths

Madalosso:
Task ref: https://app.asana.com/0/1201924619794704/1209213260904246

> The attempt to access https://gap.karmahq.xyz/project/0xe3c6951bcc5d08b04a9150e51af31c7b16caf6e92c306a22c82a5bd6930a3760/grants immediately redirects to https://gap.karmahq.xyz/project/agroforestdao-1 (note that the /grants isn't carried over the redirect)


amaurymagalhaes:
Tested and seems properly fixed. Nice work!

amaurymagalhaes (changes requested):
I think using usePathname would be better in that case

PR #417, opened 2025-03-06
feat: Add communitySlug prop to IndicatorsHub component

fabianferno:
Update IndicatorsHub and ManageIndicatorsPage to use community slug instead of ID when rendering the IndicatorForm


amaurymagalhaes:
Dev is already sync with main.

PR #484, opened 2025-05-27
Display unique projects

maheshmurthy:


amaurymagalhaes:
I think it's better do it in the backend, can cause a mess in the frontend(like pagination, filtering...).

PR #503, opened 2025-06-11
Fix/refactor to remove redundant calls

maheshmurthy:
---
  - https://app.asana.com/0/0/1210563111331367


ealhaj (commented):
Thank you for the effort on this PR. However, the current solution doesn’t resolve the core issue it aims to address. In addition, it doesn’t contribute any meaningful improvements to state management, performance, or overall architecture.

Going forward, we’ve aligned as a team to use Zustand for state management, so we won’t be introducing Providers or React Context for this purpose.

Additionally, we should consistently use React Query for all API requests to maintain a unified data-fetching strategy across the codebase.

Finally, we’ll be adopting a standard folder structure for any newly implemented or refactored features to ensure consistency and better maintainability across the project.

ealhaj:
The folder structure is not implemented yet, I was discussing with Amaury about it, we are intending to refactor it and make it more clean and modular.

PR #510, opened 2025-06-17
Improve page load time

maheshmurthy:
---
  - https://app.asana.com/0/0/1210563111331374


amaurymagalhaes:
Unfortunately these changes does a lot of not related changes, like removing SEOs/dynamic metatags.
Also, the pre-fetch on hover is leading to some navigation issues, due overriding Nextjs native navigation with loader and setting states.

PR #588, opened 2025-08-13
Fix/total project count

AngeloAyranji:


components/Pages/Communities/Impact/StatCards.tsx
```diff
@@ -78,7 +80,7 @@ export const CommunityStatCards = () => {
   const stats = [
     {
       title: "Total Projects",
-      value: data?.totalProjects ? formatCurrency(data.totalProjects) : "-",
+      value: totalGrants ? formatCurrency(totalGrants) : "-",
```
amaurymagalhaes, line 83: This constant name can cause confusion

amaurymagalhaes (changes requested):

PR #631, opened 2025-09-01
fix: add program scores to community admin page

AngeloAyranji:


amaurymagalhaes (changes requested):
Asked you couple changes in BE and FE.
FE: Size limit. Maybe an example csv.
But looks good to me in general.

PR #634, opened 2025-09-03
Hadi/feat/contract verification

HadiKhai:


HadiKhai:
@amaurymagalhaes  Concerning point #6, the save needs to be done before verifying otherwise data would be lost because we need to save the contract before verifying it

HadiKhai:
@amaurymagalhaes fixed the other comments

HadiKhai:
> Amaury asked for a 2nd opinion here.
> 
> The dropdown on 1.) still looks a bit detached from the input, but it's not getting cut anymore so that's good image
> 
> About 7.), I looked up the back-end code and understood why it needed to be done this way, just as a suggestion it could be good to disable "Verify" and "Add another contract" buttons while this save is happening, maybe users dont realize that the "Save" button changed state and click again, causing some extra requests and saves
> 
> But for the rest, looks great!

This last push includes proper state handling for the "Save" button👍

PR #653, opened 2025-09-15
Infinite scroll of applications

maheshmurthy:
---
  - https://app.asana.com/0/0/1211356126940441


amaurymagalhaes:
Congrats! A very good piece of code.
We can use memoization on it, when we have large datasets.

PR #760, opened 2025-11-11
feat: Add staff access to merge projects

laHermes:
## Summary
Adds staff access to merge projects functionality.

## Changes
- Added staff permission check in MergeProjectDialog
- Added "Merge" button to staff section in ProjectOptionsMenu

## Backend Note
Staff merge currently uses direct on-chain attestation (user's wallet signs)

### DIFF
--- a/components/Dialogs/MergeProjectDialog.tsx
+++ b/components/Dialogs/MergeProjectDialog.tsx
@@ -33,6 +33,7 @@ import EthereumAddressToENSName from "../EthereumAddressToENSName";
 import { safeGetWalletClient } from "@/utilities/wallet-helpers";
 import { useWallet } from "@/hooks/useWallet";
 import { ensureCorrectChain } from "@/utilities/ensureCorrectChain";
+import { useStaff } from "@/hooks/useStaff";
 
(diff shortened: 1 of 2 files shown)


components/Dialogs/MergeProjectDialog.tsx
```diff
@@ -304,7 +306,7 @@ export const MergeProjectDialog: FC<MergeProjectProps> = ({
     <>
       {buttonElement ? (
         <Button
-          disabled={!isProjectAdmin}
+          disabled={!isProjectAdmin && !isStaff}
```
brunodmsi: The onClick calls a function that uses a existing on-chain attestation flow, is it tested that it will work with Staff wallets? Asking that because I noticed a change on the back-end for merging to allow staff, but this is not checked here

brunodmsi: Especially with how gap-indexer/app/modules/util/validate-attestation.ts works, validateCanAttestToRef does not check currently if isStaff

brunodmsi (approved):

PR #942, opened 2026-01-26
fix: prefer PostgreSQL UUID over MongoDB ID for indicators

brunodmsi:
## Overview

Updates indicator selection to use PostgreSQL `uuid` when available, ensuring custom indicators work correctly with impact segments.

## Problem

When adding custom indicators to impact segments on the manage-indicators page (`community/[communityId]/admin/manage-indicators`):
- The V2 API returns PostgreSQL UUIDs as `id`
- The create indicator endpoint (`POST /indicators`) returns MongoDB ObjectID as `id`
- Impact segments endpoint expects PostgreSQL UUIDs

This mismatch causes newly created custom indicators to fail when added to activities/outcomes.

## Solution

1. **Added `uuid` field to indicator types** - All indicator interfaces now include `uuid?: string | null`

2. **Prefer UUID over ID** - Components use `getIndicatorEffectiveId()` which returns `uuid || id`:
   - For V2 API indicators: uses `id` (already a PostgreSQL UUID)
   - For newly created indicators: uses `uuid` (PostgreSQL UUID from backend)

## Why This Approach

- **Backward Compatible**: Falls back to `id` if `uuid` is not present
- **Minimal Changes**: Only affects indicator selection logic
- **Works with Both APIs**: Handles V2 API (UUID as id) and create API (UUID as separate field)

## Testing Steps

1. Navigate to `community/filecoin/admin/manage-indicators`
2. Create a new Activity/Outcome
3. Click "Create Custom" to add a new custom indicator
4. Select both the new custom indicator AND system indicators
5. Save the activity
6. Verify the activity is created successfully with all indicators

## Files Changed

| File | Change |
|------|--------|
| `types/impactMeasurement.ts` | Added `uuid` to `ImpactIndicator` and `ImpactIndicatorWithData` |
| `utilities/queries/getIndicatorsByCommunity.ts` | Added `uuid` to `Indicator` and `V2Indicator` types |
| `components/Pages/Admin/IndicatorsDropdown.tsx` | Use effective ID for selection and display |
| `components/Pages/Admin/ActivityOutcomeModal.tsx` | Use effective ID when editing segments |

## Related PRs

- Backend: show-karma/gap-indexer#847 (returns `uuid` in create response)

---

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Enhancements**
  * Indicator selection and matching now prefer unique identifiers when available, improving selection accuracy.
  * Data models and queries include optional unique identifier fields to enhance compatibility and reliability across indicator-related features.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/components/Pages/Admin/IndicatorsDropdown.tsx
+++ b/components/Pages/Admin/IndicatorsDropdown.tsx
@@ -102,11 +102,19 @@ export const IndicatorsDropdown: FC<IndicatorsDropdownProps> = ({
     [indicators, newIndicators]
   );
 
+  /**
+   * Get the effective ID for an indicator (uuid preferred, fallback to id)
+   * This ensures we use PostgreSQL UUIDs for impact segments when available
+   */
+  const getIndicatorEffectiveId = (indicator: ImpactIndicator): string => {
+    return indicator.uuid || indicator.id;
+  };
+
   // Get selected indicator names for display
   const selectedNames = useMemo(() => {
     const allIndicators = [...allCommunityIndicators, ...autosyncedIndicators];
     return selectedIndicators
-      .map((id) => allIndicators.find((i) => i.id === id)?.name)
+      .map((id) => allIndicators.find((i) => getIndicatorEffectiveId(i) === id)?.name)
       .filter(Boolean);
   }, [selectedIndicators, allCommunityIndicators, autosyncedIndicators]);
 
@@ -118,14 +126,15 @@ export const IndicatorsDropdown: FC<IndicatorsDropdownProps> = ({
   }, [open]);
 
   const renderIndicatorItem = (indicator: ImpactIndicator, isDefault = false) => {
-    const isSelected = selectedIndicators.includes(indicator.id);
+    const effectiveId = getIndicatorEffectiveId(indicator);
+    const isSelected = selectedIndicators.includes(effectiveId);
 
     return (
       <button
-        key={indicator.id}
+        key={effectiveId}
         type="button"
         className="px-4 py-3 w-full hover:bg-gray-50 dark:hover:bg-zinc-700 cursor-pointer flex justify-between items-center gap-3 text-left"
-        onClick={() => handleIndicatorChange(indicator.id)}
+        onClick={() => handleIndicatorChange(effectiveId)}
       >
         <div className="flex-1 min-w-0">
           <div className="flex items-center gap-2">
--- a/components/Pages/Admin/IndicatorsDropdown.tsx
+++ b/components/Pages/Admin/IndicatorsDropdown.tsx
@@ -14,6 +14,7 @@ import { IndicatorForm } from "@/components/Forms/IndicatorForm";
 import { Button } from "@/components/Utilities/Button";
 import { useAutosyncedIndicators } from "@/hooks/useAutosyncedIndicators";
 import type { ImpactIndicator } from "@/types/impactMeasurement";
+import { getIndicatorEffectiveId } from "@/utilities/indicatorUtils";
 
 // Debounce hook for search performance
 const useDebouncedValue = (value: string, delay: number) => {
@@ -101,14 +102,6 @@ export const IndicatorsDropdown: FC<IndicatorsDropdownProps> = ({
     () => [...indicators, ...newIndicators],
     [indicators, newIndicators]
   );
-
-  /**
-   * Get the effective ID for an indicator (uuid preferred, fallback to id)
-   * This ensures we use PostgreSQL UUIDs for impact segments when available
-   */
-  const getIndicatorEffectiveId = (indicator: ImpactIndicator): string => {
-    return indicator.uuid || indicator.id;
-  };
 
   // Get selected indicator names for display
   const selectedNames = useMemo(() => {
--- a/components/Pages/Admin/ActivityOutcomeModal.tsx
+++ b/components/Pages/Admin/ActivityOutcomeModal.tsx
@@ -92,22 +92,35 @@ export const ActivityOutcomeModal = ({
     mode: "onChange",
   });
 
+  /**
+   * Get the effective ID for an indicator (uuid preferred, fallback to id)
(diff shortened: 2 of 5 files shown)


amaurymagalhaes (changes requested):
## PR #942 Code Review: `fix: prefer PostgreSQL UUID over MongoDB ID for indicators`

### Summary
This PR addresses a real problem with ID mismatches between MongoDB ObjectIDs and PostgreSQL UUIDs for indicators. However, the implementation has several issues that warrant attention before merging.

---

### Critical Issues

#### 1. **No Tests** ⚠️
This is a data integrity fix with complex matching logic, yet there are **zero tests**. Given the matching complexity in `ActivityOutcomeModal.tsx`, this is a significant risk.

```typescript
// This logic needs unit tests:
const byId = allIndicators.find(
  (ind) =>
    getIndicatorEffectiveId(ind) === segmentEffectiveId ||
    ind.id === segmentInd.id ||
    ind.uuid === segmentInd.id ||
    (segmentInd.uuid && ind.uuid === segmentInd.uuid)
);
```

**Recommendation**: Add tests for `getIndicatorEffectiveId()` and the matching logic in `ActivityOutcomeModal.tsx`.

---

#### 2. **Shotgun Matching Logic** - Code Smell
The matching logic in `ActivityOutcomeModal.tsx:103-111` uses 4 different conditions:

```typescript
getIndicatorEffectiveId(ind) === segmentEffectiveId ||
ind.id === segmentInd.id ||
ind.uuid === segmentInd.id ||
(segmentInd.uuid && ind.uuid === segmentInd.uuid)
```

This "try everything" approach suggests uncertainty about which condition actually works. It's:
- Hard to reason about
- Likely to match unintended indicators
- A maintenance burden

**Recommendation**: Document exactly which scenario each condition handles, or simplify to a deterministic matching strategy.

---

#### 3. **Dangerous Name Fallback**
```typescript
// Fallback to matching by name (case-insensitive)
const byName = allIndicators.find(
  (ind) => ind.name.toLowerCase() === segmentInd.name.toLowerCase()
);
```

Indicator names are **not unique**. Two indicators could have the same name but different data, leading to silent data corruption.

**Recommendation**: Remove name-based fallback or add logging/warnings when it's used.

---

### Medium Issues

#### 4. **Inconsistent Null Handling**
The type declares `uuid?: string | null`, but:
- `?` already makes it optional (can be `undefined`)
- `| null` adds another falsy state
- `transformIndicator` explicitly converts `undefined` to `null`

```typescript
// Type allows both undefined AND null
uuid?: string | null;

// Transform explicitly sets null
uuid: v2Indicator.uuid || null,
```

**Recommendation**: Pick one - either `uuid?: string` or `uuid: string | null`, not both.

---

#### 5. **Utility Function Lacks Edge Case Handling**
```typescript
export const getIndicatorEffectiveId = (indicator: ImpactIndicator | Indicator): string => {
  return indicator.uuid || indicator.id;
};
```

This relies on JavaScript truthiness to handle `null`, which works but is implicit. Also, if both `uuid` and `id` are somehow falsy, this returns a falsy value but the return type says `string`.

**Recommendation**: Be explicit:
```typescript
export const getIndicatorEffectiveId = (indicator: ImpactIndicator | Indicator): string => {
  if (indicator.uuid != null && indicator.uuid !== '') {
    return indicator.uuid;
  }
  return indicator.id;
};
```

---

### Minor Issues

#### 6. **JSDoc Comment is Implementation-Specific**
```typescript
/** PostgreSQL UUID - use this for impact segments when available */
uuid?: string | null;
```

Mentioning "PostgreSQL" in the type comment couples the type definition to the current backend implementation. If you migrate databases, these comments become misleading.

**Recommendation**: Use `/** Canonical unique identifier, preferred over id when available */`

---

#### 7. **Performance Consideration**
The `find()` calls in the matching logic are O(n) per segment indicator, making the overall complexity O(n×m). For communities with many indicators and segments, consider using a Map:

```typescript
const indicatorMap = new Map(
  allIndicators.flatMap(ind => [
    [ind.id, ind],
    [ind.uuid, ind],
    [getIndicatorEffectiveId(ind), ind],
  ].filter(([k]) => k))
);
```

---

### Architectural Concern

This PR treats the symptom, not the root cause. The real problem is **the backend returns inconsistent ID formats**. The frontend shouldn't need to juggle two ID systems.

**Long-term recommendation**: Standardize on one ID system in the backend, then migrate the frontend to use only that ID.

---

### What's Good

1. Created a shared utility function instead of duplicating logic
2. Clear PR description explaining the problem and solution
3. Backward compatible approach using fallback
4. Related backend PR linked

---

### Verdict

**Request Changes** - The lack of tests for a data integrity fix is a blocker. The complex matching logic needs verification before this goes to production.

**Required before merge:**
1. Add unit tests for `getIndicatorEffectiveId()`
2. Add integration tests for the indicator selection scenarios
3. Simplify or document the 4-condition matching logic

**Recommended:**
4. Remove or add warnings for name-based fallback
5. Fix the inconsistent null/undefined type handling

components/Pages/Admin/IndicatorsDropdown.tsx
```diff
@@ -106,7 +107,7 @@ export const IndicatorsDropdown: FC<IndicatorsDropdownProps> = ({
   const selectedNames = useMemo(() => {
     const allIndicators = [...allCommunityIndicators, ...autosyncedIndicators];
     return selectedIndicators
-      .map((id) => allIndicators.find((i) => i.id === id)?.name)
+      .map((id) => allIndicators.find((i) => getIndicatorEffectiveId(i) === id)?.name)
```
amaurymagalhaes:
```ts
const selectedNames = useMemo(() => {
  const byId = new Map(
    [...allCommunityIndicators, ...autosyncedIndicators]
      .map(i => [getIndicatorEffectiveId(i), i.name])
  );
  return selectedIndicators.map(id => byId.get(id)).filter(Boolean);
}, [selectedIndicators, allCommunityIndicators, autosyncedIndicators]);
```

This would have a better performance

PR #961, opened 2026-02-04
fix(funding-platform): fix false error toast on program creation

Arthh:
## Summary
- Fixed the "Failed to create program" error toast appearing despite the BE returning 201 on the funding-platform admin page (`community/optimism/admin/funding-platform`)
- Root cause: Fastify's response serialization with `zodToJsonSchema(z.any())` for the 201 schema causes `fast-json-stringify` to return `{}`, so `extractProgramId` couldn't find the `programId` in the empty response body
- Switched to using `fetchData` directly and only checking the error status, matching the existing working pattern in `AddProgram` (funding-map/add-program)

## Test plan
- [ ] Navigate to `community/<id>/admin/funding-platform` and click "Create New Program"
- [ ] Fill in the form and submit — should see success toast, modal closes, and program list refreshes
- [ ] Verify no error toast appears
- [ ] Verify the `funding-map/add-program` flow still works as before

## Summary by CodeRabbit

* **Refactor**
  * Program creation now uses a modal-driven flow (isOpen/onClose) instead of automatic navigation.
  * Removed automatic redirect after creation; UI stays in place and triggers success/close callbacks.
  * Success notification updated to "Program created successfully!"

* **Bug Fixes**
  * Creation-result handling simplified to reliably return an identifier string when available.

* **Tests**
  * Updated tests to expect no redirect and to assert success callbacks and toast behavior.

### DIFF
--- a/components/FundingPlatform/CreateProgramModal.tsx
+++ b/components/FundingPlatform/CreateProgramModal.tsx
@@ -1,6 +1,5 @@
 "use client";
 import { zodResolver } from "@hookform/resolvers/zod";
-import { useRouter } from "next/navigation";
 import { useState } from "react";
 import { Controller, useForm } from "react-hook-form";
 import toast from "react-hot-toast";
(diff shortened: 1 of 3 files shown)


components/FundingPlatform/CreateProgramModal.tsx
```diff
@@ -96,31 +96,28 @@ export function CreateProgramModal({
...
       );
 
-      // Create program using service
-      const result = await ProgramRegistryService.createProgram(
-        address,
-        community.chainID,
-        metadata
+      // Create program using fetchData directly (response body is not used,
+      // only the error status matters — same pattern as AddProgram)
+      const [, error] = await fetchData(
```
brunodmsi: we should always use services when making requests, if the current implementation fails to deliver on something, you can add it createProgram so it returns the error properly

brunodmsi (changes requested):

brunodmsi (approved):

PR #964, opened 2026-02-05
fix: resolve race condition in admin promotion/demotion UI

brunodmsi:
## Overview

Fix race condition causing UI to hang after promoting/demoting a member to admin.

## Problem

When promoting a member to admin, the UI would hang and not reflect the change until the page was manually refreshed. The transaction actually succeeded on the blockchain, but the UI displayed stale data.

**Root causes identified:**
1. **QueryClient mismatch**: `refreshProject()` in the project store created a NEW `QueryClient` instance instead of using the shared one, so cache invalidation had no effect on the UI
2. **Cache invalidation timing**: Cache was only invalidated AFTER the long polling completed (up to 25 minutes max), leaving the UI displaying stale data during the entire wait

## Solution

1. **Use shared queryClient**: Import the shared `queryClient` from `@/utilities/query-client` instead of creating a new instance in the project store
2. **Early cache invalidation**: Invalidate cache immediately after the indexer is notified (before polling starts), so the UI can begin refetching while polling continues in the background
3. **Final invalidation**: Keep a final cache invalidation after polling confirms success to ensure fresh data

## Why This Approach

- Minimal changes with maximum impact
- Maintains existing polling logic for confirmation
- UI becomes responsive immediately after transaction, with final confirmation after polling
- Consistent queryClient usage across the codebase

## Testing Steps

1. Navigate to a project's Team tab
2. Click the shield icon to promote a member to admin
3. Confirm the promotion
4. **Expected**: UI should update to show the new admin role without needing a page refresh
5. Repeat for demoting an admin back to member

## Technical Details

```
Before:
Transaction → Notify Indexer → Poll (1.5s × N) → Invalidate Cache
                                    ↑
                             UI shows stale data

After:
Transaction → Notify Indexer → Invalidate Cache → Poll → Final Invalidation
                                    ↓                          ↓
                             UI starts refetching      Ensures fresh data
```

## Files Changed

- `store/project.ts` - Use shared queryClient instead of creating new instance
- `components/Dialogs/Member/PromoteMember.tsx` - Early cache invalidation + consistent import
- `components/Dialogs/Member/DemoteMember.tsx` - Early cache invalidation + consistent import

## Checklist

- [x] Changes tested manually
- [x] No breaking changes
- [x] Consistent with existing codebase patterns

---
🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

---
  - https://app.asana.com/0/0/1212944852711819

### DIFF
--- a/__tests__/unit/utilities/indexer-notification.test.ts
+++ b/__tests__/unit/utilities/indexer-notification.test.ts
@@ -13,7 +13,7 @@ import * as queryKeysModule from "@/utilities/queryKeys";
 
 // Mock dependencies
 jest.mock("@/utilities/fetchData");
-jest.mock("@/components/Utilities/PrivyProviderWrapper", () => ({
+jest.mock("@/utilities/query-client", () => ({
   queryClient: {
     invalidateQueries: jest.fn(),
   },
@@ -24,7 +24,7 @@ const mockFetchData = fetchDataModule.default as jest.MockedFunction<
 >;
 
 // Import queryClient after mocking
-import { queryClient } from "@/components/Utilities/PrivyProviderWrapper";
+import { queryClient } from "@/utilities/query-client";
 
 const mockInvalidateQueries = queryClient.invalidateQueries as jest.MockedFunction<
   typeof queryClient.invalidateQueries
--- a/components/Dialogs/Member/DeleteMember.tsx
+++ b/components/Dialogs/Member/DeleteMember.tsx
@@ -6,7 +6,7 @@ import { type FC, Fragment, useState } from "react";
 import { useAccount } from "wagmi";
 import { Button } from "@/components/Utilities/Button";
 import { errorManager } from "@/components/Utilities/errorManager";
-import { queryClient } from "@/components/Utilities/PrivyProviderWrapper";
+import { queryClient } from "@/utilities/query-client";
 import { useAttestationToast } from "@/hooks/useAttestationToast";
 import { useGap } from "@/hooks/useGap";
 import { useOffChainRevoke } from "@/hooks/useOffChainRevoke";
--- a/components/Dialogs/Member/DemoteMember.tsx
+++ b/components/Dialogs/Member/DemoteMember.tsx
@@ -2,20 +2,10 @@ import { Dialog, Transition } from "@headlessui/react";
 import { ShieldExclamationIcon } from "@heroicons/react/24/outline";
 import * as Tooltip from "@radix-ui/react-tooltip";
 import { type FC, Fragment, useState } from "react";
-import { useAccount } from "wagmi";
 import { Button } from "@/components/Utilities/Button";
-import { errorManager } from "@/components/Utilities/errorManager";
-import { useAttestationToast } from "@/hooks/useAttestationToast";
-import { useSetupChainAndWallet } from "@/hooks/useSetupChainAndWallet";
+import { useMemberRoleChange } from "@/hooks/useMemberRoleChange";
 import { useTeamProfiles } from "@/hooks/useTeamProfiles";
-import { useWallet } from "@/hooks/useWallet";
 import { useProjectStore } from "@/store";
-import fetchData from "@/utilities/fetchData";
-import { getProjectMemberRoles } from "@/utilities/getProjectMemberRoles";
-import { INDEXER } from "@/utilities/indexer";
-import { queryClient } from "@/utilities/query-client";
-import { retryUntilConditionMet } from "@/utilities/retries";
-import { getProjectById } from "@/utilities/sdk";
 
 interface DemoteMemberDialogProps {
   memberAddress: string;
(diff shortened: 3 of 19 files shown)


amaurymagalhaes:
## PR #964 Code Review: fix: resolve race condition in admin promotion/demotion UI

### Summary

Solid, well-scoped bugfix that correctly identifies and addresses two real root causes: (1) a `QueryClient` singleton mismatch in `store/project.ts` where `refreshProject()` created a throwaway `QueryClient` instance whose cache invalidation was invisible to the React tree, and (2) cache invalidation timing that left the UI stale during the entire polling phase. The fix is minimal, correct, and well-documented. The PR description is excellent -- clear problem statement, before/after flow diagrams, and testing steps.

---

### Critical Issues (P0/P1)

**None found.** The core logic change is sound.

---

### Other Issues (P2/P3)

**1. [P2] Fire-and-forget `refreshProject()` in early invalidation block**

In both `PromoteMember.tsx` (line ~97 of the new code) and `DemoteMember.tsx` (line ~97), the early `refreshProject()` call is not awaited:

```typescript
queryClient.invalidateQueries({
  queryKey: ["memberRoles", project?.uid],
});
refreshProject();  // <-- not awaited
```

`refreshProject()` is an async function that fetches the project, refreshes grants, and invalidates queries. Calling it without `await` means:
- Any error thrown inside it will become an unhandled promise rejection
- The subsequent `checkIfAttestationExists` polling starts immediately, potentially racing with the refresh

**Suggestion**: Either `await` it, or explicitly catch errors:

```typescript
// Option A: await it
await refreshProject();

// Option B: fire-and-forget with error handling
refreshProject().catch((err) =>
  errorManager("Error during early project refresh", err)
);
```

Option B is arguably more appropriate here since the intent is to be non-blocking while polling starts, but the unhandled rejection must still be caught.

---

**2. [P2] `queryClient.invalidateQueries` is also fire-and-forget in the early block**

Same location in both files -- `invalidateQueries` returns a Promise. While React Query is generally forgiving about this, consistency with the later block (which also doesn't await it) would be better if both were explicitly handled. This is minor since React Query internally handles this gracefully, but worth noting.

---

**3. [P3] Duplicate invalidation pattern -- consider extracting a helper**

Both `PromoteMember.tsx` and `DemoteMember.tsx` now contain identical cache invalidation logic:

```typescript
// Early invalidation (identical in both files)
queryClient.invalidateQueries({ queryKey: ["memberRoles", project?.uid] });
refreshProject();

// ... later ...

// Final invalidation (identical in both files)
await refreshProject();
queryClient.invalidateQueries({ queryKey: ["memberRoles", project?.uid] });
```

These two files are already near-duplicates of each other (same structure, same flow, only differing in `addAdmin` vs `removeAdmin` and text labels). A shared utility or custom hook (e.g., `useAdminRoleMutation`) would reduce this duplication. Not blocking for this PR, but worth tracking as tech debt.

---

**4. [P3] The `queryKey: ["memberRoles", project?.uid]` uses optional chaining**

At this point in the flow, `project` is guaranteed non-null (checked at the top of the function with `if (!address || !project) return`). Using `project?.uid` instead of `project.uid` is harmless but subtly misleading -- it suggests `project` might be null when it cannot be at this point. Extremely minor nit.

---

### Suggestions

1. **Consider adding a test for the queryClient singleton**: The root cause was that `store/project.ts` created its own `QueryClient`. A simple unit test asserting that `useProjectStore.getState().refreshProject` uses the shared `queryClient` (e.g., by spying on `queryClient.invalidateQueries`) would prevent regression. The PR description mentions manual testing only.

2. **The backwards-compat re-export in `PrivyProviderWrapper.tsx` is clean**: I confirmed no files still import `queryClient` from `PrivyProviderWrapper` -- the `export { queryClient }` re-export there is a good safety net. This can be removed in a future cleanup PR once the migration is considered stable.

3. **PR description quality is exemplary**: The before/after flow diagram, clear root cause analysis, and testing steps are exactly what a bugfix PR should look like. Well done.

---

### Score: 8/10

**Merge recommendation: APPROVE**

This is a correct, minimal fix for a real user-facing bug. The only actionable item before merge is addressing the unhandled promise rejection on the fire-and-forget `refreshProject()` call (P2 issue #1 above) -- a one-line fix. Everything else is nits or future cleanup. The store fix is the real win here and it is unambiguously correct.

amaurymagalhaes:
## Code Review -- PR #964: fix: resolve race condition in admin promotion/demotion UI

**Reviewer**: Frontend Agent (Claude Opus 4.6)

---

### Analysis

This PR addresses a real and well-diagnosed bug: the `refreshProject()` function in the Zustand store was creating a **new `QueryClient` instance** on every call, meaning its `invalidateQueries` call had zero effect on the shared React Query cache powering the UI. The fix correctly switches to the shared `queryClient` singleton from `@/utilities/query-client` and adds an early cache invalidation so the UI can start refetching immediately after the indexer is notified, rather than waiting for polling (up to 1000 retries x 1.5s = 25 minutes worst case) to complete.

The root cause analysis is accurate, the solution is targeted, and the PR description is excellent. That said, there are several issues -- some architectural, some correctness-related -- that should be addressed before merge.

---

### Errors & Warnings

#### 1. CRITICAL: Unawaited `refreshProject()` in early invalidation (both dialogs)

```typescript
// Lines ~95-100 in both PromoteMember.tsx and DemoteMember.tsx
queryClient.invalidateQueries({
  queryKey: ["memberRoles", project?.uid],
});
refreshProject(); // <-- fire-and-forget, not awaited

await checkIfAttestationExists(() => { ... })
```

`refreshProject()` is an async function that fetches the project from the API, updates Zustand state, and invalidates the `["project"]` query. By not awaiting it, you create a subtle race condition:

- The `refreshProject` network request races against `checkIfAttestationExists` polling.
- If `refreshProject` resolves while polling is in progress, the Zustand `set({ project })` call could overwrite the store with stale data that the polling loop then reads.
- Any error thrown by `refreshProject` is silently swallowed (unhandled promise rejection).

**Recommendation**: Either `await refreshProject()` or explicitly handle the fire-and-forget with `.catch()` to prevent unhandled rejections:

```typescript
// Option A: await it
await refreshProject();

// Option B: explicit fire-and-forget with error handling
refreshProject().catch((err) => {
  console.error("Early project refresh failed:", err);
});
```

#### 2. `invalidateQueries` returns a Promise that is not awaited

In both dialog components, the early `queryClient.invalidateQueries(...)` call is not awaited. While React Query will still trigger refetches, the invalidation is technically asynchronous. If you want to guarantee the invalidation completes before proceeding:

```typescript
await queryClient.invalidateQueries({
  queryKey: ["memberRoles", project?.uid],
});
```

The same applies to the final invalidation calls at the end. This is a minor point since React Query handles this gracefully, but for consistency and correctness, awaiting is preferred.

#### 3. Incomplete migration leaves codebase in split-import state

The PR migrates `PromoteMember.tsx` and `DemoteMember.tsx` from importing `queryClient` via `@/components/Utilities/PrivyProviderWrapper` to the new canonical `@/utilities/query-client`. However, **at least 5 other files** still import from the old location:

- `components/Dialogs/Member/DeleteMember.tsx`
- `hooks/useInviteLink.ts`
- `utilities/indexer-notification.ts`
- `hooks/v2/useProjectImpacts.ts`
- `components/Shared/ActivityCard/MilestoneCard.tsx`
- (and likely others from the search results)

The `PrivyProviderWrapper` now re-exports from `@/utilities/query-client` for backwards compatibility, so this does not cause a functional bug. However, having two import paths for the same singleton is a maintenance hazard and creates confusion about which is canonical.

**Recommendation**: While I understand keeping the PR scoped, at minimum add a `@deprecated` JSDoc comment on the re-export in `PrivyProviderWrapper.tsx`, and consider filing a follow-up issue to migrate remaining imports. Or better yet, include the migration in this PR since it is a trivial find-and-replace.

---

### Suggestions

#### 4. Extract shared role-change logic into a custom hook

`PromoteMember.tsx` and `DemoteMember.tsx` are nearly identical (~90% shared code). The only differences are:
- `addAdmin` vs `removeAdmin` method call
- The condition check in polling (`=== "Admin"` vs `!== "Admin"`)
- UI labels ("Promote" vs "Remove as admin")

This duplication means any future fix to one must be manually applied to the other (as this PR itself demonstrates -- the exact same 8-line change was applied to both files).

**Recommendation**: Extract a `useMemberRoleChange` hook:

```typescript
function useMemberRoleChange(action: "promote" | "demote") {
  // shared wallet setup, attestation toast, polling, cache invalidation
  // returns { execute, isLoading }
}
```

This is not a blocking concern for this PR, but it is a strong improvement to make alongside these changes.

#### 5. Consider `queryKey` consistency

The early invalidation uses `["memberRoles", project?.uid]` while `refreshProject()` internally invalidates `["project"]`. These are different query keys targeting different cached data. This is correct behavior, but the intent should be documented more clearly. A reader might wonder whether both invalidations are necessary or if one subsumes the other.

#### 6. Defensive check on `project?.uid` in query key

Both dialog components use `project?.uid` in the query key. If `project` were somehow `undefined` at this point, the query key would be `["memberRoles", undefined]`, which would not match any cached query. The function already guards with `if (!address || !project) return;` at the top, so `project` is guaranteed to be defined at the invalidation point. However, using the non-null assertion `project.uid` (without `?.`) would make this guarantee explicit and avoid confusing future readers:

```typescript
queryClient.invalidateQueries({
  queryKey: ["memberRoles", project.uid], // project is guaranteed non-null here
});
```

#### 7. The `as any` cast on `walletSigner` (pre-existing, not introduced by this PR)

Both files cast `walletSigner as any`. While this is a pre-existing issue, it is worth noting as it defeats TypeScript's type safety at a critical blockchain interaction boundary. A follow-up to type the wallet signer properly would reduce risk of runtime errors.

---

### Conclusion

**The core fix is correct and well-reasoned.** Replacing the per-call `new QueryClient()` with the shared singleton is the right solution, and the early cache invalidation strategy is a sound approach to improve perceived responsiveness.

However, **the unawaited `refreshProject()` call (issue #1) should be addressed before merge**, as it introduces the same category of bug (race condition) that this PR aims to fix. The remaining suggestions are improvements that would strengthen the change but are not strictly blocking.

**Verdict**: Request changes on issue #1 (unawaited async call). The rest are non-blocking recommendations.

| Category | Assessment |
|----------|-----------|
| Root cause analysis | Correct and thorough |
| Fix approach | Sound -- shared singleton + early invalidation |
| Correctness | One unawaited async call needs attention |
| Code quality | Good, but duplication between Promote/Demote is a concern |
| Migration completeness | Partial -- other files still use old import path |
| PR description | Excellent -- clear problem statement and diagrams |

PR #970, opened 2026-02-06
fix: remove dead V1 /programs/:id code path

brunodmsi:
## Summary
- Removed the `useProgram` hook, `getProgram` service method, and `INDEXER.PROGRAMS.GET` URL builder that targeted a non-existent `GET /programs/{programId}` endpoint
- Cleaned up `ApplicationListWithAPI.tsx` and `AIPromptConfiguration.tsx` which were the only two consumers
- Net removal of 44 lines of dead code that was silently 404ing on every funding-platform admin page load

## Test plan
- [ ] Load `/community/{id}/admin/funding-platform/{programId}/applications` — verify table renders, infinite scroll works, AI Score column visibility unchanged
- [ ] Load `/community/{id}/admin/funding-platform/{programId}/question-builder` — verify AI config section loads, prompt tabs render, legacy settings work
- [ ] Confirm `GET /programs/{programId}` no longer fires in the Network tab

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * AI-related UI now derives behavior solely from form/schema settings rather than external program data, simplifying configuration and reducing unexpected variability.
  * AI column visibility and prompt configuration consistently follow schema-provided values.
* **Tests**
  * Adjusted test coverage to reflect the new schema-only behavior, removing legacy fallback scenarios.

### DIFF
--- a/__tests__/integration/components/FundingPlatform/helper/getAIColumnVisibility.test.ts
+++ b/__tests__/integration/components/FundingPlatform/helper/getAIColumnVisibility.test.ts
@@ -50,41 +50,6 @@ describe("getAIColumnVisibility", () => {
     });
   });
 
-  describe("When using fallback langfusePromptId", () => {
-    it("should show AI score column when fallback langfusePromptId is provided", () => {
-      const formSchema = undefined;
-      const langfusePromptId = "fallback-prompt-789";
-
-      const result = getAIColumnVisibility(formSchema, langfusePromptId);
-      expect(result.showAIScoreColumn).toBe(true);
-      expect(result.showInternalAIScoreColumn).toBe(false);
-    });
-
-    it("should prefer aiConfig langfusePromptId over fallback", () => {
-      const formSchema = {
-        aiConfig: {
-          langfusePromptId: "config-prompt",
-        },
-      };
-      const langfusePromptId = "fallback-prompt";
-
-      const result = getAIColumnVisibility(formSchema, langfusePromptId);
-      expect(result.showAIScoreColumn).toBe(true);
-      expect(result.showInternalAIScoreColumn).toBe(false);
-    });
-
-    it("should use fallback when aiConfig langfusePromptId is missing", () => {
-      const formSchema = {
-        aiConfig: {},
-      };
-      const langfusePromptId = "fallback-prompt";
-
-      const result = getAIColumnVisibility(formSchema, langfusePromptId);
-      expect(result.showAIScoreColumn).toBe(true);
-      expect(result.showInternalAIScoreColumn).toBe(false);
-    });
-  });
-
   describe("Edge cases", () => {
     it("should handle undefined formSchema", () => {
       const result = getAIColumnVisibility(undefined);
@@ -117,15 +82,6 @@ describe("getAIColumnVisibility", () => {
       };
 
       const result = getAIColumnVisibility(formSchema);
-      expect(result.showAIScoreColumn).toBe(false);
-      expect(result.showInternalAIScoreColumn).toBe(false);
-    });
-
-    it("should handle empty string fallback prompt ID as falsy", () => {
-      const formSchema = undefined;
-      const langfusePromptId = "";
-
-      const result = getAIColumnVisibility(formSchema, langfusePromptId);
       expect(result.showAIScoreColumn).toBe(false);
       expect(result.showInternalAIScoreColumn).toBe(false);
     });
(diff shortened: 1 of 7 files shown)


amaurymagalhaes:
## Code Review: PR #970 — fix: remove dead V1 /programs/:id code path

**Reviewer**: pr-review-agent (Claude Opus 4.6)

---

### Summary

Clean, focused removal of a dead code path (`GET /programs/{programId}`) that was silently 404ing. The change is directionally correct and well-scoped at +4/-48 lines across 5 files. However, the cleanup is **incomplete** — the `getAIColumnVisibility` helper function and its test suite were not updated to remove the now-unreachable fallback parameter, leaving orphaned code and misleading test coverage.

---

### Critical Issues (P0/P1)

**1. [P1] `getAIColumnVisibility` has an orphaned fallback parameter — behavioral regression risk**

The function signature in `components/FundingPlatform/helper/getAIColumnVisibility.ts` still accepts a second `langfusePromptId?: string` parameter:

```typescript
export function getAIColumnVisibility(
  formSchema: unknown,
  langfusePromptId?: string  // <-- Dead parameter, no caller ever passes this now
)
```

And the implementation still branches on it:
```typescript
showAIScoreColumn: !!(aiConfig?.langfusePromptId || langfusePromptId),
```

After this PR, the call site in `ApplicationListWithAPI.tsx` (line ~122 in the diff) passes only one argument:
```typescript
() => getAIColumnVisibility(config?.formSchema),
```

This means `langfusePromptId` is always `undefined`, and the `|| langfusePromptId` branch is dead code. While this is not a runtime bug (since the endpoint was already 404ing and the value was always `undefined` in practice), it is:

- **Misleading**: Future developers will think the fallback is functional.
- **Incomplete cleanup**: The stated goal of removing the dead V1 code path is not fully achieved.
- **Test rot**: The test file `__tests__/integration/components/FundingPlatform/helper/getAIColumnVisibility.test.ts` has an entire `"When using fallback langfusePromptId"` describe block (4 test cases) and a `"legacy program with fallback prompt ID"` test that exercise a code path that can never be triggered at runtime.

**Fix**: Update `getAIColumnVisibility` to remove the second parameter:

```typescript
export function getAIColumnVisibility(
  formSchema: unknown
): {
  showAIScoreColumn: boolean;
  showInternalAIScoreColumn: boolean;
} {
  const formSchemaWithAiConfig = formSchema as FormSchemaWithAiConfig | undefined;
  const aiConfig = formSchemaWithAiConfig?.aiConfig;

  return {
    showAIScoreColumn: !!aiConfig?.langfusePromptId,
    showInternalAIScoreColumn: !!aiConfig?.internalLangfusePromptId,
  };
}
```

Then remove the "When using fallback langfusePromptId" describe block and the "legacy program with fallback prompt ID" test case from the test file. Update the JSDoc accordingly.

---

**2. [P1] `GrantProgram.langfusePromptId` field is still typed but now unqueryable**

The `FundingProgramResponse` type in `src/features/funding-map/types/funding-program.ts` still declares `langfusePromptId?: string`. The community programs endpoint (`GET /communities/{communityId}/programs`) may still return this field from the indexer, so this is not necessarily wrong. But the PR description says the single-program endpoint never existed — if `langfusePromptId` was only meaningful via the now-removed single-program fetch, consider whether it should be documented as deprecated on the type or if it is still returned by the community programs list endpoint.

This is informational, not blocking.

---

### Other Issues (P2/P3)

**3. [P2] Unused `GrantProgram` import may still be referenced only by dead service method**

In `services/programs.ts`, the `GrantProgram` import is retained because `getCommunityPrograms` still uses it. However, `GrantProgram` is itself marked `@deprecated` in `ProgramList.tsx`:
```typescript
/** @deprecated Use FundingProgramResponse from @/src/features/funding-map/types/funding-program instead */
export type GrantProgram = FundingProgramResponse;
```

This is a pre-existing issue, not introduced by this PR, but worth noting for a follow-up migration.

**4. [P3] Comment update in `ApplicationListWithAPI.tsx` is slightly imprecise**

Diff line ~105:
```diff
-  // Fetch program config and program data to determine AI column visibility
+  // Fetch program config to determine AI column visibility
```

Good cleanup, no issue here — just confirming this is accurate after the change.

---

### Suggestions

1. **Complete the cleanup in this PR**: Update `getAIColumnVisibility` to remove the dead second parameter and update its tests. This keeps the PR self-contained as "remove all dead V1 program code" rather than leaving orphaned artifacts. This is the minimum I would want before merging.

2. **Consider a follow-up PR** to migrate `GrantProgram` type alias usages to `FundingProgramResponse` directly, since the alias is already marked deprecated.

3. **Verify the indexer**: The PR description states `GET /programs/{programId}` never existed. If there is any chance it was recently added or planned, this removal would break integration. Confirm with the backend/indexer team. (This is likely already verified given the PR description confidence, but worth a sanity check.)

---

### Overall Score: **7/10**

| Category | Assessment |
|----------|-----------|
| Correctness | Good — no runtime regressions since the endpoint was already 404ing |
| Completeness | Incomplete — `getAIColumnVisibility` and its tests need updating |
| Code quality | Clean removal, well-structured diff |
| Testing | Existing tests not updated to match new reality |
| PR description | Excellent — clear summary, test plan, and rationale |

**Merge recommendation**: **Approve with required changes**. The `getAIColumnVisibility` function and its test suite should be updated to remove the dead fallback parameter before merging. This is a 5-minute fix that makes the cleanup complete.

PR #980, opened 2026-02-10
feat(donations): FIAT donation flow - polling, progress UI, and modal lifecycle

brunodmsi:
## Overview

Frontend changes for the complete FIAT donation flow: real-time donation status polling, a redesigned success modal with progress stepper, amount propagation to the card flow, and proper modal lifecycle management.

## Problem

1. **No real polling**: After Stripe payment success, `OnrampSuccessModal` showed static Stripe event data with no mechanism to poll the backend for updated donation status (pending → completed).
2. **Amount not propagating**: `OnrampFlow` had its own independent `amount` state and never received the `initialAmount` entered in the parent modal.
3. **Modal stacking**: Closing the success modal only cleared local state but didn't close the parent `SingleProjectDonateModal`, leaving the user looking at the donation form again.
4. **Hardcoded chain ID**: A magic number `8453` was used as fallback instead of referencing the network constants.
5. **No error visibility**: Polling failures were silently swallowed with no feedback to the user.

### Donation polling
- Added `getDonationByUid` to `donationsService` — calls authenticated `GET /v2/donations/:uid/:chainId`
- New `useDonationPolling` hook — uses `useQuery` with 5s `refetchInterval`, stops on terminal states (completed/failed)
- Exposes `error` state so consumers can show polling failure messages

### Redesigned OnrampSuccessModal
- 3-step progress stepper: Payment → Delivering (pulse animation) → Done
- Real-time status updates from backend polling
- Transaction details card: crypto amount, network, status badge, explorer link
- State transitions: delivering → completed (green) or failed (red)
- Subtle error message when polling fails: "Unable to check donation status"

### Amount propagation
- `SingleProjectDonateModal` passes `initialAmount={amount}` to `OnrampFlow`
- `OnrampFlow` initializes from `initialAmount` and syncs via `useEffect`

### Modal lifecycle
- `SingleProjectDonateModal` passes `onDonationComplete={onClose}` to `OnrampFlow`
- `donationUidRef` captures UID before `clearSession()` resets mutation data
- `handleSuccessModalClose` calls `onDonationComplete()` to close parent modal

### Minor fixes (PR review round 2)
- Replaced `DEFAULT_CHAIN_ID = 8453` magic number with `NETWORK_CHAIN_IDS['base']`
- Exposed `error: Error | null` from `useDonationPolling` hook
- Added polling error message in `OnrampSuccessModal`

## Why This Approach

- **Ref for donationUid**: The `useOnramp` hook's `session` is cleared by `clearSession()` before the success modal opens. A ref captures the UID before clearing so it survives the state reset.
- **YAGNI-clean**: Single-use components were inlined. `Set` for terminal statuses replaced with direct comparisons. `useMemo` removed where a plain `const` suffices.
- **Polling stops automatically**: React Query's `refetchInterval` callback returns `false` on terminal states — no manual cleanup needed.
- **Named constant over magic number**: `NETWORK_CHAIN_IDS['base']` is self-documenting and stays in sync if the chain ID ever changes.

## Testing Steps

1. Open a project donation modal, enter an amount (e.g. $50)
2. Select "Card" payment method → verify the amount shows "50" in the fiat input
3. Complete a Stripe test payment (card: `4242 4242 4242 4242`)
4. **Success modal**: Verify 3-step progress stepper appears with "Payment" completed and "Delivering" active
5. **Polling**: Observe status updating from "Delivering" to "Completed" when webhook processes
6. **Explorer link**: Verify "View" link points to correct block explorer
7. **Close**: Click "Done & Close" → both success modal AND parent donation modal close
8. **Failed state**: If donation fails, Step 2 shows red X, status badge shows "Failed"
9. **Polling error**: If backend is unreachable, "Unable to check donation status" message appears
10. **Unit tests**: `npx jest hooks/donation components/Donation/SingleProject/__tests__ --no-coverage` — all pass

## Architecture

```mermaid
sequenceDiagram
    participant User
    participant Modal as SingleProjectDonateModal
    participant Flow as OnrampFlow
    participant Stripe as StripeOnrampEmbed
    participant Success as OnrampSuccessModal
    participant Hook as useDonationPolling
    participant API as Backend API

    User->>Modal: Enter amount, select Card
    Modal->>Flow: initialAmount, onDonationComplete
    Flow->>Stripe: initiateOnramp(amount, currency)
    Stripe-->>Flow: onSuccess(sessionData)
    Note over Flow: Capture donationUid via ref
    Flow->>Success: sessionData, donationUid, chainId
    Success->>Hook: useDonationPolling(uid, chainId)
    loop Every 5s until terminal
        Hook->>API: GET /v2/donations/:uid/:chainId
        API-->>Hook: { status, transactionHash, ... }
        Hook-->>Success: Update status display
    end
    User->>Success: Click "Done & Close"
    Success->>Flow: onClose → onDonationComplete()
    Flow->>Modal: Close parent modal
```

## Files Changed

| File | Change |
|------|--------|
| `services/donations.service.ts` | Added `getDonationByUid` method |
| `hooks/donation/useDonationPolling.ts` | Polling hook with `useQuery` + `refetchInterval` + error exposure |
| `hooks/donation/__tests__/useDonationPolling.test.tsx` | Tests including error state |
| `OnrampSuccessModal.tsx` | Progress stepper, polling, error message, named chain ID constant |
| `OnrampFlow.tsx` | Added `initialAmount`, `onDonationComplete`, `donationUidRef` |
| `SingleProjectDonateModal.tsx` | Passes `initialAmount` and `onDonationComplete` to OnrampFlow |

## Release Notes

* **New Features**
  * Donation success modal now displays a multi-step progress tracker showing Payment, Delivering, and Done stages
  * Added real-time donation status updates with transaction explorer link integration
  * Donation amounts are now pre-populated and modal automatically closes upon successful completion

### DIFF
--- a/components/Donation/SingleProject/OnrampFlow.tsx
+++ b/components/Donation/SingleProject/OnrampFlow.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { CreditCard, Loader2 } from "lucide-react";
-import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, { useCallback, useMemo, useRef, useState } from "react";
 import { Button } from "@/components/ui/button";
 import { OnrampProvider, type StripeOnrampSessionData } from "@/hooks/donation/types";
 import { useOnramp } from "@/hooks/donation/useOnramp";
@@ -36,12 +36,6 @@ export const OnrampFlow = React.memo<OnrampFlowProps>(
     const [successSessionData, setSuccessSessionData] = useState<StripeOnrampSessionData | null>(
       null
     );
-
-    useEffect(() => {
-      if (initialAmount) {
-        setAmount(initialAmount);
-      }
-    }, [initialAmount]);
 
     const { country, isLoading: isCountryLoading } = useCountryDetection();
     const isCountryAllowed = useMemo(() => isCountrySupported(country), [country]);
--- a/components/Donation/SingleProject/OnrampFlow.tsx
+++ b/components/Donation/SingleProject/OnrampFlow.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { CreditCard, Loader2 } from "lucide-react";
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Button } from "@/components/ui/button";
 import { OnrampProvider, type StripeOnrampSessionData } from "@/hooks/donation/types";
 import { useOnramp } from "@/hooks/donation/useOnramp";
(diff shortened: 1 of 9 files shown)


amaurymagalhaes (commented):
## PR Review: feat(donations): FIAT donation flow - polling, progress UI, and modal lifecycle

**Summary**: Solid, well-structured feature PR. The polling hook is cleanly designed, the progress stepper UI is a clear improvement over the static modal, and the ref-based UID capture is the right call given `clearSession()` semantics. Tests are thorough and cover the important edge cases. A couple of real issues below, but nothing blocking.

---

### Critical Issues (P0/P1)

**1. [P1] `aria-labelledby` points to conditionally-rendered element**
`OnrampSuccessModal.tsx` — The `id="onramp-success-title"` is placed on the `<p>` subtitle element (diff line ~689-693), but the `<h2>` with `formattedFiatAmount` is rendered conditionally above it. When `formattedFiatAmount` is null, the dialog still has an accessible name since the `<p>` with the id is always rendered — so this is actually fine structurally. However, the semantic issue is that `aria-labelledby` points to a `<p>` subtitle (`"Payment Successful — Crypto is being delivered..."`) rather than a concise heading. A screen reader will announce the entire sentence as the dialog name. Consider either:
- Moving the `id` to a short, always-rendered `<h3>` or similar heading element
- Or keeping as-is if the verbose announcement is acceptable

**Severity**: P1 accessibility — functional but suboptimal for screen readers.

**2. [P1] `resolvedChainId` computed in modal is unused for polling — `chainId` prop used instead**
`OnrampSuccessModal.tsx` — The component computes `resolvedChainId` (diff line ~543-547) from `txDetails?.destination_network || network`, which resolves the correct chain ID from the transaction. But the `useDonationPolling` call (diff line ~521-529) uses the raw `chainId` prop directly:

```tsx
const { donation, isPolling, status: polledStatus, error: pollingError } = useDonationPolling({
  donationUid,
  chainId, // ← raw prop, not resolvedChainId
});
```

If `chainId` (the prop from the parent) ever differs from the resolved chain ID (from `destination_network`), polling would hit the wrong endpoint. In practice these likely match for now since it's all Base, but this is a latent bug.

**Fix**: Either pass `resolvedChainId` to the polling hook (requires reordering to compute it first), or confirm that `chainId` is always canonical and remove `resolvedChainId` for the explorer URL too.

---

### Other Issues (P2/P3)

**3. [P2] `useEffect` for `initialAmount` sync can overwrite user-typed input**
`OnrampFlow.tsx` diff line ~144-148:

```tsx
useEffect(() => {
  if (initialAmount) {
    setAmount(initialAmount);
  }
}, [initialAmount]);
```

If the parent re-renders with the same `initialAmount` prop (e.g., due to unrelated state changes), this effect is harmless because the dependency hasn't changed. But if `initialAmount` genuinely changes while the user has manually edited the amount field, their input gets silently overwritten. The `useState(initialAmount || "")` on line 137 already handles initial seeding. Unless there's a specific scenario where the parent updates `initialAmount` after mount, consider removing the `useEffect` entirely — the `useState` initializer is sufficient.

**4. [P2] `donationUidRef` read during render may be stale**
`OnrampFlow.tsx` diff line ~251:

```tsx
donationUid={donationUidRef.current}
```

Refs don't trigger re-renders. The ref is set in `handleStripeSuccess` (line ~173) right before `clearSession()` and `setSuccessSessionData()`. Since `setSuccessSessionData` triggers the re-render that shows `OnrampSuccessModal`, the ref *should* have the correct value by the time the modal renders. This works because `handleStripeSuccess` is synchronous — the ref assignment and state update happen in the same microtask. Just noting this is a subtle ordering dependency.

**5. [P3] Repetitive ternary chains for status-derived UI**
`OnrampSuccessModal.tsx` diff lines ~595-636 — Six separate ternary blocks all switching on the same `resolvedStatus`. A lookup map would be cleaner:

```tsx
const STATUS_CONFIG = {
  completed: { title: "Donation Complete", subtitle: "...", label: "Completed", ... },
  failed: { title: "Donation Failed", ... },
  delivering: { title: "Payment Successful", ... },
} as const;
const config = STATUS_CONFIG[resolvedStatus];
```

**6. [P3] Test uses `document.querySelector` for assertions**
`OnrampSuccessModal.test.tsx` — The `getSubtitleParagraph` helper (diff line ~992) uses `document.querySelector("p.text-sm.text-gray-500")` which couples tests to Tailwind class names. If styling changes, tests break for non-behavioral reasons. Consider adding a `data-testid` attribute instead.

**7. [P3] Rerender without wrapper in test**
`OnrampFlow.test.tsx` diff line ~856:

```tsx
rerender(<OnrampFlow {...defaultProps} initialAmount="200" />);
```

The `rerender` call doesn't pass the wrapper, which means the QueryClientProvider is lost. This might work by accident if the component tree is still mounted inside the original wrapper, but it's fragile. RTL's `rerender` from the same `render` call does preserve the wrapper, so this is actually fine — just noting for clarity.

---

### Suggestions

1. **Consider `refetchOnWindowFocus: false`** on the polling query to avoid surprise refetches when the user tabs back to the page during delivery.
2. The polling hook uses `retry: false` — if the first poll fails due to a transient network blip, the user immediately sees an error. Consider `retry: 1` or `retry: 2` with a short delay for resilience.
3. The `encodeURIComponent(uid)` in `getDonationByUid` is good defensive coding. The `chainId` parameter is a number so it's safe as-is.

---

### Overall Score: **8/10** — Approve with minor fixes

Well-written, well-tested, clean architecture. The polling hook design is exactly right (React Query `refetchInterval` callback is the idiomatic approach). The ref-based UID capture is clever and correctly handles the `clearSession()` timing issue. The progress stepper is a meaningful UX improvement.

The P1 `chainId` vs `resolvedChainId` mismatch should be addressed before merge. The `useEffect` for `initialAmount` sync is worth a second look. Everything else is polish.

**Recommendation**: Approve after addressing #2 (chainId mismatch).

amaurymagalhaes (commented):
i would just get rid of parseFloats to a more solid way

PR #1188, opened 2026-03-31
fix: move milestone status filtering to API, address review comments

amaurymagalhaes:
## Overview

Moves milestone status filtering (pending/completed/verified) from the frontend to the backend API, and addresses all review comments from the original PR.

## Problem

The original implementation fetched all project updates from the API and then filtered them client-side in the browser. This meant:
- Unnecessary data transferred over the network
- Business logic duplicated on the frontend
- State management complexity in the UI layer

## Solution

1. **Backend (gap-indexer)**: Added `milestoneStatus` query parameter to the `GET /v2/projects/:projectSlug/updates` endpoint. The service layer filters `projectMilestones` and `grantMilestones` by status before returning.

2. **Frontend (gap-app-v2)**:
   - `getProjectUpdates()` service appends `?milestoneStatus=` query param when provided
   - `useProjectUpdates()` hook includes `milestoneStatus` in React Query cache key for proper cache isolation
   - `useProjectProfile()` and `UpdatesContent` pass the filter param through
   - `ActivityFeed` no longer does any milestone status filtering — it renders whatever items are passed
   - Fixed `window.location.pathname` SSR crash → `usePathname()` from next/navigation

## Why This Approach

Server-side filtering is the correct architecture: less data over the wire, business logic in the right layer, and simpler frontend state.

## Testing Steps

1. Navigate to a project's Updates tab
2. Select the "Milestones" filter pill
3. A milestone status dropdown appears (All / Pending / Completed / Verified)
4. Select "Completed" — URL updates to `?milestoneStatus=completed` and the feed shows only completed milestones
5. The API request includes `?milestoneStatus=completed` in the network tab
6. Refresh page — filter state is preserved from URL

## Related

- Backend PR: https://github.com/show-karma/gap-indexer/pull/1088

## Checklist

- [x] Server-side filtering implemented in gap-indexer
- [x] Frontend passes filter param to API
- [x] React Query cache key includes milestoneStatus
- [x] SSR crash (window.location.pathname) fixed
- [x] ActivityFeed simplified — no client-side status filtering
- [x] Tests updated to reflect server-side filtering architecture

## Summary by CodeRabbit

* **New Features**
  * Milestone status filter added to Updates with options: All, Pending, Completed, Verified.
  * Filter appears when Milestones are selected and syncs with the URL (persists across navigation); clearing filters removes the param.
  * Updates now respect milestone-status filtering server-side and surface an updating/loading state while changes apply.

* **Tests**
  * Added comprehensive tests covering UI behavior, URL sync, service logic, and hook/loading interactions.

### DIFF
--- a/components/Pages/Project/v2/Content/UpdatesContent.tsx
+++ b/components/Pages/Project/v2/Content/UpdatesContent.tsx
@@ -73,6 +73,11 @@ export function UpdatesContent({ className }: UpdatesContentProps) {
       }
       params.delete("sort");
 
+      // Clear milestone status when milestones tab is no longer active
+      if (!newFilters.includes("milestones")) {
+        params.delete("milestoneStatus");
+      }
+
       const newURL = params.toString() ? `?${params.toString()}` : pathname;
       router.replace(newURL, { scroll: false });
     },
--- a/components/Pages/Project/v2/MainContent/ActivityFilters.tsx
+++ b/components/Pages/Project/v2/MainContent/ActivityFilters.tsx
@@ -72,12 +72,11 @@ export function ActivityFilters({
 
   const totalCount = Object.values(counts).reduce((sum, c) => sum + (c || 0), 0);
 
-  // Show milestone status filter when milestones are visible:
-  // either no filter is active (All) or milestones filter is explicitly active
+  // Show milestone status filter only when the Milestones pill is explicitly active
   const showMilestoneStatusFilter =
     milestoneStatusFilter !== undefined &&
     onMilestoneStatusChange !== undefined &&
-    (activeFilters.length === 0 || activeFilters.includes("milestones"));
+    activeFilters.includes("milestones");
 
   return (
     <div className={cn("flex flex-col gap-4", className)} data-testid="activity-filters">


amaurymagalhaes:
## PR Review

**Overall:** The feature works and has solid test coverage (32 tests), but there's a critical architectural flaw — all filtering is done client-side on fully-fetched data — and a few real bugs worth fixing before merge.

Score: **5/10** — Do not merge as-is.

---

### Critical Issues (P0/P1)

**1. [P1] All filtering is client-side on the full dataset — backend should own this**

`services/milestone-status-filter.service.ts` + `ActivityFeed.tsx:293`

```ts
// ActivityFeed.tsx — runs after allUpdates is already fully fetched
filtered = filterByMilestoneStatus(filtered, milestoneStatusFilter);
```

`allUpdates` is the **entire** project feed fetched from the indexer. The frontend fetches everything and then discards most of it on the client. As project history grows, this gets increasingly wasteful.

The indexer already knows milestone status (it indexes the `verified[]` attestations and `completed` flags). The correct fix is to pass `?milestoneStatus=completed` to the API and let the backend filter. This PR should include a backend endpoint change, or at minimum be flagged as a known tech-debt shortcut with a follow-up ticket.

---

**2. [P1] `window.location.pathname` — SSR crash**

`UpdatesContent.tsx:84`

```ts
const newURL = params.toString() ? `?${params.toString()}` : window.location.pathname;
```

`window` is not available during server-side rendering. Next.js App Router components can render on the server. This will throw `ReferenceError: window is not defined` in SSR/ISR contexts.

**Fix:**
```ts
import { usePathname } from "next/navigation";
// ...
const pathname = usePathname();
// ...
const newURL = params.toString() ? `?${params.toString()}` : pathname;
```

---

### Other Issues (P2/P3)

**3. [P2] `"impact"` type in `MILESTONE_TYPES` — untested, likely wrong**

`milestone-status-filter.service.ts:47`

```ts
const MILESTONE_TYPES: ReadonlySet<string> = new Set(["milestone", "grant", "impact"]);
```

None of the 248 lines of tests cover `"impact"` type. It either silently filters impact items incorrectly or is dead code. Check the `UnifiedMilestone["type"]` union — if `"impact"` isn't a real type, remove it.

---

**4. [P2] `completed !== false` check is too broad**

`milestone-status-filter.service.ts:76`

```ts
if (item.completed !== false) {
  return "completed";
}
```

This returns `"completed"` for any truthy value including `0`, `""`, or unexpected shapes. Given `completed` can be `false | object`, the check should be:

```ts
if (item.completed !== null && item.completed !== false) {
  return "completed";
}
```

Or more explicitly:
```ts
if (typeof item.completed === "object" && item.completed !== null) {
  return "completed";
}
```

---

**5. [P2] Re-exporting a type from a UI component**

`ActivityFilters.tsx:12`

```ts
export type { MilestoneStatusFilter } from "@/services/milestone-status-filter.service";
```

This creates a confusing dual source of truth. Consumers are already importing directly from the service in every test file. This re-export from a UI component is backwards — remove it, consumers should import from the service.

---

**6. [P3] Non-milestone items always pass through status filter — surprising UX**

`milestone-status-filter.service.ts:97`

```ts
// Non-milestone items pass through (null status)
if (itemStatus === null) {
  return true;
}
```

When a user selects "Pending", they see pending milestones **plus all grant updates, endorsements, etc.** The UX intent is unclear. If the milestone status filter only applies when the "Milestones" pill is active, this might be fine — but when "All" is active, filtering by milestone status and still seeing non-milestone updates is confusing. Needs a product decision and a test that explicitly asserts this is intentional.

---

### Summary of required fixes before merge

1. Fix `window.location.pathname` → `usePathname()` (SSR crash)
2. Audit `"impact"` type inclusion
3. Harden the `completed !== false` check
4. Remove the re-export from `ActivityFilters.tsx`
5. Track the client-side-filtering architectural debt as a follow-up issue

PR #1381, opened 2026-05-10
fix: suppress project impact 404 noise

maheshmurthy:
## Summary
- treat 404 responses from the project impacts endpoint as an expected empty state instead of reporting them as frontend API errors
- add focused unit coverage for success, 404, and non-404 impact fetch behavior

## Testing
- pnpm exec vitest run --project unit __tests__/unit/services/project-impacts.service.test.ts
- pnpm exec biome check services/project-impacts.service.ts __tests__/unit/services/project-impacts.service.test.ts
- pnpm run test *(baseline failure reproduced on refreshed `origin/main`: unrelated auth/cart/project-content suites already fail in this environment, including `components/Pages/Project/v2/__tests__/ProjectMainContent.test.tsx` and auth/e2e-localstorage tests)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for project impact queries to appropriately distinguish between missing projects and other system errors, improving reliability and enabling more accurate error reporting.

* **Tests**
  * Added unit test coverage for project impact retrieval to validate proper handling of successful queries, missing projects, and various error conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/show-karma/gap-app-v2/pull/1381)

### DIFF
--- a/__tests__/unit/services/project-impacts.service.test.ts
+++ b/__tests__/unit/services/project-impacts.service.test.ts
@@ -1,0 +1,82 @@
+/**
+ * @file Tests for project-impacts.service
+ * @description Tests project impacts fetch behavior and error reporting
+ */
+
+vi.mock("@/utilities/enviromentVars", () => ({
+  envVars: {
+    NEXT_PUBLIC_GAP_INDEXER_URL: "http://localhost:4000",
+  },
+}));
+
+vi.mock("@/components/Utilities/errorManager", () => ({
+  errorManager: vi.fn(),
+}));
+
+vi.mock("@/utilities/fetchData");
+
+import { errorManager } from "@/components/Utilities/errorManager";
+import { getProjectImpacts } from "@/services/project-impacts.service";
+import fetchData from "@/utilities/fetchData";
+
+const mockFetchData = fetchData as vi.MockedFunction<typeof fetchData>;
+const mockErrorManager = errorManager as vi.MockedFunction<typeof errorManager>;
+
+describe("project-impacts.service", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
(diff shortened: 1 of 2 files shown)


amaurymagalhaes:
Closing without merging — this PR does not fix DEV-244 and is superseded.

**Root-cause mismatch.** DEV-244 / Sentry `7470550140` is `AxiosError: Network Error`, which is a request that never received an HTTP response (offline, DNS, CORS preflight, ad-blocker, TLS reset, navigation abort). It is not a 404. In `utilities/fetchData.ts`, a Network Error hits the `!err.response` branch and returns `status = 500`, so the `if (status === 404) return []` guard added here would never trigger for the reported error.

**Already fixed.** The exact "AxiosError: Network Error" family was resolved by DEV-236 (commit 2f9332927) with a centralized, multi-layered fix that also covers the impact page:
- `utilities/sentry/transientErrors.ts` — `isTransientNetworkError()`
- `instrumentation-client.ts` — global Sentry `beforeSend` drop (defence in depth)
- `components/Utilities/errorManager.tsx` — drop before `captureException`
- `utilities/queries/defaultOptions.ts` — React Query retries (≤2)

`getProjectImpacts` routes errors through `errorManager`, and `fetchData` returns the raw AxiosError for no-response failures, so the impact-page Network Error is correctly suppressed and retried.

**Also stale.** This branch edits the old single-arg `getProjectImpacts`; `main` now has the `options`/`isAuthorized`/`signal` signature, so the diff no longer applies cleanly.

If we want 404-as-empty as a separate intentional improvement, it should be a new ticket rebased onto the current signature, with a guard so it doesn't mask genuine project-not-found errors.

PR #1385, opened 2026-05-10
fix: guard invalid project route params

maheshmurthy:
## Summary
- reject malformed project route params like `/project/-nan-12` before the SSR path tries to fetch project data
- treat obviously invalid slugs as not-found so the project page fails closed instead of surfacing an SSR crash
- add focused coverage for the project route param guard

## Testing
- `pnpm exec vitest run --project unit utilities/__tests__/getProjectCachedData.test.ts`
- `pnpm exec biome check utilities/queries/getProjectCachedData.ts utilities/__tests__/getProjectCachedData.test.ts`
- `pnpm run test` *(fails on current `origin/main` baseline too in unrelated donation cart, auth/localStorage, and project content suites; also reproduces Vitest worker `EPIPE` instability in this Hermes environment)*

Linear: DEV-242

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for project data caching functionality, covering valid project access and proper error handling for invalid or malformed project identifiers.

* **Bug Fixes**
  * Strengthened validation of project route parameters with explicit format verification to detect and reject invalid identifiers, ensuring users receive appropriate error pages when accessing projects with malformed parameters.

[Review Change Stack](https://app.coderabbit.ai/change-stack/show-karma/gap-app-v2/pull/1385)

### DIFF
--- a/utilities/__tests__/getProjectCachedData.test.ts
+++ b/utilities/__tests__/getProjectCachedData.test.ts
@@ -1,0 +1,63 @@
+import type { Project } from "@/types/v2/project";
+
+const mockGetProject = vi.fn();
+const mockNotFound = vi.fn(() => {
+  throw new Error("NEXT_NOT_FOUND");
+});
+const mockRedirect = vi.fn();
+
+vi.mock("@/services/project.service", () => ({
+  getProject: (...args: unknown[]) => mockGetProject(...args),
+}));
+
+vi.mock("next/navigation", () => ({
+  notFound: () => mockNotFound(),
+  redirect: (...args: unknown[]) => mockRedirect(...args),
+}));
+
+function createMockProject(overrides: Partial<Project> = {}): Project {
+  return {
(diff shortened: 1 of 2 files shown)


Arthh:
Closing in favor of #1406.

The Linear ticket title mentions `/project/-nan-12`, which suggested a malformed-slug bug, but the actual Sentry events tell a different story. The latest event on `GAP-FRONTEND-1YJ` (today on `/project/anu-in`, 554+ occurrences total) comes from valid project slugs returning a null impacts payload. Specifically, the error message is `Project Impacts API Error: null`, meaning `error` is null AND `data` is null — both legitimate "no impacts" signals that the service was misclassifying as a backend failure.

The slug regex guard in this PR isn't wrong defensively, but Next.js routing and the indexer already 404 on clearly malformed slugs, so it won't reduce the 554 events from valid slugs. #1406 fixes the actual source at `services/project-impacts.service.ts:38` by splitting the guard so null/non-array responses resolve to `[]` silently while real errors still report.

PR #1387, opened 2026-05-10
fix: require signer before community admin check

maheshmurthy:
## Summary
- gate `useCheckCommunityAdmin` on an available signer before the admin query can run
- keep the hook returning `false` until signer initialization finishes instead of calling the SDK with `undefined`
- add a regression test for the no-signer startup race on the grant funding route

## Testing
- `pnpm exec vitest run --project unit __tests__/unit/hooks/useCheckCommunityAdmin.test.tsx`
- `pnpm run test` *(fails here and on refreshed `origin/main` because of existing unrelated baseline failures in donation cart/auth/RBAC/UI tests)*

### DIFF
--- a/__tests__/unit/hooks/useCheckCommunityAdmin.test.tsx
+++ b/__tests__/unit/hooks/useCheckCommunityAdmin.test.tsx
@@ -272,6 +272,17 @@ describe("useCheckCommunityAdmin", () => {
       expect(mockIsCommunityAdminOf).not.toHaveBeenCalled();
     });
 
+    it("should not fetch when signer is not ready", () => {
+      mockUseSigner.mockReturnValue(undefined as ReturnType<typeof useSigner>);
+
+      const { result } = renderHook(() => useCheckCommunityAdmin(mockCommunity), {
+        wrapper: createWrapper(queryClient),
+      });
+
+      expect(result.current.isLoading).toBe(false);
+      expect(mockIsCommunityAdminOf).not.toHaveBeenCalled();
+    });
+
     it("should fetch when all conditions are met", async () => {
       mockIsCommunityAdminOf.mockResolvedValueOnce(true);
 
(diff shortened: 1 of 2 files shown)


amaurymagalhaes:
Closing as **superseded** — DEV-261's root cause is already fixed on `main`.

### Root cause
The funding route runs `useCheckCommunityAdmin → isCommunityAdminOf → GAP.getCommunityResolver(signer, getGapRpcConfig())` **without a chainID**. In the SDK, `currentChainId = chainId || Number((await signer.provider.getNetwork())?.chainId || ...)` — with no `chainId` arg it dereferences `signer.provider`, and when an authenticated user's signer hasn't initialized yet, `signer` is `undefined` → `TypeError: can't access property "provider", e is undefined`. It's caught by `isCommunityAdminOf`, but `errorManager` re-reports it to Sentry, hence the persistent noise.

### Why this PR isn't the right fix
- **Treats the symptom, not the cause.** The same call also throws ethers `NETWORK_ERROR` when the wallet is on a different chain than the community. Gating on `!!signer` doesn't fix that.
- **Introduces a regression.** `isAdmin` is a read-only call that doesn't need a signer. Gating the query on `!!signer` re-couples a read to wallet-signer readiness, so authenticated users whose signer is slow/absent (embedded/Privy wallets, session restore) get `isAdmin = false` and lose admin UI on the funding page.

### Already fixed on main
Commit `2086b07e0` ("fix(communities): pass chainID to getCommunityResolver to avoid ethers NETWORK_ERROR") changes the call to `getCommunityResolver(signer, getGapRpcConfig(), chainID)`. With `chainID` passed, `currentChainId = chainID` short-circuits (never touches `signer.provider`) and the resolver is built from a static RPC provider — fixing both the TypeError and the cross-chain error, with no signer needed. This commit is on `main` but not in this branch; merging this PR on top would re-add a signer gate the fix deliberately removed.

Closing in favor of `2086b07e0`.

PR #1388, opened 2026-05-10
fix: fetch public program details without auth

maheshmurthy:
## Summary
- make the public program detail hook fetch funding program configs without auth bootstrap
- add a focused hook test covering the unauthenticated fetch path

## Testing
- `pnpm exec vitest run --project unit __tests__/features/programs/hooks/use-program.test.tsx`
- `pnpm run test` *(fails on this branch and on refreshed `origin/main` due to existing baseline issues in unrelated suites, including donation cart storage/localStorage failures and unrelated component expectation mismatches)*

## Risks
- public program detail fetches now skip auth headers, matching the existing public apply route for the same endpoint

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for the program retrieval hook to verify proper data fetching and loading state management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/show-karma/gap-app-v2/pull/1388)

### DIFF
--- a/__tests__/features/programs/hooks/use-program.test.tsx
+++ b/__tests__/features/programs/hooks/use-program.test.tsx
@@ -1,0 +1,62 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { renderHook, waitFor } from "@testing-library/react";
+import type React from "react";
+
+vi.mock("@/utilities/fetchData", () => ({
+  __esModule: true,
+  default: vi.fn(),
+}));
+
+import { useProgram } from "@/src/features/programs/hooks/use-program";
+import fetchData from "@/utilities/fetchData";
+
+const buildClient = () =>
+  new QueryClient({
+    defaultOptions: {
+      queries: { retry: false, gcTime: 0, staleTime: 0 },
+    },
+  });
+
+const wrapper =
+  (queryClient: QueryClient) =>
+  ({ children }: { children: React.ReactNode }) => (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+  );
+
+describe("useProgram", () => {
+  let queryClient: QueryClient;
+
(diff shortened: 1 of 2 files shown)


brunodmsi:
Closing without merge — diagnosis is wrong and the fix regresses admin functionality.

## Findings

The indexer route \`GET /v2/funding-program-configs/:programId\` does not require auth. It is registered with \`preHandler: [optionalAuthentication()]\` (\`app/modules/v2/api/routes/funding-program/funding-program.routes.ts:120-142\`), and its OpenAPI description says: _"Authenticated admins/reviewers receive full config including accessCode."_

That means today:
- Anonymous visitors already get the public response (no auth needed).
- Logged-in admins/reviewers get extra fields (e.g., \`accessCode\`).

Flipping \`isAuthorized\` from \`true\` → \`false\` in \`use-program.ts\` does not "make the call public" — the indexer was already public. It strips the token, so **logged-in admins viewing this route lose access to admin-only fields**. That's a real regression.

## On the original Sentry signal

\`TokenManager.getToken()\` (\`utilities/auth/token-manager.ts:96-142\`) catches Privy SDK errors internally and resolves to \`null\` for anonymous visitors. It does not reject. The captured \`{ code, data, message }\` shape in Sentry is most likely an unhandled rejection from Privy's own internals (e.g., \`getAccessToken()\` against \`auth.privy.io\` with stale session cookies), not from this fetch.

## Suggested next steps (not in this PR)

1. If the Privy rejection reproduces, fix it at the source — wrap the Privy bootstrap in \`.catch()\` and log without rethrowing.
2. Leave the public/admin contract intact: anonymous gets public, authenticated gets admin extras.
3. If \`fetchData\`'s default \`isAuthorized=true\` is genuinely error-prone, that's a separate refactor (invert the default or add a \`publicFetch()\` helper).

PR #1389, opened 2026-05-10
fix: defer privy boot on public apply pages

maheshmurthy:
## Summary
- avoid idle-loading Privy for anonymous visitors on public apply routes
- request Privy only when the user explicitly triggers login, then retry login after the SDK finishes loading
- add focused tests for the public apply route guard and deferred login flow

## Testing
- `pnpm exec biome check components/Utilities/PrivyProviderWrapper.tsx hooks/useAuth.ts __tests__/unit/components/Utilities/PrivyProviderWrapper.test.tsx __tests__/unit/hooks/useAuth.privy-load.test.tsx`
- `pnpm exec vitest run --project unit __tests__/unit/components/Utilities/PrivyProviderWrapper.test.tsx`
- `pnpm exec vitest run --project unit __tests__/unit/hooks/useAuth.privy-load.test.tsx`
- `pnpm run test` *(fails in this worktree and on refreshed `origin/main` with the same existing baseline suites, including `components/Pages/Project/v2/__tests__/ProjectMainContent.test.tsx` and the known localStorage-based auth suites such as `__tests__/unit/hooks/useAuth.test.tsx`)*

Linear: DEV-264

### DIFF
--- a/__tests__/unit/components/Utilities/PrivyProviderWrapper.test.tsx
+++ b/__tests__/unit/components/Utilities/PrivyProviderWrapper.test.tsx
@@ -1,5 +1,10 @@
 import { act, render } from "@testing-library/react";
-import React from "react";
+
+const mockUsePathname = vi.fn(() => "/");
+
+vi.mock("next/navigation", () => ({
+  usePathname: () => mockUsePathname(),
+}));
 
 // Mock heavy dependencies
 vi.mock("@tanstack/react-query", () => ({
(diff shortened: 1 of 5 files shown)


amaurymagalhaes:
**Closing in favor of #1498.** tl;dr:

- **Regression:** deferring Privy boot leaves the bridge `ready=false` forever on apply pages → navbar "Sign in" button never renders (infinite skeleton) and closed-program apply pages spin on "Checking permissions…" forever. Worse than the bug it targets.
- **Wrong tool:** DEV-264 is a transient third-party network failure (`Failed to fetch`, extension frame `frame_ant.js`), not an app bug. Deferral doesn't make it resilient — clicking "Sign in" still hits `auth.privy.io`.
- **Already fixed:** that error class is dropped in `beforeSend` via `isTransientNetworkError` (DEV-236), which landed after this PR and after the single Sentry event.

#1498 instead adds the missing regression coverage for `transientErrors.ts` locking in the suppression. Thanks for the investigation here — the anonymous-apply-flow insight was useful, just not the right fix for this signature.

PR #1390, opened 2026-05-11
fix(communities): pass chainID to getCommunityResolver to avoid ethers NETWORK_ERROR

amaurymagalhaes:
## Summary

Fixes Sentry [GAP-FRONTEND-1ZY](https://karma-crypto-inc.sentry.io/issues/7470168634/) — `Error: network changed: 1 => 10 (code=NETWORK_ERROR, version=6.11.0)` thrown from `getNetwork` while checking community admin status.

### Root cause

`isCommunityAdminOf` called `GAP.getCommunityResolver(signer, getGapRpcConfig())` without an explicit `chainId`. The SDK then falls back to probing `signer.provider.getNetwork()` — but the user's wallet may be on a different chain than the community (e.g. wallet on Mainnet `1`, community on Optimism `10`). Ethers v6 detects the mismatch and throws `NETWORK_ERROR`.

### Fix

Pass `community.chainID` as the third argument. Per `karma-gap-sdk/core/class/GAP.ts:470`, an explicit `chainId` + `rpcConfig` makes the SDK build a static `JsonRpcProvider` for the target chain and skip the signer's network probe entirely — correct behavior for this read-only check.

Other `getCommunityResolver` callers (`AddAdminDialog`, `RemoveAdminDialog`) are write operations where the user is already on the correct chain; they are unaffected.

## Test plan

- [x] New unit tests in `__tests__/utilities/sdk/communities/isCommunityAdmin.test.ts`:
  - asserts `getCommunityResolver` is called with `community.chainID`
  - happy path returns `resolver.isAdmin` result
  - returns `false` on NETWORK_ERROR (fallback path)
- [x] Verified red → green: test fails on previous code, passes with the fix
- [ ] Sentry issue stops occurring after deploy

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit test suite for community admin verification functionality to validate multiple scenarios and error handling
* **Bug Fixes**
  * Enhanced RPC configuration handling in community admin checks for improved reliability

[Review Change Stack](https://app.coderabbit.ai/change-stack/show-karma/gap-app-v2/pull/1390)

### DIFF
--- a/__tests__/utilities/sdk/communities/isCommunityAdmin.test.ts
+++ b/__tests__/utilities/sdk/communities/isCommunityAdmin.test.ts
@@ -17,7 +17,7 @@ vi.mock("@/components/Utilities/errorManager", () => ({
   errorManager: vi.fn(),
 }));
 
-const mockGetCommunityResolver = GAP.getCommunityResolver as unknown as vi.Mock;
+const mockGetCommunityResolver = vi.mocked(GAP.getCommunityResolver);
 
 const community = {
   uid: "0xcb67cd16cbdf4e9c3b6ed4c8f9424411a48be796d690750afc84f9288e7c7996",
--- a/__tests__/utilities/sdk/communities/isCommunityAdmin.test.ts
+++ b/__tests__/utilities/sdk/communities/isCommunityAdmin.test.ts
@@ -1,0 +1,69 @@
+import { GAP } from "@show-karma/karma-gap-sdk";
+import type { Community } from "@/types/v2/community";
+import { getGapRpcConfig } from "@/utilities/gapRpcConfig";
+import { isCommunityAdminOf } from "@/utilities/sdk/communities/isCommunityAdmin";
+
+vi.mock("@show-karma/karma-gap-sdk", () => ({
+  GAP: {
+    getCommunityResolver: vi.fn(),
+  },
+}));
+
+vi.mock("@/utilities/gapRpcConfig", () => ({
+  getGapRpcConfig: vi.fn(() => ({ 10: "https://rpc.optimism" })),
+}));
+
+vi.mock("@/components/Utilities/errorManager", () => ({
+  errorManager: vi.fn(),
+}));
+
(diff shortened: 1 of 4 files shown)


amaurymagalhaes:
### Codebase audit for the same root cause

Scanned all sites that pass a wallet signer to SDK contract calls to find other instances of the ethers v6 `NETWORK_ERROR: network changed: X => Y` pattern.

| Site | Status |
|------|--------|
| `utilities/sdk/communities/isCommunityAdmin.ts` | Fixed in this PR (first commit) |
| `utilities/sdk/projects/projectOwnershipTransfered.ts` | **Same bug** — fixed in commit 8c0f7155 |
| `components/Pages/Admin/AddAdminDialog.tsx` | Safe — write op, `setupChainAndWallet` switches chain first |
| `components/Pages/Admin/RemoveAdminDialog.tsx` | Safe — same |
| `hooks/useGrantCompletionRevoke.ts` | Safe — same |
| `utilities/gap/buildAttestationTransaction.ts` | Safe — write path, callers switch chain |
| `utilities/sdk/getContractOwner.ts` | Safe — caller (`useContractOwner`) builds a chain-specific `JsonRpcProvider`, not a wallet signer |

`isOwnershipTransfered` had the identical pattern: read-only check, signer + rpcConfig but no `chainId`. The current caller (`TransferOwnershipDialog`) happens to switch chain first, so it's a latent bug rather than an active one — but the fix is trivial and consistent. Added matching unit tests (chainID guard + happy path + NETWORK_ERROR fallback).

amaurymagalhaes:
### Babysit sign-off

- Self-review score: **98/100**
- Bot threads resolved: **1** (CodeRabbit — `vi.mocked()` typing)
- CI: **all checks green** ✅
- Includes a codebase audit that surfaced one additional latent instance of the same root cause (`projectOwnershipTransfered.ts`), fixed in commit 8c0f7155 with matching tests.

Ready for human review.

PR #1429, opened 2026-05-15
fix(profile): silence Authorization header errors on public project pages (DEV-256)

amaurymagalhaes:
## Summary

Fixes [DEV-256](https://linear.app/showkarma/issue/DEV-256) — recurring `Authorization header is required` Sentry errors on `/project/:projectId`.

This is the **client-side complement to #1427** (which fixed the SSR/prefetch path for DEV-257). #1427 already plumbed `isAuthorized` through the *services* and the *server prefetch*. This PR finishes the job by plumbing it through the *client hooks* and the public-profile component callers, and by hardening `fetchData` so this class of bug can't silently recur.

> This branch is stacked on top of `amaury/dev-257-fix-connection-closed-funding-page` (#1427). The diff below collapses to just the client-layer changes once #1427 merges.

## Root cause

When an anonymous visitor lands on `/project/:projectId` client-side:

1. The page mounts, hooks (`useProjectGrants`, `useProjectImpacts`, `useProjectUpdates`) fire client-side.
2. Each hook calls the service with `isAuthorized=true` (the default).
3. `fetchData` calls `TokenManager.getToken()`, which returns `null`.
4. `fetchData` proceeds to fire the request **without** the `Authorization` header.
5. The indexer responds with `"Authorization header is required"`, which lands in Sentry.

## Changes

- **Hooks**: `useProjectGrants`, `useProjectImpacts`, `useProjectUpdates`, `useProjectProfile` accept a `{ isAuthorized }` option. Default stays `true` (back-compat for authenticated callers).
- **Public-profile callers** (7 components — Activity, ProjectWrapper, Roadmap, AboutContentWrapper, FundingContentWrapper, UpdatesContent, ProjectProfileLayout, ProjectActivityChart) pass `{ isAuthorized: false }` so anonymous visitors hit the public-fetch path the backend already supports (V2 grants/updates use `optionalAuthentication`; V1 impacts is public).
- **`fetchData` hardening**: when `isAuthorized=true` and `TokenManager.getToken()` returns null, short-circuit with a synthetic `[null, "AUTH_REQUIRED_NO_TOKEN", null, 401]` instead of firing a header-less request. Fails loudly at the network boundary rather than silently 401'ing at the backend. Exports `AUTH_REQUIRED_NO_TOKEN_ERROR` for callers.
- **Sentry filter**: drop both `"Authorization header is required"` (in case SDK callers built outside `fetchData` hit the path) and the `AUTH_REQUIRED_NO_TOKEN` sentinel.

[...]

PR #1428, opened 2026-05-15
fix(funding): silence transient Network Error noise and add retry UI

amaurymagalhaes:
## Overview

Fixes [DEV-236](https://linear.app/showkarma/issue/DEV-236) / Sentry `GAP-FRONTEND-13P` — recurring `AxiosError: Network Error` on the project funding page (923 events, 49 users).

## Problem

The Sentry signature `AxiosError: Network Error` originates from `XMLHttpRequest.onerror` — the request never received an HTTP response. Causes are unactionable from a Sentry event (offline blip, CORS preflight, ad-blocker, route abort) and the captured stack contains only minified Axios bundle frames.

Compounding the noise:

- Project services (`getProjectGrants/Updates/Impacts`) explicitly call `errorManager()` on every fetch failure, which calls `Sentry.captureException` — so every transient hiccup logs an event even though React Query retries.
- React Query retried network errors **only once**, then gave up.
- `fetchData` had no timeout for unauthenticated requests — hung connections never failed.
- `FundingContentWrapper` rendered an indefinite loading spinner on error: no error state, no retry CTA. Violates the repo Pre-PR checklist rule "every data-fetching component handles loading, empty, AND error states — never return null".

## Solution

1. **`utilities/sentry/transientErrors.ts`** *(new)* — `isTransientNetworkError()` / `isAxiosAbortError()` helpers detecting axios errors with no HTTP response (`ERR_NETWORK`, timeouts, `"Failed to fetch"`, `"Load failed"`, aborts).
2. **`components/Utilities/errorManager.tsx`** — early-return for transient errors so Sentry stops getting spammed. This is the root cause of the noise.
3. **`instrumentation-client.ts`** — `beforeSend` filter as defense in depth for paths that bypass `errorManager` (raw React errors, third-party SDK fetches).
4. **`utilities/queries/defaultOptions.ts`** — retry transient errors up to **2×** with exponential backoff capped at 5 s; never retry abort/401/429.
5. **`utilities/fetchData.ts`** — 30 s default timeout for unauthenticated requests so hung connections fail fast and the retry policy can kick in. Authenticated requests keep their longer ceiling for legacy long-poll endpoints.
6. **`components/Pages/Project/v2/Content/FundingContentWrapper.tsx`** — three explicit states (loading skeleton / error with Try again CTA / success).

## Why This Approach

[...]

PR #1436, opened 2026-05-15
fix(gasless): use EOA's current on-chain nonce for EIP-7702 authorization

brunodmsi:
Linear: [DEV-300](https://linear.app/showkarma/issue/DEV-300/fix-eip-7702-nonce-mismatch-breaking-gasless-attestations)

## Summary

Every gasless attestation after the first one per chain was failing for embedded-wallet (email/Google/Farcaster) users with:

> `Invalid EIP-7702 authorization: The nonce does not match the userOperation sender address`

`utilities/gasless/providers/zerodev.provider.ts` was hardcoding `nonce: 0` when signing the EIP-7702 authorization tuple. Bundlers validate that the tuple's `nonce` equals the authority EOA's current on-chain transaction count, so this only worked on the very first delegation per chain (when the count happens to be 0) and silently broke for every call after.

The bug landed in `eb9342202` (2025-12-22) and was preserved through the strategy-pattern refactor in `d974055ef` (2026-01-07). Every chain configured with `useEIP7702: true` is affected — Optimism, Arbitrum, Base, Mainnet, Polygon, Scroll, Sei, plus all testnets. The Alchemy/Celo path is unaffected because the Alchemy SDK fetches the nonce internally.

## Changes

- `zerodev.provider.ts`: read the EOA's current transaction count from the public client and pass it as the authorization `nonce`.
- `privy-signer.ts`: throw on a missing `nonce` instead of silently defaulting to 0, so the same class of bug can't surface through another caller (Alchemy SDK, future provider, etc).
- Tests updated: replaced the "should default to 0" assertion with a "should throw" assertion, added a non-zero passthrough test, and added a regression test on `ZeroDevProvider.createClient` that asserts the fetched nonce flows into `signAuthorization`.

## Test plan

- [x] `pnpm test __tests__/unit/utilities/gasless/` — 7 files, 148 tests pass
- [x] `pnpm test __tests__/unit/utilities/gasless/privy-signer.test.ts` — 21/21 pass
- [x] `pnpm test __tests__/unit/utilities/gasless/providers.test.ts` — 21/21 pass (including the new regression test)
- [ ] Manual: complete a milestone twice in a row on Optimism with an embedded wallet — first call should still succeed, second call should now succeed (previously failed with the bundler error)
- [ ] Manual: same flow on Base, to confirm the per-chain nonce is fetched per request

## Summary by CodeRabbit

* **Bug Fixes**
  * EIP-7702 authorization nonce is now fetched from on-chain pending state and used for signing, preventing nonce mismatches and bundler rejections.
  * Signing requires an explicit nonce and will error if omitted.

* **Tests**
  * Expanded tests to assert explicit-nonce behavior, preserve provided nonces (including large/multibyte values), ensure payloads change when nonce changes, verify nonce is fetched before signing and re-fetched per call, and confirm failures propagate when nonce fetch fails.

[Review Change Stack](https://app.coderabbit.ai/change-stack/show-karma/gap-app-v2/pull/1436?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/__tests__/unit/utilities/gasless/privy-signer.test.ts
+++ b/__tests__/unit/utilities/gasless/privy-signer.test.ts
@@ -273,7 +273,23 @@ describe("Privy Signer", () => {
       expect(result.chainId).toBe(42220);
     });
 
-    it("should use default nonce of 0", async () => {
+    it("should throw if nonce is not provided", async () => {
+      // A missing nonce used to silently default to 0, which is only
+      // valid on the EOA's very first EIP-7702 delegation and was
+      // rejected by the bundler on every subsequent call. The signer
+      // now requires the caller to supply the authority's current
+      // on-chain nonce explicitly.
+      const signer = await createPrivySignerForGasless(mockEmbeddedWallet, 10);
+
+      await expect(
+        signer.signAuthorization!({
+          contractAddress: "0xabcdef1234567890abcdef1234567890abcdef12" as `0x${string}`,
+          chainId: 10,
+        })
+      ).rejects.toThrow("EIP-7702 signAuthorization requires an explicit nonce");
+    });
+
+    it("should preserve a non-zero nonce", async () => {
       const mockSignature =
         "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef1b";
       mockProvider.request.mockResolvedValue(mockSignature);
@@ -282,10 +298,10 @@ describe("Privy Signer", () => {
       const result = await signer.signAuthorization!({
         contractAddress: "0xabcdef1234567890abcdef1234567890abcdef12" as `0x${string}`,
         chainId: 10,
-        // nonce not specified, should default to 0
+        nonce: 7,
       });
 
-      expect(result.nonce).toBe(0);
+      expect(result.nonce).toBe(7);
     });
 
     it("should correctly parse signature components for v=27", async () => {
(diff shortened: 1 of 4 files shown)


brunodmsi:
## Smoke check analysis

**Result:** 82 passed · 1 failed · 2 flaky (passed on retry) — exit code 1.

### The one hard failure

`e2e/tests/smoke/dashboard-pages.smoke.spec.ts:235` — **`T-DASH-08: communities listing page loads`**. Failed all 3 retries.

```
expect(hasCommunityContent).toBeTruthy()
Received: false
```

The test navigates to `/communities`, mocks only `**/v2/communities/stats**`, then waits up to 8s for *any* element matching `/communit/i` (heading or text). Nothing matched. The page either didn't render community-related copy in time, or another network call (not in the mock) blocked rendering.

### The two flaky ones (passed on retry, not blocking)

- `donation-pages.smoke.spec.ts:111` — `T-DON-04`: empty state when no programs exist
- `donation-pages.smoke.spec.ts:127` — `T-DON-05`: auto-redirect when only one program exists

### Why this is not caused by this PR

This PR's diff is scoped to:
- `utilities/gasless/providers/zerodev.provider.ts`
- `utilities/gasless/utils/privy-signer.ts`
- Two unit-test files in `__tests__/unit/utilities/gasless/`

The failing routes (`/communities`, `/donate/*`) do not import any gasless module. No code path between this diff and those pages.

Corroborating signals from the same CI run:
- ✅ `build` — bundle compiles (4m33s)
- ✅ `test (1)` — full unit suite (8m19s)
- ✅ `quality-gate`, `static-checks`, `react-doctor`, `qa-execution (auth)`, `qa-pipeline`
- Smoke passed on the last three merged PRs (#1428, #1429, #1432) and the smoke surface has not changed on this branch

### Recommendation

Re-run smoke. If `T-DASH-08` persists across reruns it's a real (separate) regression on `/communities` that warrants its own ticket — but it's orthogonal to the EIP-7702 nonce fix in this PR.

PR #1439, opened 2026-05-15
fix: gate milestone edit/revoke on on-chain attester/recipient

brunodmsi:
## Summary

Editing a milestone on `/project/forest---an-efficient-and-lightweight-filecoin-node` while connected as a team member (`0x857D…BA63`, labelled "Owner" in the UI) reverted with `GAP:Not owner.` during gas estimation. Two cooperating defects:

1. The Edit button was rendered for any "authorized" address (project admin, super admin, contract owner). The on-chain revoke gate is far stricter.
2. The Team Members list rendered every member as "Owner" because the role check used `ProjectResolver.isOwner`, which is deliberately permissive on-chain.

Both are fixed here.

### Bug 1 — wrong gate

`Gap.sol::_multiRevoke` requires:

```solidity
require(
  revoker == owner() ||           // Gap contract owner (Karma deployer)
  target.attester == revoker ||   // attestation's original attester
  target.recipient == revoker,    // attestation's recipient
  "GAP:Not owner."
);
```

So a user can revoke a milestone iff they are the Karma super-admin, the milestone's attester, or its recipient. `Project.isOwner` / `Project.isAdmin` from the SDK are not equivalent to this check — they read a different (and intentionally permissive) function on `ProjectResolver`.

For the Forest milestone (`0xac1805…`):
- attester: `0x23b7…` (Karma signer)
- recipient: `0xb471…05a1` (project owner)
- connected wallet: `0x857D…BA63` — matches neither → revert.

### Bug 2 — role mislabelling

`ProjectResolver.isOwner(uid, addr)` returns `true` for *any* of: the explicit projectOwner mapping, the contract `_owner` (Karma deployer), OR `projectAdmins[uid][addr]`. Admins therefore read as "Owner". The previous `getProjectMemberRoles` mirrored this contract behaviour and labelled every member as "Owner".

Direct on-chain probe on Base confirmed this (`projectAdmins` raw mapping):

| Address | projectOwner mapping | projectAdmins | Should be |
|---|---|---|---|
| 0xb471…05a1 | ✓ | false | Owner |
| 0x741e…0a84 | — | **true** | Admin |
| 0x857d…BA63 | — | **true** | Admin |

### Edit/Revoke gate

- New pure helper `utilities/milestones/canEditMilestone.ts`. Mirrors `Gap.sol::_multiRevoke`: `isContractOwner || addr == milestone.recipient || addr == milestone.attester`. No hook, no async — three values combined with `||`.
- `ActivityCard.tsx` computes `canEdit` for milestone-type activities and threads it through `ActivityCard/MilestoneCard.tsx`.
- Edit pencil, Revoke-completion button, and `GrantMilestoneSimpleOptionsMenu`'s Edit affordance are now gated on `canEdit`. Mark Complete keeps its current `isAuthorized` gate (different on-chain path). Delete keeps its off-chain revoke fallback.
- `convertToUnifiedMilestones` (`hooks/v2/useProjectUpdates.ts`) now plumbs both `attester` (real) and `recipient` (real) onto `UnifiedMilestoneSource.{projectMilestone,grantMilestone.milestone}`. Previously `attester` was misleadingly populated from `milestone.recipient`.

### Team list labelling

`utilities/getProjectMemberRoles.ts` rewrite:

- Owner = `ProjectResolver.projectOwner(uid)` on-chain, falling back to indexer `project.owner` only when the mapping is `0x0`.
- Admin = `ProjectResolver.projectAdmins(uid, addr)` raw mapping (not `isAdmin`).
- Member = everyone else.
- When the resolver is unreachable (chain not configured or RPC missing), fall back to labelling `project.owner` as Owner and omit per-member admin status; no silent role guesses for the other members.

## Tests

15 new unit tests:

- `canEditMilestone.test.ts` (9) — contract-owner bypass, recipient/attester matching for grant and project milestones, missing inputs, case-insensitive match.
- `getProjectMemberRoles.test.ts` (6) — Owner from on-chain mapping, Admin from `projectAdmins`, plain Member, zero-address fallback to indexer, missing RPC, "never two Owners" regression.

All 39 pre-existing tests in `ActivityCard.test.tsx` and `useProjectUpdates.test.ts` still pass.

## Test plan

- [ ] Connect as the Forest project owner (`0xb471…05a1`) → Edit visible on grant milestones; Revoke Completion visible on completed milestones.
- [ ] Connect as Kai (`0x857D…BA63`) or another team member → Edit and Revoke Completion hidden. Mark Complete and Delete still visible.
- [ ] Connect as a Karma super admin → Edit visible regardless of project role.
- [ ] Forest Team Members panel shows exactly one Owner, others as Admin or Member based on `projectAdmins` mapping.
- [ ] Switch to a project on a different chain (Optimism / Celo) and confirm the Team list still renders correctly (resolver address resolution by chain).

## Out of scope / follow-ups

- Delete-grant (`multiGrantDelete`) hits the same `Gap.sol::_multiRevoke` gate but has a silent off-chain revoke fallback, so the failure isn't user-visible. Could be gated identically — tracked separately.
- SDK's `Project.isOwner` is misleading because of the on-chain conflation. Consider renaming to `canAct` or deprecating in a follow-up RFC.

### DIFF
--- a/__tests__/components/Milestone/GrantMilestoneOptionsMenu.test.tsx
+++ b/__tests__/components/Milestone/GrantMilestoneOptionsMenu.test.tsx
@@ -77,13 +77,14 @@ describe("GrantMilestoneOptionsMenu", () => {
     vi.clearAllMocks();
   });
 
-  it("shows Edit option for pending milestones", async () => {
+  it("shows Edit option for pending milestones when canEdit is true", async () => {
     const user = userEvent.setup();
     render(
       <GrantMilestoneOptionsMenu
         milestone={mockPendingMilestone}
         completeFn={mockCompleteFn}
         alreadyCompleted={false}
+        canEdit={true}
       />
     );
 
@@ -95,6 +96,24 @@ describe("GrantMilestoneOptionsMenu", () => {
     expect(screen.getByText("Delete")).toBeInTheDocument();
   });
 
+  it("hides Edit option when canEdit is false (non-owner)", async () => {
+    const user = userEvent.setup();
+    render(
+      <GrantMilestoneOptionsMenu
+        milestone={mockPendingMilestone}
+        completeFn={mockCompleteFn}
+        alreadyCompleted={false}
+        canEdit={false}
+      />
+    );
+
+    await user.click(screen.getByRole("button", { name: "Open milestone actions" }));
+
+    expect(screen.queryByText("Edit")).not.toBeInTheDocument();
+    expect(screen.getByText("Mark as Complete")).toBeInTheDocument();
+    expect(screen.getByText("Delete")).toBeInTheDocument();
+  });
+
   it("hides Edit option for completed milestones", async () => {
     const user = userEvent.setup();
     render(
@@ -102,6 +121,7 @@ describe("GrantMilestoneOptionsMenu", () => {
         milestone={mockCompletedMilestone}
         completeFn={mockCompleteFn}
         alreadyCompleted={true}
+        canEdit={true}
       />
     );
 
@@ -118,6 +138,7 @@ describe("GrantMilestoneOptionsMenu", () => {
         milestone={mockPendingMilestone}
         completeFn={mockCompleteFn}
         alreadyCompleted={false}
+        canEdit={true}
       />
     );
 
@@ -135,6 +156,7 @@ describe("GrantMilestoneOptionsMenu", () => {
         milestone={mockCompletedMilestone}
         completeFn={mockCompleteFn}
         alreadyCompleted={true}
+        canEdit={true}
       />
     );
 
@@ -150,6 +172,7 @@ describe("GrantMilestoneOptionsMenu", () => {
         milestone={mockPendingMilestone}
         completeFn={mockCompleteFn}
         alreadyCompleted={false}
+        canEdit={true}
       />
     );
 
@@ -163,6 +186,7 @@ describe("GrantMilestoneOptionsMenu", () => {
         milestone={mockPendingMilestone}
         completeFn={mockCompleteFn}
         alreadyCompleted={false}
+        canEdit={true}
       />
     );
 
--- a/__tests__/components/Milestone/MilestoneCard.test.tsx
+++ b/__tests__/components/Milestone/MilestoneCard.test.tsx
@@ -174,27 +174,27 @@ describe("MilestoneCard", () => {
 
   describe("Rendering - Basic Elements", () => {
     it("should render milestone card with title", () => {
-      render(<MilestoneCard milestone={mockProjectMilestone} isAuthorized={false} />);
+      render(<MilestoneCard milestone={mockProjectMilestone} isAuthorized={false} canEdit={false} />);
 
       expect(screen.getByText("Test Project Milestone")).toBeInTheDocument();
     });
 
     it("should render milestone description", () => {
-      render(<MilestoneCard milestone={mockProjectMilestone} isAuthorized={false} />);
+      render(<MilestoneCard milestone={mockProjectMilestone} isAuthorized={false} canEdit={false} />);
 
       const description = screen.getByTestId("read-more");
       expect(description).toHaveTextContent("This is a test milestone description");
     });
 
     it("should render milestone badge", () => {
-      render(<MilestoneCard milestone={mockProjectMilestone} isAuthorized={false} />);
+      render(<MilestoneCard milestone={mockProjectMilestone} isAuthorized={false} canEdit={false} />);
 
       expect(screen.getByText("Milestone")).toBeInTheDocument();
       expect(screen.getByAltText("Milestone")).toBeInTheDocument();
     });
 
     it("should render creation date and attester", () => {
-      render(<MilestoneCard milestone={mockProjectMilestone} isAuthorized={false} />);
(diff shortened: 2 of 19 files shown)


brunodmsi:
## Review response (commit 90a48d9)

### Addressed (blocking)

- **#1 (P1) — `canEdit` default of `false`**: Made `canEdit` a **required** prop on `GrantMilestoneOptionsMenu`, `GrantMilestoneSimpleOptionsMenu`, and both `MilestoneCard` variants. Threaded it through the older `components/Milestone/MilestoneCard.tsx` (so the orphan path doesn't silently hide Edit). Updated `GrantMilestoneOptionsMenu.test.tsx` to pass `canEdit={true}` in positive cases and added a `canEdit={false}` negative case.
- **#2 (P2) — completion attestation not in candidates**: `canEditMilestone` now also checks `source.{grantMilestone.milestone,projectMilestone}.completed?.attester`. Completion-attester divergence is rare in practice but the helper now matches the on-chain gate's "*target* attestation" semantics.
- **#3 (P2) — resolver-address-missing silently empties the map**: Restored the indexer-owner fallback. When no `projectResolver` is configured for a chain (or RPC URL is missing), we still label `project.owner` as "Owner" and omit per-member admin status rather than rendering a roleless Team list. New test added.
- **#8 (P3) — `_projectInstance` dead weight**: Removed the unused second parameter from `getProjectMemberRoles` and updated all six call sites (`TeamContent`, `TeamMemberCard`, `Team/index`, `Team/MemberCard`, `useMemberRoles`, `useMemberRoleChange`). Dropped two more dead locals (`_programId`, `_communityData`).

### Quality gate

The v2 `ActivityCard/MilestoneCard.tsx` is at **661 lines now (was 663 baseline, 674 in commit 1)**. Reverted the inline-pencil and revoke-completion conditional wrappers introduced in commit 1 — the primary dropdown surface (`GrantMilestoneSimpleOptionsMenu`) remains gated. Inline-pencil and revoke-completion tighter gating tracked as a follow-up below.

### Tracked as follow-ups (not blocking this PR)

- **#4 (P2) — silent demotion to "Member" on per-member RPC failure**: agree it's inconsistent with the empty-map policy. Will replace with `undefined` (omit the entry) in a follow-up.
- **#5 (P2) — `as any` casts in `convertToUnifiedMilestones`**: pre-existing pattern; the fix is to extend `GrantMilestoneWithDetails`/`ProjectMilestone` types with the optional `attester` field. Out of scope here.
- **#6 (P2) — cross-chain `isOwner` mismatch**: pre-existing. Worth a `useIsContractOwnerForChain(chainId)` follow-up.
- **#7 (P3) — comment density**: trimmed the JSDoc blocks on `MilestoneCard.tsx` to clear the size gate. Remaining longer blocks live in `canEditMilestone.ts` and `getProjectMemberRoles.ts` (the source-of-truth WHY commentary).
- **#9, #10 (P3) — test edge cases / `useAuth` mock**: ActivityCard test still passes; nothing was hitting a real hook chain. The existing test fixture mocks `useAuth` via `vitest.setup`.
- **Inline-pencil + revoke-completion buttons in `ActivityCard/MilestoneCard.tsx`**: still ungated, fail with the existing toast for non-owners. Same UX hole as the Delete-grant button. Will gate both in a follow-up that also raises the file-size budget for this file.

### Tests

102 passing across 6 files (new + previously-existing related): `canEditMilestone`, `getProjectMemberRoles`, `GrantMilestoneOptionsMenu`, `MilestoneCard` (orphan), `ActivityCard`, `useProjectUpdates`.

brunodmsi:
Thanks for the careful read.

**P3 (PR description blocker)**: fixed. The "returns empty map" line dated from commit 1; commit 90a48d9 restored the indexer-owner fallback per the prior review and I forgot to update the body. Now reads: *"When the resolver is unreachable (chain not configured or RPC missing), fall back to labelling `project.owner` as Owner and omit per-member admin status; no silent role guesses for the other members."*

**P2 — `canEditMilestone` accuracy depends on the API `recipient` field**: agreed and confirmed during local QA. When the public Base RPC briefly rate-limited my `projectAdmins` call for one member, my `catch → "Member"` mislabeled an actual admin. The fix is two-part: (1) extend `ProjectMilestone` / `GrantMilestoneWithDetails` with `attester` + `recipient` so the cast and the `milestoneAny.*` reads go away, (2) change my catch to omit the entry rather than guess "Member". Tracking as a follow-up — happy to bundle into a small follow-up PR right after this merges so this PR stays scoped.

**P2 — `completed?.recipient` not in candidates**: also agreed. Same `UnifiedMilestoneSource` type gap — `completed` exposes only `attester`, not `recipient`. Will include in the same follow-up PR.

**P3 — `as { attester?: string }` cast**: same root cause as P2#1; goes away when types are fixed.

**P3 — N RPC round-trips**: noted. Forest has 3 members so we're fine today. Multicall batching becomes worth it around 8–10 members; will flag for the SDK work.

**P3 — dead `_programId` / `_communityData`**: fair, scope creep. Their removal was load-bearing for clearing the file-size quality gate after adding `canEdit` to the props/destructure of an already-oversized file — could have been a separate cleanup commit, agree.

The SDK suggestion (`Project.getOwner()` / `Project.getAdmins()` exposing the raw mappings) is the right long-term shape and lines up with the "rename `isOwner` → `canAct` or deprecate" follow-up from the first review. Will draft an RFC.

Ready to merge from my side once you're satisfied.

PR #1498, opened 2026-05-22
test(sentry): lock in transient auth.privy.io 'Failed to fetch' suppression (DEV-264)

amaurymagalhaes:
## Summary

Closes DEV-264 (`TypeError: Failed to fetch (auth.privy.io)` on the program apply flow) and supersedes #1389.

The Sentry signature is an **unactionable transient third-party network failure** (offline, DNS, TLS reset, ad-blocker / privacy extension blocking the host — the event's frame is `frame_ant.js`, an extension). This error class is **already dropped** in `instrumentation-client.ts` `beforeSend` via `isTransientNetworkError`, which matches the `"failed to fetch"` message fragment. That filter (DEV-236) landed *after* both the single Sentry event and #1389 were created, so the issue is already handled on `main`.

`utilities/sentry/transientErrors.ts` had **zero** test coverage. This PR adds regression tests that lock in the behavior:
- `TypeError: Failed to fetch` (the DEV-264 signature) is treated as transient and suppressed
- real HTTP errors carrying a `response` are **not** suppressed (they stay actionable)
- existing code / abort / casing branches

## Why not #1389

#1389 deferred Privy boot on apply routes. That approach:
- leaves the Privy bridge `ready=false` forever on apply pages, so the navbar "Sign in" button never renders (infinite skeleton) and closed-program apply pages show an infinite "Checking permissions…" spinner — a regression worse than the original transient error
- is the wrong tool for a transient third-party fetch failure, and is redundant with the existing DEV-236 filter
- isn't resilient anyway (a user clicking "Sign in" still triggers the same `auth.privy.io` fetch)

## Testing

`pnpm exec vitest run --project unit __tests__/unit/utilities/sentry/transientErrors.test.ts` — 7 passing

### DIFF
--- a/__tests__/unit/utilities/sentry/transientErrors.test.ts
+++ b/__tests__/unit/utilities/sentry/transientErrors.test.ts
@@ -1,0 +1,50 @@
+import {
+  isAxiosAbortError,
+  isTransientNetworkError,
+} from "../../../../utilities/sentry/transientErrors";
+
+describe("isTransientNetworkError", () => {
+  it("suppresses the Privy 'Failed to fetch (auth.privy.io)' signature (DEV-264)", () => {
+    // The Privy SDK bootstrap fires a fetch to auth.privy.io. When the browser
+    // never gets a usable response (offline, DNS, TLS reset, ad-blocker /
+    // privacy extension blocking the host), it rejects with a bare
+    // `TypeError: Failed to fetch`. Sentry annotates the title with the host
+    // (`(auth.privy.io)`) but the underlying message is just "Failed to fetch".
+    // This is unactionable third-party noise and must be dropped in beforeSend.
+    expect(isTransientNetworkError(new TypeError("Failed to fetch"))).toBe(true);
+  });
+
+  it("matches the 'failed to fetch' message regardless of casing", () => {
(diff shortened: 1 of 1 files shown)


amaurymagalhaes:
## Babysit sign-off ✅

This PR is healthy and ready to merge.

**Self-review score:** 96/100
- Tightly scoped: adds a 7-case regression suite (`__tests__/unit/utilities/sentry/transientErrors.test.ts`) locking in that the existing transient-network filter suppresses the Privy `auth.privy.io` "Failed to fetch" signature (DEV-264), while still letting real HTTP errors (responses with a status) through to Sentry. No production code changed — zero regression surface.

**Bot review threads resolved:** 0 (no actionable threads raised by CodeRabbit / karma-coding-ai).

**CI status:** all 23 checks green.

**Note on the earlier `quality-gate` failure:** it was unrelated to this diff. react-doctor's ruleset drifted mid-day on 2026-05-22 (warnings 10623→3185, errors 32→61 repo-wide), which the team addressed by pinning react-doctor and refreshing `quality-baseline.json` in #1499. This branch was based on a commit predating that refresh, so it carried the stale baseline. Merging the updated `main` brought in the pinned config + refreshed baseline, and the gate now compares 61→61 = pass.

PR #1505, opened 2026-05-25
fix(forms): upgrade @hookform/resolvers to v5 for zod v4 validation

amaurymagalhaes:
## Summary
Restore form validation under Zod v4 by upgrading `@hookform/resolvers` v4 → v5, and add regression coverage.

## Problem
The Zod v3→v4 migration broke validation on every form. Zod v4's `ZodError` exposes `.issues` instead of the old `.errors` alias, but `@hookform/resolvers@4.x`'s `zodResolver` checks `Array.isArray(error.errors)` and **re-throws** the `ZodError` when it's absent. React Hook Form doesn't catch that, so it surfaced as an **unhandled promise rejection**, crashing forms on any validation failure. This produced the production Sentry errors (GAP-FRONTEND-21Y, 21Z, 221–225): project creation, grant/milestone creation, and community forms.

## Solution
- Bump `@hookform/resolvers` `^4.0.0` → `^5.4.0` (v5 supports Zod v4).
- Add `utilities/zodResolver.ts` — a thin wrapper. v5 types the resolver as `Resolver<Input, Ctx, Output>`, splitting a schema's input/output types; for schemas using coercion/transforms/unions this no longer matches `useForm<z.infer<...>>`. The wrapper restores the v4 single-type behavior (documented; runtime identical).
- Repoint the 7 forms with input≠output schemas to the wrapper. The remaining forms keep the raw resolver, which works unchanged.
- Fix pre-existing a11y errors in `ReportConfigPage` (surfaced by the pre-commit hook on the touched file): group-heading `<label>`→`<span>`, segmented-control buttons to the `aria-pressed` toggle pattern.

## Testing steps
- `tsc --noEmit`: 0 errors.
- New resolver smoke test (wrapper + raw entry points) and validation-failure tests for ProjectDialog, CommunityDialog, ContributorProfileDialog, and the NewGrant milestone form — all green.
- Before/after proof: with the buggy `@hookform/resolvers@4.1.3`, 28 of 47 of the new tests fail; with `5.4.0`, all 47 pass.
- Manually verified in-browser (authenticated): submitting forms with invalid input renders field errors with zero `ZodError`/unhandled rejections.
- Added a global guard (`__tests__/setup.ts`) that fails any unit test leaking an unhandled rejection.

## Risk
Low. The only breaking surface of the bump is resolver typing, handled by the wrapper (0 type errors); valid submissions are unchanged at runtime. The `ReportConfigPage` a11y tweaks are behavior-preserving.

[...]

PR #1507, opened 2026-05-25
fix(profile): stop validating an incomplete contributor form on load

amaurymagalhaes:
## Summary
Hardens the source of the `name`-undefined crash (Sentry **GAP-FRONTEND-222/225**) so the globally-mounted contributor profile form never validates a partially-filled state. Builds on #1505 (the Zod v4 resolver fix, already merged).

## Problem
`ContributorProfileDialog` is rendered on every authenticated page (via `DeferredLayoutComponents`). When a contributor profile loaded, its effect populated the form with repeated `setValue(..., { shouldValidate: true })` calls **while `name` was still empty**, validating an incomplete form. That produced a `name: expected string, received undefined` ZodError on page load, surfaced as an **unhandled promise rejection** (Sentry 222/225, and the repo-wide `smoke` failure).

#1505 stopped the resolver from *throwing*; this removes the premature validation at the source so the form doesn't produce spurious errors on load either.

## Solution
- Populate the form in a single `reset({...})` (which does **not** validate) instead of six `setValue(..., { shouldValidate: true })` calls.
- Add `defaultValues` so the form is never in an undefined state.

## Testing steps
- New test: loading a profile with a missing `name` populates the form (asserts the GitHub field fills) **without** surfacing the "This name is too short" error — fails on the old per-field-validate code, passes now. The global unhandled-rejection guard also fails the test if the throw resurfaces.
- `ContributorProfileDialog.test.tsx`: 6/6 pass. `tsc --noEmit`: 0 errors.

## Risk
Low. `reset()` is the idiomatic way to load values into React Hook Form; valid-profile behavior is unchanged (happy-path test still passes).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation error messages displaying incorrectly when loading contributor profile information.

[Review Change Stack](https://app.coderabbit.ai/change-stack/show-karma/gap-app-v2/pull/1507?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/__tests__/components/Dialogs/ContributorProfileDialog.test.tsx
+++ b/__tests__/components/Dialogs/ContributorProfileDialog.test.tsx
@@ -114,8 +114,9 @@ vi.mock("@/hooks/useAuth", () => ({
 }));
 
 const mockRefetchProfile = vi.fn().mockResolvedValue({ data: undefined });
+let mockProfileValue: { data?: Record<string, unknown> } | undefined;
 vi.mock("@/hooks/useContributorProfile", () => ({
-  useContributorProfile: () => ({ profile: undefined, refetch: mockRefetchProfile }),
+  useContributorProfile: () => ({ profile: mockProfileValue, refetch: mockRefetchProfile }),
 }));
 
 const mockRefetchTeamProfiles = vi.fn();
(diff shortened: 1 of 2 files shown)


amaurymagalhaes:
### Babysit summary

**Self-review:** complete. `tsc --noEmit` 0 errors; `ContributorProfileDialog.test.tsx` 6/6 pass (incl. new regression test for loading a name-less profile without premature validation).

**Bot reviews:** CodeRabbit — no findings. No unresolved review threads.

**CI: all checks green**, including `smoke` — which now passes because #1505 (merged) deployed to staging and resolved the underlying `name`-undefined crash (Sentry GAP-FRONTEND-222/225). This PR removes the premature validation at its source as defense-in-depth.

PR #1520, opened 2026-05-26
fix(rpc): derive chain RPC URLs from a single Alchemy key

brunodmsi:
Community admins (observed on Celo) intermittently saw the **create-program / add-funding** UI gated, reporting they were "no longer admin" — though the wallet is still enlisted on-chain.

## Root cause
The frontend admin gate (`isCommunityAdminOf`) reads the on-chain resolver over the Celo RPC, and on any error returns `false` (admin treated as guest). That RPC came from `NEXT_PUBLIC_RPC_CELO`, which had the **Alchemy key embedded directly in the URL**. When the key was rotated/expired, every Celo read failed (401/403/429) and the gate silently closed. Separately, `NEXT_PUBLIC_ALCHEMY_KEY` was declared in the env schema but never wired to anything.

## Fix
- `getRPCUrlByChainId` is now the single source of truth and derives each chain's URL from the one `NEXT_PUBLIC_ALCHEMY_KEY` via an `ALCHEMY_SUBDOMAIN_BY_CHAIN_ID` data table (mirrors the existing `DEFAULT_RPC_URLS_BY_CHAIN_ID`). Precedence: **Alchemy key → per-chain `NEXT_PUBLIC_RPC_*` override → public default.** Alchemy wins so rotating the one key takes effect without having to clear a stale per-chain URL.
- The viem client path (`getOrCreateClient`) routes through the same resolver — one place to reason about.
- Wires `NEXT_PUBLIC_ALCHEMY_KEY` into `envVars` (was dead config).
- No chain-specific branches: chains Alchemy doesn't serve (Lisk, Sei) are just absent from the table and keep their existing URLs.

## Testing
- `rpcClient.test.ts`: 16 pass — Alchemy preference over a configured URL, malformed-key rejection, unmapped-chain pass-through.
- 47 admin-check tests (`isCommunityAdminOf`, `useCheckCommunityAdmin`, `useIsCommunityAdmin`) green; Biome + `tsc` clean.

## Companion / rollout
- Backend companion: show-karma/gap-indexer#1907 (routes the server-side admin check through a multi-RPC FallbackProvider).
- Ops: set `NEXT_PUBLIC_ALCHEMY_KEY` on Vercel (not the now-unused embedded `NEXT_PUBLIC_RPC_CELO`). Note this switches **all** Alchemy-mapped chains to the shared key — confirm those networks are enabled on the Alchemy app, or trim the table to Celo only.

> Pre-existing, unrelated: `pnpm run build` currently fails on `@stripe/crypto` / `agentation` module resolution; `tsc --noEmit` and the unit suite pass.

### DIFF
--- a/__tests__/unit/utilities/rpcClient.test.ts
+++ b/__tests__/unit/utilities/rpcClient.test.ts
@@ -181,5 +181,61 @@ describe("rpcClient", () => {
 
       expect(url).toBe("https://optimism.llamarpc.com");
     });
+
+    it("prefers the Alchemy URL over a configured per-chain URL when the key is set", async () => {
+      vi.resetModules();
+      vi.doMock("@/utilities/enviromentVars", () => ({
+        envVars: {
+          ALCHEMY_KEY: "abc123-_KEY",
+          RPC: { CELO: "https://stale-embedded-key.example.com" },
+        },
+      }));
+
+      const { getRPCUrlByChainId } = await import("@/utilities/rpcClient");
+
+      expect(getRPCUrlByChainId(42220)).toBe("https://celo-mainnet.g.alchemy.com/v2/abc123-_KEY");
+    });
+
+    it("keeps the configured URL for chains Alchemy does not serve even when the key is set", async () => {
+      vi.resetModules();
+      vi.doMock("@/utilities/enviromentVars", () => ({
+        envVars: {
+          ALCHEMY_KEY: "abc123",
+          RPC: { LISK: "https://rpc-lisk.example.com" },
+        },
+      }));
+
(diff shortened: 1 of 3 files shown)


brunodmsi:
Closing: investigation (Sentry, 37k events over 7d, errorId AUTH_COMMUNITY_001) confirmed the failure is exclusively in gap-indexer (project gap-indexer; zero events in gap-frontend/karma-frontend). The frontend NEXT_PUBLIC_RPC_CELO key is valid and the frontend admin check works. This RPC-key consolidation is unnecessary for the incident and risks overriding the working key. Backend fix tracked in show-karma/gap-indexer#1907.

PR #1537, opened 2026-05-29
fix(project): include project owner in Updates and Impact edit authorization

amaurymagalhaes:
## Problem

Project owners reported losing the ability to **edit and delete milestones** (and impact entries) on the project profile — the controls simply disappeared from the UI.

Root cause: several project views computed edit authorization as:

```ts
const isAuthorized = isOwner || isProjectAdmin;   // (or, in the legacy Activity view, isProjectAdmin alone)
```

This omits `isProjectOwner`. Note:
- `isOwner` (from `useOwnerStore`) is the **global EAS resolver super-admin**, not the project owner.
- `isProjectAdmin` is set **only** from the on-chain `isAdmin()` check — it has **no linked-wallet/API fallback**.
- `isProjectOwner` is the flag that *does* resolve ownership via `compareAllWallets` (covering email/embedded-wallet owners and multi-wallet accounts).

So a legitimate project owner who is not separately registered as an on-chain admin — e.g. a Privy email-login owner whose active wallet is the embedded wallet — resolved to `isAuthorized = false`, and all edit/delete controls silently stopped rendering.

This was a latent gap: sibling components already do it correctly (`GrantDetailLayout.tsx`, `TeamContent.tsx`); these views were missed by the earlier `isProjectOwner` authorization fix.

## Fix

Authorize on `isOwner || isProjectAdmin || isProjectOwner` across all affected project views:

- `components/Pages/Project/v2/Content/UpdatesContent.tsx`
- `components/Pages/Project/v2/MainContent/ImpactContent.tsx`
- `components/Milestone/MilestonesList.tsx`
- `components/Pages/Project/Roadmap/index.tsx`
- `components/Pages/Project/Impact/index.tsx`
- `components/Pages/Project/Impact/EmptyImpactScreen.tsx`
- `components/Pages/Project/Updates/Updates.tsx`
- `components/Pages/Project/ProjectWrapper/ProjectNavigation.tsx`
- `components/Pages/Project/Activity.tsx` (legacy — previously checked `isProjectAdmin` alone; now also includes `isOwner` + `isProjectOwner`)

## Tests

- Updated store mocks in the affected suites to include `isProjectOwner`.
- Added regression tests for the **project-owner-but-not-admin** case in `UpdatesContent` and `ImpactContent`.
- Affected unit suites pass (UpdatesContent, ImpactContent, MilestonesList, ProjectNavigation, ProjectNavigationTabs); typecheck passes.

## Out of scope

[...]

PR #1542, opened 2026-05-29
fix(auth): heal Privy/Wagmi hydration race for wallet display and data fetching

amaurymagalhaes:
## Problem

Privy and Wagmi initialize independently and asynchronously. On cold and returning page loads, `authenticated` flips `true` **before** `wallets[0].address` hydrates and before the access token is ready. Nothing coordinates a "auth is now ready → re-evaluate" signal, so three user-visible symptoms appear — all of which a manual page refresh "fixes":

1. **"No wallet connected" / blank avatar & name** for an actually-connected user — the navbar reads `address` while it's still `undefined`.
2. **Endpoints fail on first load, work after refresh** — authenticated requests fire before the token is ready and 401. `fetchData` swallows the 401 into a *successful* `[null, …]` tuple, so React Query treats it as data and never refetches.
3. The data that resolved empty during the gap never recovers because there's no refetch trigger when auth becomes ready.

## Fix

Three targeted changes, one per symptom:

**`src/components/navbar/navbar-user-menu.tsx`** — when authenticated but identity hasn't resolved yet (no farcaster/email and no address), render `NavbarUserSkeleton` instead of a blank avatar and "No wallet connected". A wallet user always resolves an address and a social user always resolves a social identity, so this only covers the transient hydration window.

**`utilities/fetchData.ts`** — retry an authorized indexer request **once** on a 401, after clearing the token cache and fetching a fresh token. This self-heals requests that raced the token instead of dead-ending as empty data until a refresh. Non-401s, public (`isAuthorized=false`) requests, and the case where no fresh token is available are unaffected.

**`hooks/useAuth.ts`** — add an auth-ready refetch barrier: when the wallet `address` first transitions `undefined → defined` while authenticated, call `queryClient.invalidateQueries()` once so queries that resolved empty/401 during the gap refetch with auth now ready.

## Tests

- `fetchData`: 401 retry-with-fresh-token success; no retry when no fresh token; no retry for public requests.
- `navbar-user-menu`: shows skeleton (not "No wallet connected") while an authenticated user's wallet is still hydrating.
- `useAuth`: invalidates queries once when the address resolves after auth; does not invalidate when the address is already present at mount.
- All affected suites pass (fetchData, navbar-user-menu, useAuth = 83 tests); `tsc --noEmit` passes.

## Notes / out of scope

- This is independent of the project-owner authorization fix (PR #1537) and branches from `main`.
- Per-account caveat for the original Tower Exchange report still stands separately: whether `isProjectOwner` resolves depends on the owner address `0x0036…86ca` being linked to that Privy account.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented incomplete user menu UI during the wallet/auth hydration gap.
  * Trigger query refetch when wallet details become available after authentication.
  * Added one-time token-refresh on authorized 401 responses to recover failed requests.

* **Tests**
  * Added unit tests covering wallet hydration behavior, the auth-ready refetch barrier, and 401 retry scenarios for authorized requests.

### DIFF
--- a/__tests__/navbar/unit/navbar-user-menu.test.tsx
+++ b/__tests__/navbar/unit/navbar-user-menu.test.tsx
@@ -629,4 +629,27 @@ describe("NavbarUserMenu", () => {
       expect(followSection).toBeInTheDocument();
     });
   });
+
+  describe("Wallet Hydration (Privy↔Wagmi gap)", () => {
+    it("shows the skeleton instead of 'No wallet connected' while an authenticated user's wallet is still hydrating", () => {
+      // authenticated=true, but `user` and `wallets[0].address` haven't resolved
+      // yet (the deferred-SDK race). The menu must not render a blank avatar or
+      // claim "No wallet connected" for an actually-connected user.
+      _localRefs.navPermsState.current.isLoggedIn = true;
+      _localRefs.navPermsState.current.address = undefined;
+      _localRefs.navPermsState.current.ready = true;
+      _localRefs.authState.current.authenticated = true;
+      _localRefs.authState.current.address = undefined;
+      _localRefs.authState.current.user = null;
+
+      renderWithProviders(<NavbarUserMenu />, {
+        mockUseAuth: createMockUseAuth(_localRefs.authState.current),
+      });
+
+      expect(screen.queryByText("No wallet connected")).not.toBeInTheDocument();
+      // The skeleton replaces the menu entirely — no menubar or avatar yet.
+      expect(screen.queryByRole("menubar")).not.toBeInTheDocument();
+      expect(screen.queryByRole("img")).not.toBeInTheDocument();
+    });
+  });
 });
--- a/__tests__/unit/hooks/useAuth.auth-ready-barrier.test.tsx
+++ b/__tests__/unit/hooks/useAuth.auth-ready-barrier.test.tsx
@@ -1,0 +1,169 @@
+/**
+ * @file Tests for the useAuth auth-ready refetch barrier
+ * @description Privy/Wagmi hydrate the wallet asynchronously, so `authenticated`
+ * flips true before `wallets[0].address` is populated. useAuth invalidates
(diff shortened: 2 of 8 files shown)


amaurymagalhaes:
## Branch health summary

- **Self-review:** passed (98/100). Addressed the one finding — removed `any` from the new `fetchData` 401-retry path (typed `unknown` narrowing).
- **Bot reviews:** CodeRabbit completed with **no actionable threads**; 0 open review threads.
- **Green checks:** `build`, `test (1)`, `smoke`, `static-checks`, `react-doctor`, `gate-check`, `scope`, `baseline-guard`, `qa-execution (public)`, `dogfood`, Vercel.
- **`quality-gate` ❌ — not caused by this PR.** The 4 flagged regressions (`Forms/MilestoneUpdate.tsx`, `Admin/ControlCenter/MilestonesSection.tsx`, and two `payout-disbursement` files) are pre-existing main-branch drift vs a stale baseline; none are in this PR's diff. The only over-limit file this PR touches (`useAuth.test.tsx`) was kept **net-zero** by moving the new tests into `useAuth.auth-ready-barrier.test.tsx`, so it no longer appears in the regression list.
- **`qa-execution (auth)`:** long-running/known-flaky shard, independent of this change; the `public` shard and unit tests pass.

Net: this PR introduces **zero** new quality regressions and all checks it influences are green.

amaurymagalhaes:
## ✅ Branch health — all green

- **Self-review:** 98/100.
- **Bot reviews:** CodeRabbit ✅, 0 unresolved threads (karma-coding-ai: none).
- **CI:** all checks pass — `build`, `test`, `smoke`, `static-checks`, `react-doctor`, `qa-execution (auth)`, `qa-execution (public)`, `qa-pipeline`, `quality-gate`, `baseline-guard`, `verdict`, Vercel.
- **mergeStateStatus:** CLEAN.

Note: the `quality-baseline` label authorizes the size-baseline refresh for 4 pre-existing oversized files that had drifted in main (none touched by this PR).

PR #1546, opened 2026-06-02
fix(auth): resolve identity from the linked wallet, not the first connected one

amaurymagalhaes:
## Overview

After logging in with MetaMask, logging out via the user dropdown, then logging in with email, the app kept showing the MetaMask avatar and pulled on-chain data from the MetaMask wallet — and the transaction signer could disagree with the displayed account. It only corrected itself after manually disconnecting in MetaMask.

## Problem

Privy's `useWallets()` returns **every** wallet physically connected in the browser. MetaMask in particular **cannot be disconnected programmatically** (Privy's `wallet.disconnect()` is a documented no-op for it), so a wallet from a previous session lingers in the list — often as `wallets[0]`.

The app derived identity from `wallets[0]`:
- `useAuth().address` → drove the navbar avatar, profile name, and on-chain/profile lookups.
- `PrivyWagmiProviders` synced `wallets[0]` into the outer wagmi config that `useAccount()` reads, so signing/attestation flows could use the stale wallet.
- `useClaimTransaction` signed the claim and keyed `claim-eligibility` on `wallets[0]` (a money path).

So an email/social user with a leftover MetaMask connection inherited the previous wallet's identity and on-chain data.

## Solution

New helper `selectPrimaryWallet(user, wallets)` (`utilities/auth/select-primary-wallet.ts`): prefer the connected wallet that is actually **linked to the authenticated user** (via the existing `compareAllWallets`, which covers standard/smart/embedded/farcaster/cross-app wallets and therefore **multiple linked wallets**), falling back to `wallets[0]` only before the user resolves or for login methods without a browser wallet.

Applied in three places so identity and signer never diverge:
- **`hooks/useAuth.ts`** — `primaryWallet`/`address` now come from the helper.
- **`components/Utilities/PrivyWagmiProviders.tsx`** — syncs the same selected wallet into the outer wagmi config; the sync is now **address-keyed** so it re-syncs when the selected wallet changes (e.g. the embedded wallet connecting after a stale MetaMask), instead of pinning to the first wallet for the session.
- **`src/features/claim-funds/hooks/use-claim-transaction.ts`** — signs and keys eligibility on the linked wallet; the mid-operation guard re-derives from live state.

Multiple linked wallets are handled: when all connected wallets belong to the user, selection keeps Privy's active wallet (`wallets[0]`) and does not reorder.

## Testing

- `__tests__/utilities/auth/select-primary-wallet.test.ts` — helper: stale-skip, multiple linked wallets, active-wallet follow, hydration fallback.
- `__tests__/unit/hooks/useAuth.test.tsx` — stale MetaMask → email, multi-wallet, hydration fallback.
- `__tests__/unit/features/claim-funds/use-claim-transaction-real.test.ts` — claim signs with the embedded (linked) wallet, never the stale MetaMask; eligibility keyed on the embedded address.

Manual QA (local, against staging API): MetaMask login → logout via dropdown → email login now shows the email account's avatar and data, not the lingering MetaMask.

### How to reproduce / verify manually
1. Log in with MetaMask.
2. Log out via the user dropdown (do **not** disconnect in MetaMask).
3. Log in with email.
4. The navbar should show your email account's avatar, and on-chain/profile data should be the email account's — not the MetaMask wallet.

## Release Notes

* **Bug Fixes**
  * Fixed wallet selection to prevent claims and transactions from using stale or unlinked wallets
  * Improved primary wallet resolution to correctly identify your authenticated linked wallet when multiple wallets are connected

* **Tests**
  * Added comprehensive test coverage for wallet selection behavior and authentication flows

### DIFF
--- a/__tests__/unit/features/claim-funds/use-claim-transaction-real.test.ts
+++ b/__tests__/unit/features/claim-funds/use-claim-transaction-real.test.ts
@@ -18,6 +18,7 @@ const {
   mockWriteContract,
   mockWalletClient,
   mockWallets,
+  mockUserRef,
   mockSanitizeErrorMessage,
   mockSwitchOrAddChain,
   mockWaitForTransactionReceipt,
@@ -31,8 +32,11 @@ const {
   const mockWalletClient = { writeContract: mockWriteContract };
   const mockWallets: Array<{
     address: string;
+    walletClientType?: string;
     getEthereumProvider: () => Promise<unknown>;
   }> = [];
+  // Privy user (with linkedAccounts) used by selectPrimaryWallet; null = anonymous.
+  const mockUserRef: { current: unknown } = { current: null };
   const mockSanitizeErrorMessage = vi.fn((err: unknown, defaultTitle: string) => ({
     title: defaultTitle,
     message: err instanceof Error ? err.message : "Unknown error",
@@ -45,6 +49,7 @@ const {
     mockWriteContract,
     mockWalletClient,
     mockWallets,
+    mockUserRef,
     mockSanitizeErrorMessage,
     mockSwitchOrAddChain,
     mockWaitForTransactionReceipt,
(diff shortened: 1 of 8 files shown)


amaurymagalhaes:
-

PR #1563, opened 2026-06-03
fix(sitemap): fail the build instead of shipping a degraded sitemap

amaurymagalhaes:
## Overview

The build regenerates the sitemap from the indexer's `/v2/sitemap` endpoint (`scripts/generate-sitemap.ts`, run in `production.yml`). When the indexer was slow or unreachable, an **established** kind (one with a published chunk floor) fell back to **empty child sitemaps** — silently dropping thousands of project/grant/milestone URLs from Google's index while the build still reported success. Build logs showed `projects/impacts/grants/milestones` timing out on attempts 1–2 of 3 on every deploy, riding the edge; one slower cold start ships the degraded artifact.

## Problem

The floor preserves the *count* of child chunks but emits any un-fetched chunk as an empty `<urlset>`. Child files aren't committed, so on a clean CI build there's no last-good content to fall back to — a cold/slow indexer silently produces empty children for whole kinds.

## Solution

- **Fail the build** (exit non-zero, nothing written) when any kind with a published floor comes back with an incomplete listing. The committed sitemap is left untouched, so the last good deploy keeps serving real URLs and the next deploy retries — strictly safer than deploying empties. (Orphaned chunks Google already crawled are still served as empty-200 by the existing `app/sitemaps/[kind]/sitemap/[chunk]/route.ts` fallback, so no 404s.)
- **Sequential per-kind fetch** instead of `Promise.all`, so the kinds don't all hit the indexer's cold aggregation at once — the contention behind the timeouts.
- **Per-page timeout 15s → 30s** for cold-cache headroom.

A brand-new kind with no floor still falls back to a single chunk (nothing to regress).

## Testing

- `__tests__/scripts/generate-sitemap.test.ts`:
  - New: build **fails and leaves the committed sitemap byte-for-byte untouched** when an established kind is unreachable.
  - Success-path cases now run against an in-process **fake indexer** instead of relying on the empty-children fallback.
- Validated manually (vitest spawns `npm`, which can't resolve on Windows locally; runs in CI):
  - Unreachable + committed 29-entry floor → **exit 1**, file unchanged, message lists all degraded kinds.
  - Fake indexer → **exit 0**, 29 entries, grants/1=1000 + grants/2 URLs, host rewritten to canonical, sequential fetch.
- `pnpm tsc --noEmit` and `biome check` clean.

## Pairs with

show-karma/gap-indexer `fix/sitemap-cold-cache-warmer` — that PR removes the root cause (the indexer's cold cache) via a scheduled warmer; this PR is the seatbelt that guarantees a degraded sitemap can never deploy even if a fetch still fails.

## Note

This restores reliable sitemap **discovery** (~25k URLs, no regressions). It does not by itself raise the **indexed** page count — that ceiling is the ~25k "Crawled – currently not indexed" pages (content quality), a separate track.

### DIFF
--- a/__tests__/scripts/generate-sitemap.test.ts
+++ b/__tests__/scripts/generate-sitemap.test.ts
@@ -37,6 +37,67 @@ function locCount(file: string): number {
   return (fs.readFileSync(file, "utf-8").match(/<loc>/g) ?? []).length;
 }
 
+// In-process indexer stub. By default every kind returns a single short page
+// (3 URLs) so the generator sees a complete listing for all kinds; pass
+// `pageSizes` to override per-kind pagination (e.g. grants spanning 3 chunks).
+function startFakeIndexer(
+  pageSizes: Record<string, number[]> = {}
+): Promise<{ url: string; close: () => Promise<void> }> {
+  const server = http.createServer((req, res) => {
+    const url = new URL(req.url ?? "", "http://localhost");
+    if (url.pathname === "/v2/sitemap") {
+      const kind = url.searchParams.get("kind") ?? "";
+      const page = Number(url.searchParams.get("page") ?? "1");
+      const size = (pageSizes[kind] ?? [3])[page - 1] ?? 0;
+      const urls = Array.from(
+        { length: size },
+        (_, i) => `https://staging.karmahq.xyz/x/${kind}/${page}/${i}`
+      );
+      res.writeHead(200, { "Content-Type": "application/json" });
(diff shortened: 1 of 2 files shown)


amaurymagalhaes:
Superseded by #1566 (serve sitemaps dynamically from the indexer). That approach removes build-time sitemap generation entirely, so there's no degraded build artifact left to guard against — the fail-safe this PR added is no longer needed.

PR #1566, opened 2026-06-03
feat(sitemap): serve sitemaps dynamically from the indexer (ISR + SWR)

amaurymagalhaes:
## Overview

Alternative to the build-time-generation fixes (#1563 / gap-indexer #1978). Instead of regenerating the sitemap at build time and caching it on the backend, the **frontend owns freshness**: dynamic route handlers fetch the indexer on demand and cache the responses in Next's Data Cache with **stale-while-revalidate**.

- Crawler traffic triggers the background refresh — **no cron, no KV**.
- A slow or unreachable indexer keeps serving the **last good** cached response instead of an empty one, so a degraded sitemap can never be deployed.
- The index is sized from `/v2/sitemap/counts` on every (cached) refresh, so it **grows automatically** — no committed floor to maintain.
- Removes build-time generation entirely (no `generate-sitemap` step, no committed `public/sitemap*.xml`).

The expensive part (the Mongo aggregation that derives the URL list) stays in the indexer — the frontend only fetches and caches.

## Changes

- `app/sitemap-index.xml/route.ts` + `app/sitemap.xml/route.ts` — dynamic index sized from counts (alias serves identical content).
- `app/sitemaps/[kind]/sitemap/[chunk]/route.ts` — fetches the chunk's URLs from `/v2/sitemap`, canonicalizes the host, empty-but-valid 200 past the end. (`static` + `communities` stay as their existing local-data sitemap routes.)
- `utilities/sitemap.ts` — shared XML builders, chunk math, host canonicalization, and cached indexer fetchers (24h `revalidate`).
- Drop `generate-sitemap` from the build; delete the script, its test, and the committed `public/sitemap*.xml`.

## Testing

- 24 unit tests (helpers + both route handlers, `fetch` stubbed). `pnpm tsc --noEmit` + biome clean.
- **Smoke-tested in `next dev`** against a fake indexer (counts: projects 2500, grants 1500, funding-programs 77, impacts/milestones 0):
  - `/sitemap-index.xml` → 10 entries, correct chunk sizing (projects ×3, grants ×2, others ×1 + static + communities); `/sitemap.xml` alias byte-matches.
  - children: projects/1=1000, projects/3=500, grants/2=500 URLs; host rewritten `staging`→`www.karmahq.xyz`; bogus kind → 404.

## Trade-off / open question

The one cost vs. the backend approach: a **true cold miss** (first crawl after a deploy, when Next's Data Cache is purged) still hits the indexer live. If the indexer's own Redis is also cold at that moment, that first request is slow (one request, not user-facing). The gap-indexer cache warmer (#1978) is the cheap complementary insurance for that — so a sensible combo is **this PR + keep #1978, close #1563**. Decision needed.

Supersedes #1563 (no build generation to make fail-safe).

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for sitemap routes and utilities, including validation of XML responses, URL canonicalization, and chunk handling.

* **Refactor**
  * Migrated sitemap generation from build-time to dynamic runtime serving, enabling real-time sitemap updates without requiring rebuilds.

* **Chores**
  * Updated build script to remove pre-generation step for sitemaps.


amaurymagalhaes:
## Verified against a production build (not just `next dev`)

`next build` + production server, indexer pointed at a controllable fake:

- **Build is decoupled from the indexer** — build succeeds with the indexer **down**, makes **0 indexer calls**, and all three sitemap routes are **ƒ Dynamic**.
- **Data Cache works** — 3 requests to a route → **1** indexer fetch.
- **Stale-while-revalidate serves last-good** — with the indexer returning 503, the route's revalidation hits it and fails, but every response stays **HTTP 200 with the full last-good index** (never 500/empty). Recovers automatically when the indexer returns.
- **Cold-miss limitation is clean** — no cached value + indexer down → one **HTTP 500** (Google retries), server stays healthy, serves 200 the moment the indexer recovers.
- Real staging-indexer response shapes match; `staging.karmahq.xyz` hosts rewritten to `www`.

### Bug found and fixed by this testing
`/sitemap.xml` is a reserved Next metadata path and was being **statically prerendered at build time** — calling the indexer during the build and baking a failed response if it was down. Fixed in d5e13d86 with `export const dynamic = "force-dynamic"` on both index routes. The `next dev` smoke test could not have caught this; the production build did.

amaurymagalhaes:
## Babysit status

**Self-review:** ~99/100. Found and fixed one issue during review — an unbounded child chunk number (`/sitemaps/projects/sitemap/99999999999999999.xml`) parsed to an unsafe integer and triggered a pointless indexer fetch; now capped at `MAX_SITEMAP_CHUNK` with a 404 (90ca4d87). Added tests for the bound and for a partial `/counts` payload.

**Local gates:** 26 unit tests pass, `tsc --noEmit` clean, biome clean.

**CI:** all checks green **except `quality-gate`**, which is a **pre-existing, repo-wide failure unrelated to this PR**:
- The failure is `[vitest] No "useRouter" export is defined on the "next/navigation" mock` in `ActivityFeed.tsx → useIsCommunityAdmin → useAuth` — files this PR does not touch.
- It only fails in the full-suite coverage run; the sharded `test (1)` / `test (2)` both pass.
- `quality-gate` is currently failing on multiple branches (incl. `test/fix-failing-baseline-and-flakiness`, the branch already addressing it). Not introduced here and not fixable from this PR without touching out-of-scope shared test infra.

**Bot review:** no actionable threads. CodeRabbit is account-rate-limited (couldn't review); karma-coding-ai's QA pipeline passed ("No critical/high issues found").

amaurymagalhaes:
## Babysit update — regression fixed; remaining red is pre-existing

**Fixed in this PR:**
- `MAX_SITEMAP_CHUNK` bound + 404 guard for absurd chunk numbers (90ca4d87).
- A pre-existing test, `orphaned-chunk-route.test.ts`, covered the old static empty-200 route behavior I rewrote and was hitting the network — updated to stub the indexer and keep its GSC-no-404 intent (e1919769).

**Verified:** all sitemap tests pass (`orphaned-chunk-route`, `sitemap-routes`, `utilities/sitemap`); `build`, `test (1)`, `test (2)`, `static-checks`, `smoke`, `react-doctor`, `claude`, CodeRabbit, Vercel — all green.

**`quality-gate` is red for reasons unrelated to this PR.** Its full-suite coverage run has **37 failing test files**, none touched by this PR — e.g. `ActivityFeed`, `GrantOverview`, `AdminSection.msw`, `CommunityStatCards.msw`, `BrowseApplicationsClient`, integration/auth, mutations, smoke pages. This is the repo's known baseline flakiness (same failure set on other branches, incl. `test/fix-failing-baseline-and-flakiness`). The one sitemap entry in the list, `static/sitemap.test.ts`, is also pre-existing (expects a `/dashboard` entry the code never had; fails identically in isolation; not in this diff).

This PR cannot turn `quality-gate` green without fixing the repo-wide test baseline, which is out of scope here and owned by the dedicated flakiness branch.

amaurymagalhaes:
## Babysit — final on `quality-gate`

Investigated whether this PR can make `quality-gate` green. It can't, and here's the proof:

- The 37 failing files in the coverage run are the repo's known baseline flakiness — none touched by this PR. They're the exact set being fixed on `test/fix-failing-baseline-and-flakiness` (`a2f99ab4`, 43 files).
- **That dedicated fix branch itself still fails `quality-gate`** (latest run: failure). So even the comprehensive baseline fix doesn't make the gate pass yet.
- Its shared `setup.ts` change is locale-pinning for *local-dev* determinism (its own comment: tests "fail locally while staying green on CI") — not the cause of the CI failures.

**Conclusion:** `quality-gate` is currently red repo-wide and is not fixable from this PR without importing in-flight, non-passing team work. Everything that reflects this PR's own code is green (build, test (1)/(2), static-checks, smoke, react-doctor, claude, CodeRabbit, Vercel, dogfood/qa). Recommend merging on those, or rebasing once the baseline fix lands on main.

PR #1590, opened 2026-06-08
fix: serialize gasless UserOperations per sender to prevent AA25 nonce collisions

brunodmsi:
## Problem

Sentry [GAP-FRONTEND-22W](https://karma-crypto-inc.sentry.io/issues/7536781530/): deleting a milestone failed with

```
AA25 invalid account nonce: Another UserOperation with same sender and nonce is already being processed
```

ERC-4337 smart accounts advance their nonce only once a UserOperation is *included*. Firing two operations for the same sender before the first is mined makes both read the same nonce, so the bundler rejects the second with `AA25`.

### How the race happens
- The only in-flight guard is `isDeleting` in `useMilestone()`, surfaced as the delete dialog's `disabled` button.
- But `useMilestone()` is instantiated **per milestone row**, so each row has its own `isDeleting`. The grant in the event had 28 milestones — deleting two rows in quick succession fires two UserOps from the one wallet.
- A fresh smart-account client is built **per transaction** (`getAttestationSigner` → `createGaslessClient`), so no per-client lock can span operations.
- `multiGrantUndoCompletion` / `completeMilestone` / `multiGrantEditCompletion` have no in-flight guard at all, so a delete racing any other on-chain action collides the same way.

A per-component guard can't fix this — the nonce is **global to the wallet**, the guard is **local to a row**. The fix has to live where UserOps are submitted.

## Solution

A module-level queue keyed by smart-account sender address that serializes `eth_sendTransaction` calls. Because it lives at module scope, it spans the per-transaction clients and any unrelated component that fires its own tx. This covers **every** gasless flow (delete, complete, edit, create), not just delete.

- **`utilities/gasless/utils/userOpQueue.ts`** *(new)* — `serializeBySender(sender, task)` (promise-chain queue; a failed op delays but doesn't poison the next) and `withUserOpSerialization(provider, sender)` (a Proxy that funnels only `eth_sendTransaction` through the queue, passing all other methods through untouched).
- **`providers/zerodev.provider.ts`** — wrap the vendor `KernelEIP1193Provider` with `withUserOpSerialization`. This is the Base/production path that threw.
- **`providers/alchemy.provider.ts`** — wrap the `sendUserOperation` + `waitForUserOperationTransaction` block in `serializeBySender`.

## Testing

- `__tests__/unit/utilities/gasless/userOpQueue.test.ts` *(new, 6 cases)*: same-sender ordering, cross-sender independence, error propagation, queue-survives-rejection, and the wrapper serializing only sends.
- New tests pass; all 135 existing gasless unit tests pass; Biome clean.

### Manual
1. Log in with an email/Google (embedded-wallet) account on a gasless chain (Base).
2. On a grant with several on-chain milestones, click delete on two rows in quick succession.
3. Before the fix: the second submission could fail with `AA25`. After: the second op waits for the first to be included, then proceeds.

## Notes
- Scope is per-tab. A user with two tabs could still theoretically collide; an `AA25`-catch-and-retry could be layered into the same queue later if needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gasless transaction reliability by serializing user operations per account, preventing nonce collisions and bundler rejections during concurrent transactions.
* **New Features**
  * Providers now serialize outgoing transaction requests per sender to ensure sequential submission.
* **Tests**
  * Added unit tests covering per-sender serialization behavior, error handling, and non-transaction passthrough.

### DIFF
--- a/__tests__/unit/utilities/gasless/userOpQueue.test.ts
+++ b/__tests__/unit/utilities/gasless/userOpQueue.test.ts
@@ -1,0 +1,130 @@
+import { serializeBySender, withUserOpSerialization } from "@/utilities/gasless/utils/userOpQueue";
+
+const SENDER_A = "0xAAAA000000000000000000000000000000000000";
+const SENDER_B = "0xBBBB000000000000000000000000000000000000";
+
+const deferred = <T>() => {
+  let resolve!: (value: T) => void;
+  let reject!: (reason?: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve, reject };
+};
+
+describe("serializeBySender", () => {
+  it("runs ops for the same sender one at a time, in order", async () => {
+    const events: string[] = [];
+    const first = deferred<void>();
+
+    const op1 = serializeBySender(SENDER_A, async () => {
+      events.push("1:start");
+      await first.promise;
+      events.push("1:end");
+    });
+
+    const op2 = serializeBySender(SENDER_A, async () => {
+      events.push("2:start");
+      events.push("2:end");
+    });
+
+    // Op2 must not begin until op1 resolves, even though both were queued.
+    await Promise.resolve();
+    expect(events).toEqual(["1:start"]);
+
+    first.resolve();
+    await Promise.all([op1, op2]);
+
+    expect(events).toEqual(["1:start", "1:end", "2:start", "2:end"]);
+  });
+
+  it("does not block a different sender", async () => {
+    const events: string[] = [];
+    const blockA = deferred<void>();
+
+    const opA = serializeBySender(SENDER_A, async () => {
+      events.push("A:start");
+      await blockA.promise;
+      events.push("A:end");
+    });
+
+    const opB = serializeBySender(SENDER_B, async () => {
+      events.push("B:done");
+    });
+
+    await opB;
+    // B finished while A is still pending — senders are independent.
+    expect(events).toEqual(["A:start", "B:done"]);
+
+    blockA.resolve();
+    await opA;
+    expect(events).toEqual(["A:start", "B:done", "A:end"]);
+  });
+
+  it("propagates the task's error to the caller", async () => {
+    const boom = new Error("AA25");
+    await expect(
+      serializeBySender(SENDER_A, async () => {
+        throw boom;
+      })
+    ).rejects.toBe(boom);
+  });
+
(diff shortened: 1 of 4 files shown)


brunodmsi:
@coderabbitai Thanks for the review. Addressed:

1. **Undefined-sender test** — added (`serializes ops with an undefined sender under the shared fallback key`).
2. **Distinct key per undefined sender** — intentionally keeping the shared `"unknown"` key. When `client.account?.address` is unreadable it's almost always the *same* in-session wallet, so serializing those ops together is what prevents the AA25 nonce collision this PR fixes. Distinct per-call keys would let two ops for that same account race on the nonce — reintroducing the bug. Over-serializing two genuinely-different unknown accounts is harmless (a brief wait), so the conservative fallback is the safer trade-off.

brunodmsi:
**Quality-gate note (non-blocking):** the gate flagged `knipUnusedExports 416 → 417 (+1)`. This is knip non-determinism, not a real regression:

- Running `knip` locally on this branch reports **416 unused exports — identical to the baseline**.
- Both new exports (`serializeBySender`, `withUserOpSerialization`) are imported by `alchemy.provider.ts` / `zerodev.provider.ts`, so neither is unused.
- The report also shows `unused files -1` in the same run — a reachability reshuffle in knip's graph, unrelated to this diff.

No code change is warranted; re-running the gate. (`verdict` is the QA/dogfood pipeline — a Privy auth-setup flake — and is also non-blocking.)

brunodmsi:
### Babysit sign-off

**Self-review:** 98/100 (senior pass — no implementation changes needed).

**Bot reviews:** CodeRabbit re-reviewed ✅ — 1 nitpick applied (undefined-sender test), 1 reasoned-rejected (distinct keys would reintroduce the AA25 bug). karma-coding-ai posted a QA plan only (no code requests). No unresolved review threads.

**CI:** all substantive checks green — `build`, `test`, `static-checks`, `react-doctor`, `gate-check`, `baseline-guard`, `checklist`, `claude`, `scope`, `smoke`, `CodeRabbit`. `main` has no required status checks.

**Two non-blocking reds (neither caused by this diff):**
- `verdict` — QA/dogfood pipeline, Privy auth-setup flake.
- `quality-gate` — `knipUnusedExports 416 → 417 (+1)`. Does **not** reproduce locally (branch = 416 = baseline); the two new exports are imported by the providers and the test. The flip only appears in CI's Storybook-plugin knip graph, shaking loose one export unrelated to the gasless code — brittle-baseline noise, not a real regression.

PR #1592, opened 2026-06-08
chore(next): probe upgrade to Next.js 16.2.7 (test concurrent-render isolation)

amaurymagalhaes:
## ⚠️ Draft — hypothesis probe, do NOT merge without preview load-testing

This is a **swing at the cross-request canonical/metadata bleed** seen under concurrency (one request's `generateMetadata` output — title/og/canonical — leaking into another request's response, across routes). That symptom points at **App Router request-store (AsyncLocalStorage) isolation under concurrency**, which is framework-level. The 15.x line is maxed (15.5.18 → 15.5.19 is only a backport patch), so the only Next release that could change render isolation is the **16.x major**.

**There is no changelog evidence that 16.2.7 specifically fixes this bleed.** It's a hypothesis. The only way to confirm is to deploy this branch to a preview and run the included harness under load.

## What's in here
- `next`, `@next/bundle-analyzer`, `@next/third-parties` → **16.2.7**
- Removed the `eslint` key from `next.config.ts` (removed from `NextConfig` in 16)
- Next 16's `tsconfig.json` auto-update (`jsx: react-jsx`)
- `scripts/repro/canonical-leak-repro.mjs` — concurrency leak detector

`pnpm build` (Turbopack) ✓ · `pnpm typecheck` ✓

## How to verify (the whole point)
1. Deploy this branch to a Vercel preview.
2. `node scripts/repro/canonical-leak-repro.mjs https://<preview-url>` — exits non-zero if any response's canonical points at a different entity.
3. Compare against a **15.5.18 preview** run. If 16 shows 0 leaks under the same load and 15 shows leaks, the hypothesis holds.

## Before this could ship to production
- **`@sentry/nextjs` → v10** — v9's peer is `next 13.5.9`/`react 18`; it installs + compiles under pnpm's lenient resolution but is officially unsupported on Next 16.
- **`middleware.ts` → `proxy.ts`** — Next 16 deprecates the middleware convention (currently a warning, still works).
- Full QA pass (major version).


amaurymagalhaes:
Closing per investigation: web research found no evidence Next 16 fixes a cross-request render/metadata bleed, and 15→16 has caused regressions for others (#85470 latency, #90433 memory leak). The documented App Router cross-request leak class is the caching layers (Data/Full-Route Cache), which are version-independent and fixed in app code, not by upgrading. Keeping this branch for reference but not pursuing the upgrade as a fix.

PR #1617, opened 2026-06-11
refactor(applications): migrate AIEvaluationButton to useMutation and unify evaluation error extraction

### LINKED ISSUE #1504: [QA Dogfood] RunKarmaProfileButton uses useState + direct API call instead of useMutation
**Severity**: low
**Page**: unknown

## Description

RunKarmaProfileButton manages loading state with useState and calls fundingApplicationsAPI.runKarmaProfileEvaluation() directly, bypassing useMutation. CLAUDE.md mandates 'Always useMutation with optimistic updates — never useState + direct service calls.' This means errors are not in the React Query error cache, loading state is not shared, and there's no automatic retry behavior.

---
*Filed automatically by the QA dogfood pipeline (PR #1501).*
*Fingerprint: `run-karma-profile-button-no-use-mutation`*

amaurymagalhaes:
## Summary

The funding-application AI evaluation triggers had drifted into two different patterns. Three of the four buttons already used React Query mutation hooks with cache invalidation, while `AIEvaluationButton` still used a local `useState` flag plus a direct service call, swallowed backend error messages, and logged to the console. This PR migrates that last button onto the shared mutation architecture and unifies axios error extraction across all four triggers.

## Root cause

`AIEvaluationButton` managed its in-flight state with `useState(isEvaluating)` and called `fundingApplicationsAPI.run...` directly, so freshly-run verdicts only propagated through the local `onEvaluationComplete` callback rather than via cache invalidation. The same axios error-narrowing logic was duplicated across the four buttons, and two of them used a weaker `error instanceof Error` check that dropped the backend's domain message for axios errors.

## What changed

- `hooks/useRunAIEvaluation.ts` — new `useMutation` hook (the fourth member of the evaluation-mutation family) that dispatches to `runInternalAIEvaluation` or `runAIEvaluation` based on `isInternal`, then invalidates `["funding-application"]` and `["applications"]` before the optional `onSuccess` callback, mirroring the sibling hooks exactly.
- `components/FundingPlatform/ApplicationView/AIEvaluationButton.tsx` — consumes the hook; drives `disabled`/`aria-busy`/spinner/label off `mutation.isPending`; removes the `useState` pair and both `console.error` calls. Public props are unchanged, so the five render sites need no edits.
- `utilities/errors.ts` — adds `extractApiErrorMessage(error, fallback)` with typed `axios.isAxiosError` narrowing (no casts), adopted across all four buttons. `ReEvaluateKarmaProfileButton` and `ReEvaluateInternalButton` now surface backend response messages instead of generic axios messages.
- `RunKarmaProfileButton.tsx`, `ReEvaluateKarmaProfileButton.tsx`, `ReEvaluateInternalButton.tsx` — adopt the shared extractor.

## Testing

- `vitest run --project unit` on the three changed test files: 3 files, 48 tests passing (`errors.test.ts`, `useRunAIEvaluation.test.tsx`, `AIEvaluationButton.test.tsx`).
- `tsc --noEmit`: no errors in changed files.
- `biome check`: clean on all changed source files.

Closes #1504


amaurymagalhaes:
## Senior engineering review

**Score: 90/100**

**Verdict: approve with nits.** The PR does what it claims: `AIEvaluationButton` — the last evaluation trigger still on the useState + direct-service-call anti-pattern flagged by issue #1504's fingerprint — is migrated to a `useMutation` hook (`useRunAIEvaluation`) that mirrors the two existing sibling hooks, including cache invalidation of the verified `["funding-application"]` and `["applications"]` keys, and the duplicated axios error narrowing across all four buttons is replaced by a typed `extractApiErrorMessage`. I ran the PR's three test files in a worktree (48/48 pass), biome (clean) and tsc (no errors in changed files), and verified all five render sites need no changes. Note that #1504 literally names `RunKarmaProfileButton`, which was already migrated on main, so this PR closes the issue by finishing the residual instance of the same pattern.

### What works
- Root anti-pattern eliminated: no more `useState` + direct `fundingApplicationsAPI` calls; pending state, errors and cache propagation flow through `useMutation`, with invalidation keys verified against existing query usage in `hooks/useFundingPlatform.ts` and the sibling hooks.
- Public props of `AIEvaluationButton` unchanged; all render sites (`ApplicationContent.tsx` x4, `AIAnalysisTab/index.tsx`) checked — no broken callers.
- `extractApiErrorMessage` uses typed `axios.isAxiosError` narrowing (no `any`), removes three divergent inline copies, and genuinely fixes `ReEvaluateKarmaProfileButton` / `ReEvaluateInternalButton`, which previously dropped backend domain messages for axios errors.
- Both `console.error` calls removed; `"use client"` present on the new hook.
- Tests are meaningful, not tautological: invalidation-before-callback ordering, dispatch by `isInternal`, error path with no invalidation, plus a component-level invalidation test. Verified locally: 3 files, 48 tests, all pass; biome and tsc clean on changed files.
- Hook mirrors the established `useReEvaluateKarmaProfileAI` / `useReEvaluateInternalAI` pattern exactly.

### Findings
- **[minor]** Callback failure now masks evaluation success (`components/FundingPlatform/ApplicationView/AIEvaluationButton.tsx:55-66`) — `onEvaluationComplete` is awaited inside the mutation's `onSuccess`, so if the callback rejects, `mutateAsync` rejects: no success toast, the callback's raw error is shown as if the evaluation failed, and `mutation.isError` is true despite a successful run. The old code distinguished these ("Evaluation completed but failed to refresh the display..."). Both real callers are safe today — `ApplicationContent.handleAIEvaluationComplete` catches internally, and `ApplicationDetailView` passes `refetchApplication` which does not reject — but the contract is fragile for future callers. Fix: wrap the `await onSuccess()` in `hooks/useRunAIEvaluation.ts:47-49` in its own try/catch surfacing a distinct "completed but failed to refresh" message, or document the rejection semantics on the prop. The new test at `AIEvaluationButton.test.tsx:324` deliberately pins this behavior, so it reads as a conscious choice — open (non-blocking).
- **[minor]** `useRunAIEvaluation` with `isInternal=true` fully duplicates `useReEvaluateInternalAI` (`hooks/useRunAIEvaluation.ts:40-50`) — identical `mutationFn`, identical invalidations, identical `onSuccess` chaining. Two hooks now wrap the same endpoint with the same cache behavior, and the `onSuccess` invalidation block is copy-pasted in three files. Fix: share a common factory (e.g. `makeEvaluationMutation(mutationFn)`) or collapse the internal branch onto `useReEvaluateInternalAI`. Acceptable as-is since it mirrors the established family pattern — open (non-blocking).
- **[nit]** Hardcoded query-key strings instead of the existing key factory (`hooks/useRunAIEvaluation.ts:44-45`) — inline `["funding-application"]` / `["applications"]` literals despite `hooks/fundingPlatformQueryKeys.ts:25` defining a key factory. Consistent with the sibling hooks; a follow-up should add prefix constants and use them in all three hooks to prevent silent key drift — open (non-blocking).
- **[nit]** `extractApiErrorMessage`: dead `??` branch and inconsistent empty-string handling (`utilities/errors.ts:15-17`) — `AxiosError.message` is typed `string`, so the final `?? fallback` is unreachable, and an empty-string `data.message` is returned verbatim instead of falling through (the replaced inline code used `||`). Fix: `return error.response?.data?.message || error.message || fallback;` — open (non-blocking).
- **[nit]** Doc comment says "fourth member" but names only two siblings (`hooks/useRunAIEvaluation.ts:21-30`) — the fourth trigger, `RunKarmaProfileButton`, reuses `useReEvaluateKarmaProfileAI` rather than having its own hook. Reword to name the buttons or drop the count — open (non-blocking).
- **[nit]** PR closes #1504 whose literal subject was already fixed on main (`components/FundingPlatform/ApplicationView/RunKarmaProfileButton.tsx`) — the issue names `RunKarmaProfileButton`, which already uses a mutation on main; this PR fixes the residual instance (`AIEvaluationButton`) plus the duplicated error extraction. Closing here is defensible, but the PR body should state that the issue's named component was already migrated, to avoid confusing future archaeology — open (non-blocking).

No required findings; no follow-up commit was pushed.

Issue #1573, opened 2026-06-04
gap-app-v2: stale external wallet (MetaMask) takes over signing & ownership checks for a Privy email/Google session


amaurymagalhaes:
**Component:** gap-app-v2 (frontend / wallet + auth)
**Environment observed:** staging (gapstag.karmahq.xyz → gapstagapi, OP Sepolia)

## Summary

When a user is authenticated via Privy with **email/Google login** (so they have an embedded wallet that owns their projects), but the same browser profile still has a **previously‑connected external wallet (MetaMask) for a *different* account**, the app treats that stale external wallet as the **active signing wallet**. As a result:

1. Project actions that attest on‑chain (create project, create/complete milestone) are routed to the **foreign MetaMask account** instead of the user's embedded wallet — producing a MetaMask signature prompt for an account the user never chose, and breaking the otherwise‑gasless flow.
2. The project **ownership/permission check resolves against the wrong wallet**, so owner‑only controls (“Post an update”, “Mark Milestone Complete”, the milestone options/undo menu) intermittently disappear even though the logged‑in user *is* the project owner.

The leftover external wallet should never take over control/usage of a project for a Privy‑authenticated user whose embedded wallet is the owner.

## Steps to reproduce

1. In a browser profile, connect MetaMask to GAP at some point with account **A** (e.g. `0x9b75…`).
2. Log out, then log in with **email or Google** as a *different* user **B** whose **embedded** Privy wallet (e.g. `0x641D…`) owns a project.
3. Open one of user B’s projects and try to create a project / create or complete a roadmap milestone.

## Actual behavior

- Privy keeps the active wallet pointed at the stale external connection:
  `privy:<app>:active-wallet-connection = "metamask:0x9b750f08b73D7441d4A0eFF112648764613019A4"`
  even though the authenticated user logged in via Google and owns the project with embedded wallet `0x641D6Fd549F4F1bD336aC9E5c14be8d1f3688Fdf`.
- Creating a project triggers a MetaMask signature request to account `0x9b75…` and fails:
  `MetaMask - RPC Error: MetaMask Tx Signature: User denied transaction signature.`
- Owner controls flicker / are missing because the on‑chain owner/admin check resolves against the wrong/active wallet (see `useProjectPermissions` notes below).

## Expected behavior

For a Privy session authenticated via email/Google with an embedded wallet:
[...]

PR #1618, opened 2026-06-11
fix(auth): restrict signing and wallet identity to wallets linked to the authenticated Privy user

### LINKED ISSUE #1574: Stale external wallet (MetaMask) takes over signing & ownership checks for a Privy email/Google session
**Component:** gap-app-v2 (frontend / wallet + auth)
**Environment observed:** staging (gapstag.karmahq.xyz → gapstagapi, OP Sepolia)

## Summary

When a user is authenticated via Privy with **email/Google login** (so they have an embedded wallet that owns their projects), but the same browser profile still has a **previously‑connected external wallet (MetaMask) for a *different* account**, the app treats that stale external wallet as the **active signing wallet**. As a result:

1. Project actions that attest on‑chain (create project, create/complete milestone) are routed to the **foreign MetaMask account** instead of the user's embedded wallet — producing a MetaMask signature prompt for an account the user never chose, and breaking the otherwise‑gasless flow.
2. The project **ownership/permission check resolves against the wrong wallet**, so owner‑only controls (“Post an update”, “Mark Milestone Complete”, the milestone options/undo menu) intermittently disappear even though the logged‑in user *is* the project owner.

The leftover external wallet should never take over control/usage of a project for a Privy‑authenticated user whose embedded wallet is the owner.

## Steps to reproduce

1. In a browser profile, connect MetaMask to GAP at some point with account **A** (e.g. `0x9b75…`).
2. Log out, then log in with **email or Google** as a *different* user **B** whose **embedded** Privy wallet (e.g. `0x641D…`) owns a project.
3. Open one of user B’s projects and try to create a project / create or complete a roadmap milestone.

## Actual behavior

- Privy keeps the active wallet pointed at the stale external connection:
  `privy:<app>:active-wallet-connection = "metamask:0x9b750f08b73D7441d4A0eFF112648764613019A4"`
  even though the authenticated user logged in via Google and owns the project with embedded wallet `0x641D6Fd549F4F1bD336aC9E5c14be8d1f3688Fdf`.
- Creating a project triggers a MetaMask signature request to account `0x9b75…` and fails:
  `MetaMask - RPC Error: MetaMask Tx Signature: User denied transaction signature.`
- Owner controls flicker / are missing because the on‑chain owner/admin check resolves against the wrong/active wallet (see `useProjectPermissions` notes below).

## Expected behavior

For a Privy session authenticated via email/Google with an embedded wallet:
- The **embedded wallet linked to the authenticated user** (the one that owns the project) should be the active wallet used for signing/attestations — i.e. the gasless embedded path — regardless of any stale external connection left in the browser.
- A leftover external‑wallet connection from a prior session must not become the active wallet, must not receive signature prompts, and must not be used as the basis for ownership/permission resolution.

## Evidence it’s the wallet selection (not a real auth problem)

Clearing/overriding the stale `active-wallet-connection` to the embedded wallet and reloading made every action use the embedded wallet **gaslessly**:
- console: `Embedded1193Provider.request() called with args …`
- Privy dialog: “Sign message — Signing this message will not cost you any fees.”
- create/complete milestone then succeeded with no MetaMask prompt and no gas.

The authenticated user genuinely is the owner: the indexer returns `project.owner = 0x641d6fd…` = the embedded wallet.

## Likely areas to investigate (gap-app-v2)

- `hooks/useZeroDevSigner.ts` — `getAttestationSigner` / `attestationAddress` choose embedded vs external. They key off `isEmailOrSocialLogin` + the resolved `embeddedWallet`/`externalWallet` from `useWallets()`; when the stale external connection is active, the external wallet appears to win.
- `hooks/useSetupChainAndWallet.ts` — consumes the above for all attest flows.
- `hooks/useProjectPermissions.ts` — `candidateAddresses` + the `isConnected` early‑return; ownership flickers when the active/connected wallet is the foreign account and the on‑chain `isOwner`/`isAdmin` check (over staging’s rate‑limited RPC) is the only path that runs.
- Privy `active-wallet-connection` persistence — a stale `"metamask:…"` value survives a new email/Google login and is never reconciled to the authenticated user’s embedded wallet.
- See `gap-app-v2/CLAUDE.md` → **Auth Gotchas** (Privy and Wagmi initialize independently).

## Impact

- Users who ever connected MetaMask then switched to email/Google login can be blocked from acting on their own projects (wrong‑account signature prompts, missing owner controls), and may be silently asked to sign with an account that isn’t theirs.

## Notes

Found while validating gap-indexer #1980 (revocation handler). The revocation fix itself is verified and unrelated to this issue.

amaurymagalhaes:
## Summary

Makes "a wallet may only sign or confer authority if it is linked to the authenticated Privy user" an enforced invariant. This eliminates two related failure paths: a wallet that is *not* linked to the signed-in user being prompted to sign (e.g. a stale MetaMask popping up for a foreign address and paying gas), and project ownership/admin flags flickering to `false` during query-key transitions.

## Root cause

Signing identity was derived from whichever wallets happened to be connected rather than from the login method and the set of wallets actually linked to the authenticated Privy user. As a result:

- The ZeroDev signer path could silently fall through to an external (foreign) wallet when the embedded wallet had not yet hydrated, prompting the wrong account to sign.
- Privy's persisted `active-wallet-connection` could reassert a stale connection after reload, so wagmi's active wallet diverged from the user's linked primary wallet.
- `selectPrimaryWallet` fell back to `wallets[0]` even for authenticated users with linked wallets, surfacing a foreign address as the identity.
- `useProjectPermissions` collapsed owner/admin flags to `false` on query-key transitions, causing visible flicker.

## What changed

- **`utilities/auth/resolve-signing-wallets.ts`** (new): single source of truth. `resolveSigningWallets(user, wallets)` returns `{ embeddedWallet, externalWallet, signingMode }`. External wallets must pass `compareAllWallets` to be eligible, so a stale/foreign wallet is excluded outright; `signingMode` derives from the login method, not from connected wallets. Also hosts the relocated `didUserLoginWithEmailOrSocial`.
- **`hooks/useZeroDevSigner.ts`**: derives `embeddedWallet`/`externalWallet`/`signingMode` from the resolver and adds a hard gate — when `signingMode === "embedded"` the external path is unreachable; it throws the new domain-specific `EmbeddedWalletNotReadyError` (or the wrapped last error) instead of silently rerouting to a foreign wallet.
- **`components/Utilities/PrivyWagmiProviders.tsx`**: a ref-guarded effect uses `useSetActiveWallet` to push the linked primary wallet back as Privy's active wallet when it diverges from `wallets[0]`, rewriting the persisted active-wallet-connection so a stale connection cannot reassert itself after reload. The call has a `.catch()`.
- **`utilities/auth/select-primary-wallet.ts`**: returns `undefined` (withholds identity) when an authenticated user has linked wallet addresses but none of the connected wallets match, keeping the `wallets[0]` fallback only for the pre-auth / no-linked-wallets case.
- **`hooks/useProjectPermissions.ts`**: adds `placeholderData: keepPreviousData` so owner/admin flags persist across query-key transitions instead of flickering to `false`.

## Testing

- New `__tests__/utilities/auth/resolve-signing-wallets.test.ts` covers the resolver's trust invariant.
- `__tests__/integration/wallet/signer-trust.test.ts` and `__tests__/trust/wallet/signer-trust.test.ts` rewritten as regression guards asserting a foreign wallet's provider is never invoked for embedded-mode users.
- `__tests__/utilities/auth/select-primary-wallet.test.ts` adds the `undefined`-return cases.
- Targeted vitest across the resolver, select-primary-wallet, both signer-trust suites, `useZeroDevSigner-real`, `useAuth.primary-wallet`, and the `useAuth` suites — all passing.
- `tsc --noEmit` clean; biome clean on all changed source files.

Closes #1574, #1573

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted signing and wallet identity to wallets linked to the authenticated account.
  * Prevented stale or unlinked wallets from being used or exposed.
  * Improved embedded-wallet hydration and fallback behavior.
  * Prevented authentication and wallet reconciliation loops during reloads.

* **Tests**
  * Added coverage for wallet linkage, signer selection, hydration, fallback scenarios, and public rendering stability.


amaurymagalhaes:
## Senior engineering review

**Score: 78/100** → **91/100** after follow-up commit 9bd5276

**Verdict: needs changes → resolved by follow-up.** The PR correctly identifies and fixes the root cause of #1574/#1573: a pure `resolveSigningWallets` resolver enforces wallet-to-user linkage, `selectPrimaryWallet` stops leaking foreign addresses, and the embedded-mode hard gate makes the exact reported failure (a foreign MetaMask prompted to sign for an email/Google user) impossible — verified by regression tests against the real hook, with tsc and biome clean. However, the original gate over-restricted beyond the stated invariant, permanently locking out embedded-mode users without an embedded wallet (Farcaster, hybrid email+wallet accounts), and unscoped `keepPreviousData` in `useProjectPermissions` could flash owner-only controls on projects the user does not own. Both majors were fixed in the follow-up commit with regression tests.

### What works
- Root cause addressed at the right layer: a pure, framework-free resolver (`utilities/auth/resolve-signing-wallets.ts`) with linkage enforced via the existing `compareAllWallets`, shared by the signer and identity paths so they cannot diverge.
- Strong, non-tautological tests: `__tests__/integration/wallet/signer-trust.test.ts` renders the real `useZeroDevSigner` and asserts the foreign wallet's `getEthereumProvider` is never invoked; resolver and select-primary-wallet unit tests cover hydration, hybrid, and foreign-wallet cases.
- `selectPrimaryWallet`'s undefined-withholding is carefully scoped (only when the user provably has linked wallets), preserving the pre-auth/hydration `wallets[0]` fallback; both callers already handle `undefined`.
- Repo standards respected: `.catch()` on `setActiveWallet`, no new console statements, no `any` in source, a domain-specific `EmbeddedWalletNotReadyError` instead of a generic throw, `tsc --noEmit` and biome clean.
- Deriving `signingMode` from login method rather than connected-wallet presence eliminates the silent embedded→external fallthrough race, which was the actual signing-path vulnerability.

### Findings
- **[major]** Hard gate locks out embedded-mode users who have no embedded wallet but DO have a linked external wallet (`hooks/useZeroDevSigner.ts:276`) — the gate threw for ANY `signingMode === "embedded"` user before Case 3, even when `externalWallet` is non-null (which by construction of `resolveSigningWallets` is linked to the authenticated user). Farcaster users never get an embedded wallet (`useEnsureEmbeddedWallet` skips creation because the Farcaster `ownerAddress` counts as a linked wallet, and `createOnLogin` is `"off"`), so they hit a permanent `EmbeddedWalletNotReadyError` with a misleading "try again in a moment" message; hybrid email+wallet accounts hit the same lockout. Pre-PR these users could sign with their connected wallet. ✅ fixed in 9bd5276 — gate is now `signingMode === "embedded" && !externalWallet`, `attestationAddress` mirrors the fallback, 5 regression tests added.
- **[major]** `keepPreviousData` leaks owner/admin flags across projectId changes (`hooks/useProjectPermissions.ts:155`) — `placeholderData: keepPreviousData` applies to ANY query-key change, so navigating from a project you own to one you don't serves stale `{isProjectOwner: true}` as placeholder, and the effect at lines 159–169 writes it into the global project store, briefly rendering owner-only controls on someone else's project (server-side enforcement still blocks actions). ✅ fixed in 9bd5276 — function form of `placeholderData` keeps previous data only when the previous key's projectId/chainID match the current project; 2 regression tests added.
- **[minor]** Outer wagmi config can stay connected to the foreign wallet after `selectPrimaryWallet` starts returning `undefined` (`components/Utilities/PrivyWagmiProviders.tsx:106`) — once auth completes with no connected linked wallet, the sync effect early-returns and `useAccount()` stays pinned to the foreign address until the embedded wallet hydrates (indefinitely, if hydration fails). Suggested: disconnect the outer config when `authenticated && !primaryWallet && syncedAddressRef.current`. Open (non-blocking).
- **[minor]** Active-wallet reconciliation relies on unverified Privy SDK assumptions (`components/Utilities/PrivyWagmiProviders.tsx:144`) — the effect assumes `wallets[0]` is Privy's active wallet and that `setActiveWallet` rewrites the persisted `active-wallet-connection` localStorage entry; neither is verifiable from this codebase and there is no test coverage. The ref guard prevents loops, so this is safe but possibly ineffective. Open (non-blocking) — left out of scope of the follow-up.
- **[minor]** Wallet-login user mid-hydration now gets a generic dead-end error instead of signing (`hooks/useZeroDevSigner.ts:349`) — when `signingMode === "external"` and linkedAccounts haven't populated yet, the hook throws the generic "No wallet available for signing" instead of a specific, retryable message. Failing closed is the right trade-off, but the error is non-actionable. Open (non-blocking).
- **[nit]** Stale documentation: Farcaster embedded-wallet claim contradicted `createOnLogin: "off"` (`utilities/auth/resolve-signing-wallets.ts:42`). ✅ fixed in 9bd5276 (comment corrected as part of the gate fix).
- **[nit]** Misleading wrapped-error message in the non-embedded fallthrough (`hooks/useZeroDevSigner.ts:342`) — "Failed to obtain signer from embedded wallet" now only fires in external/none modes and mislabels external-wallet failures; make the message mode-aware. Open (non-blocking).

### Follow-up commit
Commit 9bd5276 changed: `hooks/useZeroDevSigner.ts` (gate relaxed to allow LINKED external wallets in embedded mode; `attestationAddress` falls back `embeddedWallet?.address ?? externalWallet?.address ?? null`), `hooks/useProjectPermissions.ts` (scoped `placeholderData` comparing projectId/chainID at key indices 2/3), and the stale comment in `utilities/auth/resolve-signing-wallets.ts`. Both findings were re-verified against the code before fixing (confirmed `useEnsureEmbeddedWallet` skips creation when linked addresses exist and `createOnLogin` is off; confirmed the PERMISSIONS key layout and the store write). Verification: vitest suites for `useZeroDevSigner`, `useProjectPermissions`, and `utilities/auth` — 52 tests passing including 7 new regression tests; full `tsc --noEmit` clean; biome clean on all 5 changed files. Not verified: the Privy SDK reconciliation assumptions (no automated coverage); also note the embedded-signer-construction-failure path now falls through to a linked external wallet instead of throwing — intentional, and still within the linkage invariant since `resolveSigningWallets` never returns unlinked wallets. Remaining deductions in the 91: untested reconciliation effect and the magic-index key comparison in the placeholder scoping (documented inline).

amaurymagalhaes:
## Follow-up: merge reconciliation, evaluation & QA

Picked this up to drive it to merge-ready. Summary of what changed and how it was verified.

### 1. Merge conflict resolved (was `CONFLICTING`/`DIRTY`)
`main` had advanced past the branch's last merge with a **substantial architectural change**: `useProjectPermissions` was rewritten to resolve owner/admin on the **backend across all linked wallets** (`usePermissionsQuery` / `src/core/rbac`), replacing the client-side on-chain `ProjectResolver` check this PR had patched. Resolution:

- **Adopted `main`'s backend RBAC permission resolution** for all 5 conflicted files (`useProjectPermissions.ts` + 4 test files). This *supersedes* the PR's client-side `placeholderData` flicker fix — `main` already solves the same flicker/leak (server-side resolution + `keepPreviousData` in the rbac hook + an explicit cross-project reset effect + fail-closed-on-error), and does it at a better altitude. The PR's permission-flicker contribution is no longer needed.
- The PR's **core contribution — the signing-identity layer** (`resolveSigningWallets`, `useZeroDevSigner`, `select-primary-wallet`, `PrivyWagmiProviders`) — merged cleanly and is fully preserved.

### 2. Two semantic fixes on top of the textual merge
`main` had *independently* rewritten `useZeroDevSigner` (gasless retry, `switchEmbeddedWalletChain`, **and its own foreign-wallet guard** `!(isEmailOrSocialLogin && embeddedWallet)`). The auto-merge combined both rewrites cleanly but left a subtle hazard:

- **`useZeroDevSigner` Case 3:** dropped `main`'s now-redundant `!(isEmailOrSocialLogin && embeddedWallet)` clause. Since `externalWallet` now comes from `resolveSigningWallets` (which only ever returns a wallet **linked** to the authenticated user), foreign-wallet safety is enforced *upstream*; the leftover clause only suppressed the PR's intended **linked-wallet fallback** for embedded-mode users whose embedded path fails or who have no embedded wallet (Farcaster / hybrid email+wallet), turning it into a generic "No wallet available". Added an integration regression test for this.
- **`PrivyWagmiProviders`:** added a disconnect of the outer wagmi config when an authenticated user has linked wallets but none connected (`selectPrimaryWallet` → `undefined`), so `useAccount()` can't stay pinned to a foreign address synced pre-auth — directly closing the last open `[minor]` from the prior review and the issue's "leftover wallet must not drive identity" invariant.

### 3. Evaluation vs. the ideal long-term solution — **96/100**
The root cause is that signing identity + ownership were derived from *connected* wallets rather than from the *authenticated user* and their *linked* wallets. The ideal long-term fix has three parts, and the merged result delivers all three at the right layer (not a workaround):

1. **Single source of truth** mapping `(user, wallets) → (who may sign, how)`, keyed on login method + linkage, shared by signer & identity → `resolveSigningWallets` ✅
2. **Reconcile Privy's persisted active wallet** so a stale connection can't reassert → reconciliation effect + outer-config disconnect ✅
3. **Authoritative ownership across the full linked-wallet set** → `main`'s server-side RBAC ✅

Remaining −4: the active-wallet reconciliation + outer-config-disconnect effects rely on Privy SDK behavior that isn't unit-asserted (covered indirectly by the `privy-bridge-trust` integration suite), and depend on `setActiveWallet` rewriting persisted state (not verifiable from code).

### 4. QA
- **`tsc --noEmit`**: clean. **biome**: clean on changed files.
- **Tests**: targeted + related suites green locally — resolver, select-primary-wallet, both signer-trust suites, `useZeroDevSigner(-real)`, `useAuth(-trust)`, the new backend-RBAC `useProjectPermissions`, and the `Activity`/`UpdatesContent`/`ImpactContent` consumers (**156 tests**), including a **new regression** locking in the linked-wallet fallback. Full CI suite (sharded `test (1..4)`) green.
- **Browser (agent-browser):** the fixed-build Vercel preview deploys and renders cleanly; the Privy login modal renders; **email + OTP login verified end-to-end** on the shared Privy app (`cmdoy9mgz…`). 
- **Documented constraint:** the bug's *distinguishing* path (a **stale, unlinked MetaMask** signing for an email/Google user) can't be reproduced in headless agent-browser — it requires a real wallet extension + a prior cross-account connection, and the Vercel preview origin isn't in Privy's allowlist (auth only completes on allowlisted domains like `staging.karmahq.xyz`). That invariant is therefore verified by the **deterministic regression suite** (the real-hook signer-trust tests assert a foreign wallet's provider is *never* invoked for embedded-mode users), which is the correct verification layer for wallet-state-dependent logic.

CI is green; marking ready for review.

PR #1619, opened 2026-06-11
fix(auth): eliminate Privy frame-ancestors blocks on Vercel previews via scoped preview origins + preview-env diagnostic

amaurymagalhaes:
## Summary

Login is impossible on Vercel preview deployments (#1193, prior recurrence #1176). This PR ships the durable fix as an operational runbook plus an in-repo, preview-only diagnostic that makes the previously silent failure visible.

## Root cause

The app passes a single, environment-unscoped Privy app ID to `PrivyProvider` (`utilities/enviromentVars.ts` → `PRIVY_APP_ID`). Every unbounded Vercel preview hostname therefore authenticates against the same finite, manually-curated Privy **Allowed Origins** list. `auth.privy.io` serves a `frame-ancestors` CSP derived from that list, so any preview hostname not on it has its auth iframe blocked by construction and login fails silently. `next.config.ts` already permits `frame-src` to `auth.privy.io`, confirming the block is Privy's origin policy, not ours.

The fix has two halves: an **operational** one (it cannot live in app code — it needs Privy/Vercel admin access) and an **in-repo guardrail** that surfaces the failure until a deployment is allowlisted.

## What changed

- `docs/auth/privy-preview-deployments.md` — full runbook for the durable fix: a dedicated **dev** Privy app, `NEXT_PUBLIC_PRIVY_APP_ID` scoped per Vercel environment, a Vercel Preview Deployment Suffix on `preview.karmahq.xyz`, and a single `https://*.preview.karmahq.xyz` wildcard registered once. Includes the staging gap-indexer JWT-verification step, a security guardrail (never add `*.vercel.app` to the prod app), a spike procedure to re-validate the failure signature, and a verification checklist.
- `components/Utilities/PrivyOriginDiagnostic.tsx` — a **preview-only** (`envVars.VERCEL_ENV === "preview"`), dismissible `role="alert"` advisory. It arms only after a real login attempt while still unauthenticated — `PrivyBridgeProvider` wraps the bridge `login()` pushed by `PrivyBridgeUpdater`, so every sign-in path in the app (useAuth `login`/`authenticate`, auto-login after logout) sets the new `usePrivyLoginAttempted()` flag — then trips **only** on an explicit, parent-observable `auth.privy.io` HTTP 403 (`PerformanceResourceTiming.responseStatus`, Chromium 109+) — the origin rejection — and links to the runbook. The 403 signal was verified by spike (recorded in the runbook): Privy's CORS layer echoes any origin, so the `invalid_origin` 403 from the auth API passes the CORS check and Chromium exposes `responseStatus === 403` on the resource entry; the iframe document itself reads `0`, which is why the trigger keys on the API call. It deliberately does **not** trip on a bare timeout or on `responseStatus === 0` (a slow-but-working emailed-OTP flow outlasts any reasonable timeout, and `0` is the normal cross-origin value for a successful subresource without `Timing-Allow-Origin`), so it cannot false-positive during ordinary preview QA. Visibility is derived, so a later successful login clears a shown banner.
- `components/Utilities/PrivyProviderWrapper.tsx` — renders the diagnostic as a sibling of the lazy Privy SDK, preserving the no-remount invariant.
- `contexts/privy-bridge-context.tsx` — wraps the bridge `login()` in the provider so any sign-in attempt flips `usePrivyLoginAttempted()`; this is the arming signal for the diagnostic (nothing in app code calls `useLoadPrivy()`, so the previous `usePrivyLoadRequested()` gate could never arm).
- `.env.example` — annotates the per-environment Privy app ID scoping.

## Testing

- Spike against the live Privy app from a non-allowlisted origin (`curl` + headless Chromium): preflight 204 with origin echoed; `POST /api/v1/siwe/init` → 403 `invalid_origin` with origin echoed in `Access-Control-Allow-Origin`; `performance.getEntriesByType("resource")` reports `responseStatus: 403` for the entry. Full transcript in the runbook's "Spike result" section.
- `vitest run --project unit`: 13/13 unit tests on `PrivyOriginDiagnostic.test.tsx` plus 3 integration tests (`__tests__/integration/auth/privy-origin-diagnostic.test.tsx`) that render the **real** `PrivyBridgeProvider` and simulate the actual sign-in click — guarding the arming path against regressing into dead code. Existing bridge trust/context/auth-integration suites still pass (48 total). Covers: no render outside preview; does not arm or register an observer before a login attempt; no false positive on a slow-but-working flow (no timeout trip) or on cross-origin status 0 or 200; fires on an explicit `auth.privy.io` 403; ignores 403s from other origins and entries without `responseStatus`; stays hidden when the engine lacks `responseStatus`; clears a shown banner once authentication later succeeds; dismiss hides the banner; observer disconnects on unmount.
- `biome check` clean on the changed files; `tsc --noEmit` reports no type errors in the new component.

The operational steps (dev Privy app, per-env env scoping, Vercel suffix, dev-app allowed origins, staging indexer JWT scope) require dashboard access and are tracked in the runbook for the human owner.

Refs #1193 (deliberately NOT auto-closed: the durable fix is operational — dev Privy app, per-env `NEXT_PUBLIC_PRIVY_APP_ID`, Vercel Preview Deployment Suffix, wildcard origin — and #1193 should stay open until the runbook's verification checklist passes on a real preview). Refs #1176 (prior recurrence, already closed).


amaurymagalhaes:
## Senior engineering review

**Score: 48/100** → **88/100** after follow-up commit c9a9c2c

**Verdict: needs changes** (pre-fix; the follow-up commit addresses all required findings)

The diagnosis of the root cause is excellent and the operational runbook is genuinely valuable, but the only functional code the PR originally shipped — the preview diagnostic — was provably dead: it armed on `usePrivyLoadRequested()`, whose setter `useLoadPrivy()` has zero call sites in app code, so the banner could never appear on a real preview deployment. Even if armed, the 403 trigger relied on cross-origin `PerformanceResourceTiming.responseStatus` without any spike to verify the signal, and "Closes #1193" would have auto-closed the open QA-blocking issue although the actual fix is deferred dashboard work. The 13 passing tests mocked away both broken assumptions, validating internal logic rather than the fix.

### What works

- Root-cause analysis is accurate and well-evidenced: a single environment-unscoped `NEXT_PUBLIC_PRIVY_APP_ID` maps unbounded Vercel preview hostnames onto one finite Privy Allowed Origins list, and the `frame-ancestors` CSP is served by Privy, not the app.
- The operational fix in `docs/auth/privy-preview-deployments.md` is the correct durable approach (dedicated dev Privy app, per-environment app ID, Preview Deployment Suffix, single wildcard origin), includes a real security guardrail (never `*.vercel.app` on the prod app), covers the staging gap-indexer JWT-verification gap, and documents a rejected alternative with sound reasoning.
- Component code quality is high in isolation: `"use client"` present, module-level constants, proper types, derived visibility, observer cleanup on unmount, fail-safe-toward-silence design, accessible `role="alert"` banner, dark-mode theme classes, no console noise.
- Tests run and pass under vitest and biome is clean on all changed files.
- PrivyProviderWrapper integration preserves the lazy-load/no-remount invariant — the diagnostic is a sibling, children stay at a stable tree position.

### Findings

- **[critical]** Diagnostic never arms: `usePrivyLoadRequested()` is always false at runtime (`components/Utilities/PrivyOriginDiagnostic.tsx:76-84`) — the component gates everything on `armed = isPreview && loadRequested && !authenticated`, but `loadRequested` is only set by `useLoadPrivy()`, which has zero call sites in app code; real logins go through `bridge.login` and never set it. The banner could never appear on a real preview deployment, making the component dead code. ✅ fixed in c9a9c2c — `PrivyBridgeProvider` now wraps the bridge `login()` so every sign-in path flips a new `usePrivyLoginAttempted()` flag, and the diagnostic arms on that instead.
- **[major]** Detection signal unverified: cross-origin `responseStatus` was predicted to read 0, not 403 (`components/Utilities/PrivyOriginDiagnostic.tsx:55-70`) — the trigger requires `responseStatus === 403` for an auth.privy.io entry, but cross-origin `responseStatus` is only exposed when the response passes the CORS check, and no spike had been run to confirm the signal. ✅ resolved in c9a9c2c — the spike was actually run against the live Privy app from a non-allowlisted origin: Privy's CORS layer echoes any origin in `Access-Control-Allow-Origin`, so the 403 from `siwe/init` / `passwordless/init` passes the CORS check and headless Chromium empirically reports `responseStatus === 403`. The predicted "will read 0" turned out wrong for the API-call path (the iframe document does read 0, confirming the trigger must key on the API 403, which it does). The prescribed trigger change (SDK error callbacks / duration heuristics) was deliberately NOT applied — the empirical evidence shows the existing trigger is correct; spike transcript is recorded in the runbook and component docblock.
- **[major]** "Closes #1193" would auto-close an open QA-blocking issue without the fix being applied (PR body / `docs/auth/privy-preview-deployments.md`) — the actual fix is operational dashboard work explicitly deferred to a checklist, so merging would recreate the recurrence pattern (#1176 → #1193) the runbook itself documents. ✅ fixed in c9a9c2c — PR body changed to "Refs #1193 / Refs #1176" with an explicit note that #1193 stays open until the verification checklist passes on a real preview. A separate ops-task issue was not opened: #1193 remains the tracking issue, which the finding explicitly allows and avoids duplication.
- **[major]** Tests were tautological at exactly the two unverified boundaries (`__tests__/unit/components/PrivyOriginDiagnostic.test.tsx:10-17`) — they mocked `usePrivyLoadRequested` to true and synthesized entries with `responseStatus: 403`, so they could not catch that the component never fires in the real app. ✅ fixed in c9a9c2c — added `__tests__/integration/auth/privy-origin-diagnostic.test.tsx` (3 tests) rendering the real `PrivyBridgeProvider` and simulating the actual sign-in click; it fails on the old code. Unit suite updated to mock the new hook (13/13 retained).
- **[nit]** Hardcoded GitHub URL pinned to `main` inside the component (`components/Utilities/PrivyOriginDiagnostic.tsx:46-47`) — properly extracted to module level and a docs deep-link is a reasonable exception, but consider a shared constants module if other diagnostics link to runbooks. Open (non-blocking).
- **[nit]** Diagnostic mounts unconditionally in production trees (`components/Utilities/PrivyProviderWrapper.tsx:77`) — it early-returns null outside preview but still subscribes to two contexts; gating the render at the wrapper would make production cost exactly zero and the preview-only intent visible at the call site. Open (non-blocking).

### Follow-up commit

Commit c9a9c2c:
- Rewired arming: `PrivyBridgeProvider` wraps the bridge `login()` to set a `usePrivyLoginAttempted()` flag covering all login paths (useAuth login/authenticate, auto-login after logout); diagnostic arms on it.
- Ran the previously missing spike (curl with spoofed Origin headers against auth.privy.io plus a real headless-Chromium page on a non-allowlisted origin reading `performance.getEntriesByType('resource')`): `responseStatus === 403` confirmed. Transcript added to the runbook ("Spike result (2026-06-11)"); component docblock now cites verified evidence.
- PR body switched from Closes to Refs for #1193/#1176 (verified #1193 open, #1176 already closed).
- Added 3 integration tests exercising the real provider + sign-in path; unit suite updated.

Verification run: 48/48 tests passing across the 5 touched/adjacent suites (`vitest --project unit`), biome clean on the 4 changed code files, `tsc --noEmit` shows no errors referencing the touched files. Caveat, stated honestly: the spike used the dev Privy app ID and a localhost origin rather than an actual Vercel preview running the full SDK — strong but not literal end-to-end evidence. End-to-end banner appearance on a real preview is still unproven until the next PR preview exercises it, and the operational fix itself remains deferred dashboard work; the runbook's on-preview verification checklist still gates closing #1193.

amaurymagalhaes:
Not an issue for now.

PR #1622, opened 2026-06-11
fix(milestones): fail fast on revoke errors instead of polling for 5 minutes

### LINKED ISSUE #1567: Undo milestone completion spins indefinitely when the revoke fails (poll-forever + swallowed error)
## Summary

When a user revokes a milestone completion, a failure in the underlying revoke does **not** surface as an error. Instead the UI spins for ~5 minutes and then shows a generic error. Any revoke failure — a 403/401, a network error, or the indexer simply not reflecting the change — produces a multi-minute stuck spinner on the confirm dialog.

This was surfaced by a real report on `project/tower-exchange`: revoking a completed **project (roadmap) milestone** hung indefinitely. The *immediate* root cause was a missing indexer revocation handler (fixed in gap-indexer PR #1980 / branch `claude/exciting-knuth-kauOm`), which left the milestone's `completed` field populated so the poll condition was never met. This issue tracks the **frontend resilience gap** that turned that backend miss into a 5‑minute spinner instead of a prompt, actionable error — which would protect against any future revoke failure.

## Where

- `hooks/useOffChainRevoke.ts` — `performOffChainRevoke`
- `hooks/useMilestone.ts` — `multiGrantUndoCompletion` (single + multi-grant branches)
- `utilities/retries.ts` — `retryUntilConditionMet` (default 200 × 1500ms ≈ 5 min)
- `components/Shared/ActivityCard/MilestoneCard.tsx` — `handleUndoCompletion` / `DeleteDialog isLoading={isUndoing}`

## Root cause (frontend)

1. **Swallowed failure.** `performOffChainRevoke` reports failure via a `false` return value (and a `catch` that returns `false`) rather than throwing:
   ```ts
   if (res[1]) { showError(res[1]); onError?.(res[1]); return false; }
   // ...
   } catch (error) { dismiss(); onError?.(error); return false; }
   ```
2. **Caller ignores the result.** `multiGrantUndoCompletion` does `await performOffChainRevoke(...)` without checking the boolean, then unconditionally proceeds to poll for the completion to disappear:
   ```ts
   await performOffChainRevoke({ uid: objective.completed.uid, chainID });
   await checkIfProjectCompletionExists(...); // waits for !completed
   ```
3. **Poll runs to full budget.** When the revoke didn't actually clear the completion, `retryUntilConditionMet` polls for its entire 200 × 1.5s ≈ 5 min budget before throwing a generic *"Condition was not met"*. The `DeleteDialog` confirm button (`isUndoing`) spins the whole time.

Two contributing smells:
- **Two independent `useAttestationToast()` instances** (one in `useMilestone`, one in `useOffChainRevoke`), each owning a separate `toastIdRef`, which is why two stacked loading toasts appear ("Revoking attestation…" + "Undoing milestone completion…") instead of one updating toast.
- The indexer `fetchData` timeout is 6 min, so a hung request also can't surface quickly.

## Proposed fix

- Make `performOffChainRevoke` **throw** on failure (API error in `res[1]` and caught exceptions), or have every caller check its `false` return and abort the poll. The revoke result must gate whether we poll at all.
- If the revoke succeeds but the poll exhausts its budget, surface a clear, prompt error (and consider a shorter budget for this interactive flow) instead of a silent multi-minute spin.
- Consider sharing a single `useAttestationToast()` instance across the undo flow so the user sees one coherent progress toast.

## Acceptance criteria

- Revoking a completion that fails (mock a 4xx/5xx from `REVOKE_ATTESTATION`) shows an error toast within seconds and the confirm spinner stops — no 5‑minute hang.
- A successful revoke still closes the dialog and refreshes the milestone.
- Unit tests cover: revoke API error, thrown/caught error, and poll-timeout paths (per the repo's testing patterns).

## Notes

- The backend fix (gap-indexer PR #1980) resolves the originally reported tower-exchange hang. This issue is the defensive follow-up so a future backend miss / auth failure / network blip can't reproduce the same indefinite spinner.

amaurymagalhaes:
## Summary

Undoing a milestone/grant completion could hang for roughly 5 to 11 minutes when the off-chain revoke was rejected or the network stalled. The button appeared stuck because the revoke flow encoded failure as an ignorable boolean and then ran the indexing poll unconditionally on the default (~5 minute) budget, with no bound on the revoke request itself. This change makes the revoke contract fail fast: a rejected or failed revoke now surfaces an actionable toast within seconds and keeps the dialog open, and an accepted-but-unindexed revoke times out within ~60s instead of appearing to hang.

## Root cause

- `performOffChainRevoke` returned a boolean to signal failure while owning its own UI side effects; callers discarded the boolean and proceeded to poll regardless.
- The interactive indexing poll inherited the indexer-wide ~5 minute default budget, inappropriate for a button the user is actively watching.
- The revoke POST had no request-level timeout, so it inherited the 360s indexer timeout.

## What changed

- `utilities/errors.ts`: new domain errors `OffChainRevokeError` (codes `API_ERROR | REQUEST_FAILED | INDEXING_TIMEOUT`), `IndexingTimeoutError`, and an `isSurfacedError` guard, plus a shared `INDEXING_TIMEOUT_MESSAGE` constant so both revoke paths show identical copy.
- `utilities/retries.ts`: poll exhaustion now throws a typed `RetryConditionNotMetError` (with `isRetryConditionNotMetError` guard); exported `INTERACTIVE_INDEXING_POLL` (~60s) budget; `isAbortError` broadened to recognise the `CanceledError`/`TimeoutError`/`ERR_CANCELED` shapes that `AbortSignal.timeout()` actually produces through axios, so the internal request timeout surfaces the friendly retry message.
- `hooks/useOffChainRevoke.ts`: rewritten to return `Promise<void>` and throw typed, already-surfaced errors; bounds the POST with a 30s `AbortSignal.timeout`; maps poll exhaustion to `IndexingTimeoutError`; passes caller cancellations through untouched.
- `hooks/useMilestone.ts`: applies the bounded interactive poll budget at the three undo sites, maps exhaustion to `IndexingTimeoutError`, and rethrows surfaced errors so the dialog stays open without duplicate toasts.
- `components/Shared/ActivityCard/MilestoneCard.tsx`: undo converted to React Query `useMutation` so rejection is preserved and drives the dialog state.
- `components/DeleteDialog.tsx`: skips its generic toast for already-surfaced errors while still reporting telemetry.
- Migrated the remaining revoke call sites to the throw contract: on-chain-fallback paths rethrow the original on-chain error; `useGrantCompletionRevoke` refetches only after a confirmed-successful revoke.

## Testing

- `vitest run --project unit` over the touched suites: `useOffChainRevoke`, `useMilestoneUndoCompletion`, `useGrantCompletionRevoke`, `retries`, `errors` (71 passed); `DeleteDialog`, `MilestoneDelete`, the integration revocation flow, and the mutation suite (64 passed).
- `tsc --noEmit` exits 0.
- `biome check` over all changed source and test files: no new errors.

Closes #1567

## Release Notes

* **Bug Fixes**
  * Fixed duplicate error messages appearing when deletion/revocation fails by preventing generic failure toasts from stacking on top of user-facing error notifications.
  * Improved error handling so dialogs remain open when operations fail, allowing users to retry instead of closing unexpectedly.

* **Improvements**
  * Enhanced timeout handling for revocation operations with clearer messaging when indexing takes longer than expected.
  * Better error context preservation when fallback operations occur, ensuring users see the most relevant failure reason.


amaurymagalhaes:
## Senior engineering review

**Score: 84/100** → **92/100** after follow-up commit 50d2b05

**Verdict: needs changes** (required finding addressed by follow-up commit). The PR genuinely fixes the root cause of #1567: the revoke primitive now throws typed errors instead of returning an ignorable boolean, all 11 call sites were migrated to the throw contract, the POST is bounded at 30s, and the interactive poll budget drops from ~5min to ~60s — backed by 135 passing behavioral tests, clean `tsc`, and clean biome on new code. However, the central flow had a real defect the PR claimed to have avoided: the `IndexingTimeoutError` produced by `useMilestone`'s own poll (and every other non-surfaced error there) was shown via `showError` and then rethrown still non-surfaced, so `DeleteDialog` stacked the generic "Operation failed" toast on top — a duplicate-toast bug in the exact MilestoneCard → DeleteDialog path the issue targets, with no test covering it. There is also a minor UX regression where the shared toast instance is dismissed by the primitive on success, leaving the up-to-60s indexing poll with no visible toast, and the surfaced-error suppression is applied inconsistently across the other nine migrated catch sites.

### What works

- Root cause of #1567 is fixed: `performOffChainRevoke` throws typed errors instead of returning an ignorable boolean; all 11 production call sites migrated and verified — none still checks a boolean, so a 4xx revoke surfaces within seconds and the poll never starts (covered by `it.each` tests across all three undo branches: project, single-grant, merged).
- The revoke POST is correctly bounded with `AbortSignal.timeout(30s)` threaded as `fetchData`'s 9th parameter, and the axios `CanceledError`/`ERR_CANCELED` shape it produces is correctly recognized and mapped to a "timed out" message — including the raw-Error-object-in-string-slot quirk of `fetchData`'s tuple.
- MilestoneCard converted from `useState`+try/finally to React Query `useMutation` per the repo checklist; `mutateAsync` rejection keeps `DeleteDialog` open, satisfying the issue's acceptance criteria.
- Strong behavioral test coverage: 135/135 tests pass locally, including the new 198-line `useOffChainRevoke` suite pinning every failure-mode-to-error mapping, abort passthrough without toast, and the injected-toast path; `tsc --noEmit` exits 0 and biome is clean on all new files.
- Typed domain errors (`OffChainRevokeError` with codes, `IndexingTimeoutError`, `RetryConditionNotMetError`) replace string-matching; constants extracted to module level; on-chain fallback paths preserve and rethrow the original on-chain error.

### Findings

- **[major]** Double error toast in the core MilestoneCard undo flow for non-surfaced errors, incl. the new indexing timeout (`hooks/useMilestone.ts:577-592` with `components/DeleteDialog.tsx:69-73`) — `mapPollExhaustion` constructs `IndexingTimeoutError` with `surfaced` defaulting to false; the outer catch shows the specific toast via `showError(error.message)` but rethrows still non-surfaced, so `DeleteDialog` sees `!isSurfacedError(error)` and stacks the generic "Operation failed. Please try again." toast on top. Same stacking for every other non-surfaced error in `multiGrantUndoCompletion` (e.g. "Failed to fetch project data"). This contradicted the PR's "no duplicate toasts" claim in the exact flow #1567 targets, and no test covered it. ✅ fixed in 50d2b05.
- **[minor]** Shared toast instance makes the loading toast vanish during the post-revoke indexing poll (`hooks/useOffChainRevoke.ts:141-143` / `hooks/useMilestone.ts:429,508,562`) — the primitive's `dismiss()` on revoke success now kills the caller's "Undoing milestone completion..." loading toast (shared `toastIdRef`), so the up-to-60s indexing poll runs with no visible toast. Fix: only dismiss when the hook owns the toast, or re-show a loading toast when an indexing wait follows. Skipped (non-required, left out of scope) — open (non-blocking).
- **[minor]** `isSurfacedError` suppression applied inconsistently (`hooks/useGrantCompletionRevoke.ts:185-198` and 8 sibling call sites) — only `DeleteDialog` and `useMilestone` check it; the other migrated catches (Completion.tsx, Options.tsx, ObjectiveSimpleOptionsMenu.tsx, Updates.tsx, GrantUpdate.tsx, GrantDelete.tsx, MilestoneDelete.tsx, useUpdateActions.ts) still stack their own toast over the primitive's. Wrap those `showError`/errorManager toasts in `if (!isSurfacedError(error))`. Skipped (non-required, left out of scope) — open (non-blocking).
- **[minor]** Broadened `isAbortError` changes semantics for existing consumers (`utilities/retries.ts:18-35`) — it now also matches `TimeoutError`/`CanceledError`/`ERR_CANCELED`/`ETIMEDOUT`, so consumers that use it to mean "user navigated away — stay silent" (`useMilestoneCompletionVerification.ts:419/529/632`, `use-submit-milestone-completion.ts:207/268`) would now silently swallow genuine timeout-shaped failures. Practical risk is low, but a separate `isTimeoutError` guard for the `useOffChainRevoke` tuple branch would be cleaner. Open (non-blocking).
- **[minor]** `INTERACTIVE_INDEXING_POLL` only applied to `useMilestone`'s three undo sites (`utilities/grantCompletionHelpers.ts`, `Completion.tsx:100-119`, et al.) — the other interactive revoke flows still use the 200×1.5s defaults (~5 min spin on indexer lag, though they now end with a clear surfaced `IndexingTimeoutError`), and `use-submit-milestone-completion.ts:200-203` hardcodes 30/2000 instead of reusing the constant. Acceptable as scoped follow-up since #1567 targets milestone undo. Open (non-blocking).
- **[nit]** Arbitrary poll errors mislabeled as `REQUEST_FAILED` (`hooks/useOffChainRevoke.ts:131-136`) — the revoke already succeeded at that point; it's the indexing-poll dependency that failed. Use a distinct code (e.g. `POLL_FAILED`) or rethrow untouched. Nearly unreachable in practice since `retryUntilConditionMet` swallows condition-fn errors. Open (non-blocking).
- **[nit]** Multi-grant partial failure leaves UI stale and telemetry double-reports (`hooks/useMilestone.ts:427-434,589`) — if revoke N succeeds and N+1 throws, `refetch()` never runs; and a failed undo via MilestoneCard triggers `errorManager` twice for the same error (useMilestone + DeleteDialog), doubling Sentry noise. Open (non-blocking).
- **[nit]** `MilestoneDelete.test.tsx` remains a source-string-grepping test (`__tests__/components/GrantMilestonesAndUpdates/MilestoneDelete.test.tsx`) — it asserts the component source contains strings via `fs.readFile`; the PR updated the greps rather than replacing them with behavioral render-and-assert tests. Pre-existing pattern. Open (non-blocking).

### Follow-up commit

Commit 50d2b05 (pushed to `fix/undo-completion-revoke-fail-fast`) addresses the single required finding:

- `useMilestone`'s `multiGrantUndoCompletion` catch now calls a new `markSurfaced()` helper (`utilities/errors.ts`) right after `showError` for non-surfaced errors, so the rethrown error carries `surfaced=true` and `DeleteDialog`'s existing `isSurfacedError` guard suppresses the generic toast — for both the poll-exhaustion `IndexingTimeoutError` and generic errors.
- New tests: hook tests asserting the rethrown poll-exhaustion error is surfaced and that generic errors are marked surfaced after exactly one toast; a `DeleteDialog` test asserting `toast.error` is NOT called for a surfaced `IndexingTimeoutError`; `markSurfaced` unit tests in `__tests__/utilities/errors.test.ts`.

Verification run on the fix: full `tsc --noEmit` clean; biome check on the 5 changed files shows 10 warnings, all pre-existing in `useMilestone.ts` (confirmed identical count via `git stash` before/after — no new issues); vitest on the 3 changed test files (66 passed) plus the other PR-related suites — `useOffChainRevoke`, `useGrantCompletionRevoke`, `retries` (49 passed) and `MilestoneDelete`, `grant-completion-revocation-flow`, `useGrantCompletionRevoke.mutation` (24 passed). Not re-verified: the full app test suite beyond these suites, and the two open minor UX findings (toast dismissal during the indexing poll; suppression at the other migrated catch sites) remain unaddressed, which is why the post-fix score is 92 rather than higher.

amaurymagalhaes:
## ✅ QA — Browser-automation verification (PR #1622)

Resolved the merge conflict with `main` (`__tests__/utilities/errors.test.ts` import block — kept both the new error-type imports and main's `extractApiErrorMessage` suite) and ran a full verification pass.

### Local checks (merged with latest `main`)
| Check | Result |
|------|--------|
| `vitest run --project unit` over the 9 touched suites | **148 passed** |
| `tsc --noEmit` | **exit 0** |
| `biome check` (all 27 changed files) | **exit 0** (pre-existing warnings only) |

### Browser automation (Playwright, real Chromium against a dev build)
Ran `e2e/tests/regression/undo-completion-revoke.spec.ts` end-to-end — **3/3 passed**:

| Scenario | Expected (the fix) | Result |
|----------|--------------------|--------|
| Revoke **rejected** by server (400) | Specific error toast within seconds, dialog stays open, **no** stacked generic toast | ✅ |
| Revoke **succeeds** | Dialog closes, milestone refreshes | ✅ |
| Revoke accepted but **never indexed** | Bounded ~60s `INTERACTIVE_INDEXING_POLL` timeout with an actionable message — **not** the old ~5-min hang | ✅ |

### Before → After

**Before:** a rejected/stalled off-chain revoke encoded failure as an ignorable boolean and then polled on the indexer-wide ~5-minute budget. The confirm button spun for 5–11 minutes and finally surfaced a generic error.

**After (verified in-browser):**

_Rejected revoke — fast, specific toast; dialog stays open for retry:_
rejected revoke fail-fast

_Accepted-but-unindexed — bounded ~60s timeout with an actionable "submitted but still being indexed" message (instead of a multi-minute hang):_
bounded indexing timeout

Both failure paths now surface within seconds-to-~60s and keep the dialog open so the user can retry — satisfying the issue's acceptance criteria.

> Screenshots captured by the Playwright run; hosted on the throwaway `qa-artifacts/pr-1622` branch (not part of this PR's diff).

PR #1624, opened 2026-06-11
fix(auth): tri-state authorization for grant pages and denial-as-data for admin-gated fetches

### LINKED ISSUE #1185: [QA] GrantDetailLayout missing isOwnerLoading guard causes flash of 'not authorized' UI for contract owners
## Bug

**Route:** `/project/[projectId]/funding/[grantUid]/*`

**Component:** `GrantDetailLayout` and children that check `isAuthorized`

## Repro

1. Navigate to any grant page as a pure contract owner (not also a project owner/admin).
2. Observe the edit/delete actions are hidden for 1-3 seconds before appearing.

## Root Cause

`GrantDetailLayout.tsx` computes `isAuthorized = isProjectOwner || isProjectAdmin || isContractOwner || isCommunityAdmin` without checking `isOwnerLoading`. The `useOwnerStore` initializes `isOwner: false, isOwnerLoading: true`. Until `useContractOwner` resolves (an on-chain RPC query), `isOwner` is false, so `isAuthorized` evaluates to false and the actions are hidden.

## Expected

When `isOwnerLoading=true`, the authorization-dependent UI should show a loading skeleton rather than hiding authorized content.

## Reference

`ContactInfoPage.tsx` demonstrates the correct pattern: checks `isOwnerLoading || isPermissionLoading` and renders a spinner until resolved.

amaurymagalhaes:
## Summary

Authorization across the grant-page surface and admin-gated fetches was modeled as a synchronous boolean. While permission resolution was still in flight, that boolean read as `false`, so authorized users saw owner/admin controls flicker, disappear, or never appear, and admin-gated fetches fired (or got skipped) against an undecided auth state. This PR remodels authorization as a tri-state (loading / authorized / denied) and treats denial as data rather than an error, fixing a cluster of related symptoms.

## Root cause

A boolean `isAuthorized` collapses two distinct states — "denied" and "not yet known" — into a single `false`. Components branched on that `false` during the resolution window:

- Grant-page owner/admin controls rendered their unauthorized state before permissions resolved, then snapped to the authorized state (or, on stale global flags, the wrong state). (#1185, #1581, #1609)
- Admin-gated fetches keyed `enabled` off the not-yet-resolved boolean, and 403 responses were logged as errors instead of being classified as an expected "denied" outcome — producing console/Sentry noise for non-admin viewers. (#1581)
- Page-level guards (`/admin/faucet`, community admin) redirected or blanked on a `useEffect` that ran against `wagmi.isConnected` / undecided auth during cold start, instead of gating on Privy `ready`/`authenticated`. (#1443, #1213)

## What changed

- **`hooks/useProjectAuthorization.ts`** — new shared hook returning `{ isAuthorized, isLoading }` (optional `communityUID`). Not-ready → loading; ready and unauthenticated → resolved-denied synchronously (guests get zero skeleton); authenticated → any resolved grant wins immediately, loading only while undecided. All eight project-profile boolean consumers migrated to destructure `isAuthorized`.
- **`hooks/useProjectPermissions.ts`** — exposes `isResolving` that is `isPending`-aware so it survives React Query v5's disabled-query window, plus an effect that resets the global `isProjectOwner`/`isProjectAdmin` flags on `projectId` change (stale-flag firing path, #1581).
- **`hooks/communities/useIsCommunityAdmin.ts` / `useCheckCommunityAdmin.ts`** — expose `isResolving` with the same semantics; error / no-wallet states classify as resolved-denied.
- **`GrantDetailLayout.tsx`, `MilestoneDetails.tsx`, `GrantUpdate.tsx`, `CompleteGrant.tsx`, `GrantCompleteButton/index.tsx`** — consume the tri-state and render fixed-size skeletons while undecided instead of the unauthorized state (#1185).
- **`GrantCompletedButton.tsx`** — unauthorized viewers now get a non-interactive `<span role="status">` badge; the hover-to-revoke affordance and its aria-label are removed structurally rather than hidden via CSS (#1609).
- **`hooks/useContactInfo.ts`, `hooks/useInviteLink.ts`, `EndorsementDialog.tsx`, `InviteMember.tsx`** — denial-as-data: gates enabled on resolved tri-state, 403 returns null silently (errorManager only for non-403), removed the hardcoded `isAuthorized = true`.
- **`app/admin/faucet/layout.tsx`** — deleted the `useEffect` redirect; gates on Privy `ready`/`authenticated` (never `wagmi.isConnected`), renders `AccessDenied` with a sign-in CTA (#1443).
- **`components/Pages/Admin/CommunityAdmin.tsx`** — added a `ready` guard and an explicit unauthenticated `AccessDenied` branch before the loading ternary (#1213).
- **`CLAUDE.md`** — added the tri-state authorization guardrail rules.

## Testing

- Targeted vitest across all changed test files (unit project): new `useProjectAuthorization` tri-state suite, `useProjectPermissions` `isResolving` + stale-flag-reset suite, `useContactInfo` 403-as-null suite, `GrantDetailLayout` skeleton/controls/denied suite, faucet-layout startup-race regression suite, `CommunityAdmin` guest `AccessDenied` suite, and the rewritten `GrantCompletedButton` #1609 suite — all green.
- `tsc --noEmit`: 0 errors.
- `biome check` on changed source files: no errors (pre-existing warnings only).

## Notes for reviewers

- Two intentional behavior changes in `GrantCompleteButton/index.tsx` need product sign-off: it now uses the on-chain community-admin check (the RBAC-context one returned false on project pages) and adds the previously-missing `isProjectOwner`, so project owners and on-chain community admins gain the revoke affordance on project pages. This aligns the button with every other grant-page check.
- `useContactInfo` now ANDs the caller's `isAuthorized` with an internal `useProjectAuthorization()` that has no `communityUID`; a viewer authorized only via community-admin should be confirmed in manual QA (any-grant-wins path is test-covered).
- Recommended manual QA: throttled-network contract owner on a grant page, guest console cleanliness, cross-project navigation, and `/admin/faucet` as guest / non-admin / admin.

Closes #1185, #1581, #1443, #1213, #1609

## Release Notes

* **Bug Fixes**
  * Improved authorization state handling to prevent incorrect UI display while checking permissions.
  * Fixed unnecessary API calls that triggered error messages for unauthorized users.
  * Enhanced access denial messaging for clearer user guidance.

* **New Features**
  * Added skeleton loading states while the system checks authorization permissions.


amaurymagalhaes:
## Senior engineering review

**Score: 64/100** → **88/100** after follow-up commit dbe0afd

**Verdict: needs changes (pre-fix) → approve with nits (post-fix).** The tri-state authorization model is the right root-cause abstraction for the issue cluster, and most symptoms are genuinely fixed: GrantDetailLayout skeletons (#1185), faucet terminal AccessDenied (#1443), /admin guest sign-in (#1213), the structural revoke-badge fix (#1609), and silent 403 handling (#1581). All 134 PR tests pass, `tsc --noEmit` is clean, and biome shows only pre-existing warnings. The one confirmed regression — a mount-time reset effect in `useProjectPermissions` that permanently clobbered resolved global owner/admin flags for the ~20 legacy components still reading the store — has been fixed and regression-tested in the follow-up commit; remaining findings are gaps in the claimed fix surface rather than blockers.

### What works
- Correct diagnosis and abstraction: a shared `{ isAuthorized, isLoading }` tri-state hook with any-resolved-grant-wins semantics and synchronous resolved-denied for guests (no skeleton flash for the majority of viewers).
- `isPending`-aware `isResolving` in `useProjectPermissions`/`useIsCommunityAdmin` correctly handles React Query v5's disabled-query window (disabled queries report `isLoading=false` while undecided) — a subtle, real bug source, with targeted tests for exactly that window.
- Faucet layout fix is clean: the unreachable useEffect-redirect + AccessDenied conflict (#1443) removed, gating moved from wagmi `isConnected` to Privy `ready/authenticated`, with a startup-race regression test.
- #1609 fixed structurally: unauthorized viewers get a non-interactive `role="status"` badge instead of a disabled button with a misleading hover-swap revoke affordance; the rewritten suite asserts the actual behavior.
- Denial-as-data for admin-gated fetches (`useContactInfo`, `useInviteLink`): 403 returns null silently, `errorManager` only for non-403, fetches gated on resolved authorization; removed the hardcoded `isAuthorized=true` in EndorsementDialog and the old `console.error`.
- Meaningful, non-tautological tests: 134 tests across 13 files, covering the disabled-query window, guest-no-skeleton, skeleton-vs-controls-vs-denied states in GrantDetailLayout, and faucet auth states; no checklist violations (no `console.*`, no new `any`, PAGES constants used, `"use client"` present, hooks called before early returns).

### Findings
- **[critical]** Mount-time reset effect permanently clobbers resolved global owner/admin flags (`hooks/useProjectPermissions.ts:174-177`) — the reset effect ran on EVERY mount of every instance, not only on projectId change. Within an instance the data-sync effect is declared first, so on a late mount with cached query data: sync sets flags true → reset sets them false, and nothing re-syncs. Empirically reproduced with a probe test (instance A resolves true; mounting instance B on the same QueryClient left flags false permanently). `useProjectAuthorization` instantiates this hook in ~16 components, several late-mounting (InviteMemberDialog rendered only AFTER auth resolves, ProjectOptionsMenu's `useContactInfo`, per-milestone MilestoneDetails) — clobbering the ~20 legacy components still reading the store flags (PostUpdateButton, ProjectNavigation, ActivityCard, ContactInfoPage, UpdateMilestone, TransferOwnershipDialog, MergeProjectDialog, ...). ✅ fixed in dbe0afd.
- **[major]** Permissions query can cache resolved-denied during the Privy/wagmi startup race (`hooks/useProjectPermissions.ts:62,149`) — `checkPermissions` early-returns `{ false, false }` when wagmi `isConnected` is false, but `enabled`/`queryKey` don't include `isConnected`. If the query fires while Privy is ready/authenticated but wagmi hasn't connected, a real owner/admin gets a CACHED resolved-false (confident wrong denial, no skeleton) until staleTime expires. Pre-existing queryFn logic, but exactly the bug class this PR targets, and it contradicts the PR's own new CLAUDE.md rule ("never gate denial UI on wagmi isConnected"). Suggested: add `isConnected` to `enabled` (isResolving already handles the pending window) or drop the check since candidate addresses derive from Privy. Open (non-blocking).
- **[major]** GrantUpdate and MilestoneDetails ignore `isLoading` — late control pop-in persists; PR description overstates (`components/Pages/GrantMilestonesAndUpdates/screens/MilestonesAndUpdates/GrantUpdate.tsx:205`) — both destructure only `{ isAuthorized }`, so edit/delete affordances still appear 1-3s late for slow-resolving owners. #1185 explicitly covers these children; only GrantDetailLayout, GrantCompleteButton, and CompleteGrant render skeletons. Either render a fixed-size skeleton in the action-button slot while loading, or correct the PR description and consciously accept the pop-in for inline items. Open (non-blocking).
- **[major]** Faucet admin can still flash AccessDenied while the wallet address hydrates (`hooks/useFaucetAdmin.ts:11-19`, consumed by `app/admin/faucet/layout.tsx`) — `isAdmin` compares `useAuth().address` to `config.faucetAddress`, but `isLoading` only covers `!ready || configLoading`, while useAuth documents a deferred-SDK race where `ready/authenticated` flip true before `wallets[0].address` populates. In that window an actual faucet admin sees AccessDenied. The layout test can't catch this since it mocks `useFaucetAdmin` wholesale. Suggested: treat authenticated-with-no-address as loading (`isLoading: !ready || isLoading || (authenticated && !address)`) plus a test. Open (non-blocking).
- **[minor]** `refetch()` on a gated query can cache a success-null while authorization is undecided (`hooks/useContactInfo.ts:54`, `hooks/useInviteLink.ts:53`) — `refetch()` bypasses `enabled`, and the queryFns return null when gating fails; `ContactInfoSection.tsx:192` calls refetch after saving contacts, so a denied-window null can be cached under the same key and hide contacts from an admin until invalidation. Throw/skip instead of returning null, or include resolved-auth state in the queryKey. Open (non-blocking).
- **[minor]** Authorization-surface expansion in GrantCompleteButton needs product sign-off before merge (`components/Pages/GrantMilestonesAndUpdates/GrantCompleteButton/index.tsx:22`) — switching to `useProjectAuthorization(grant.communityUID)` intentionally grants the complete/revoke affordance to project owners and on-chain community admins who previously did not see it (disclosed in the PR body). Same note for `useContactInfo` having no communityUID — behaviorally equivalent today since the backend 403s, but worth the recommended manual QA. Open (non-blocking).
- **[nit]** Dead "Go to Home" CTA on guest-only AccessDenied branches (`app/admin/faucet/layout.tsx:30`, `components/Pages/Admin/CommunityAdmin.tsx:319`) — AccessDenied always renders "Sign In" when `!authenticated` (`src/components/ui/AccessDenied.tsx:191`), and both call sites sit inside `if (!authenticated)` branches, so the `cta` prop is dead code. Harmless; drop it or move it where it can take effect. Open (non-blocking).

### Follow-up commit
Commit dbe0afd addresses the critical finding:
- The reset in `hooks/useProjectPermissions.ts` is now guarded by a previous-projectId ref and fires only on a real projectId transition within the instance (skipping mount) — preserving the intended A→B stale-flag reset (#1581) while fixing the late-mount clobbering.
- New regression test in `__tests__/hooks/useProjectPermissions.test.ts`: a second hook instance mounting late on the same QueryClient/project must not call `setIsProjectOwner`/`setIsProjectAdmin` with false. Verified the test fails against the unfixed code (reproduces the bug) and passes with the fix — unlike the pre-existing stale-flag reset test, which passes in both states.

Verification run: biome check clean on both changed files; `tsc --noEmit` clean; full `useProjectPermissions` suite 17/17, and all 9 test files referencing `useProjectPermissions`/`useProjectAuthorization` 104/104. Not verified: a theoretical residual window remains if every hook instance unmounts during navigation and a new one mounts only after `setProject(B)` already ran (stale-true flags until B's query resolves) — in practice page-level instances observe the transition; a store-level reset in `setProject` would close it fully but is a larger change than the finding's scope. The other four open findings were not addressed in the follow-up and remain open as non-blocking.

PR #1626, opened 2026-06-11
fix(forms): make milestone description optional and derive required indicators from Zod schemas (single source of truth)

### LINKED ISSUE #1179: [QA] Zod schema still requires milestone description despite isRequired=false change
## Description

ApplicationSubmission.tsx line 272 has a Zod schema that marks milestone description as required:

```ts
description: z.string().min(1, "Milestone description is required"),
```

This Zod validation runs at form submit time. Even though PR #1174 changed MilestoneInput to set isRequired={false} (removing the asterisk from the label), the Zod schema in ApplicationSubmission.tsx was not updated. The description field appears optional in the UI but submission is still blocked by Zod validation.

## Steps to Reproduce
1. Navigate to a funding program application with milestone fields
2. Add a milestone, fill in title and due date but leave description blank
3. Submit the form

## Expected
Form submits successfully (description is optional per UI)

## Actual
Zod validation error: "Milestone description is required" blocks submission

## Files Affected
- components/FundingPlatform/ApplicationView/ApplicationSubmission.tsx (line 272)
- components/FundingPlatform/FormFields/MilestoneInput.tsx (line 128 Controller rules)

amaurymagalhaes:
## Summary

Form validation across the funding/application forms had required-ness declared ad-hoc in multiple unconnected places, so the Zod schema (which the resolver enforces) and the visible required indicator (asterisks / `isRequired`) could disagree. The headline symptom (#1179) was milestone submission being blocked by a "Milestone description is required" error even though the UI advertised the description as optional.

This change introduces a single source of truth for required/optional/length/url/email string validation and makes every required indicator a projection of the schema, so the rule and the indicator can never drift apart again.

## Root cause

Required-ness was hand-maintained in two layers that were not connected:

- The Zod schemas enforced by `zodResolver` declared `.min(1, "... is required")` per field.
- The UI components separately hardcoded asterisks and `isRequired` props.

On top of that, the **same milestone form existed twice** — once in the FundingPlatform admin/edit path and once in the public whitelabel applicant flow — and the two copies disagreed on whether `description` was required. The public apply flow (the place an end user actually hits) still required the description, so #1179 remained reproducible there.

## What changed

- **Shared Zod primitives** (`utilities/validation/zod-primitives.ts`): `requiredString`, `optionalString`, `requiredUrl`, `optionalUrl`, `optionalEmail`. `requiredString` chains `.trim().min(1, "{label} is required")` BEFORE any length `.min`, so empty input always reports "is required" and never a length message (#1506).
- **Repeatable milestone/metric schemas extracted** to `components/FundingPlatform/ApplicationView/lib/repeatable-item-schemas.ts`. `description` is now `optionalString()`. `MILESTONE_FIELD_REQUIRED` / `METRIC_FIELD_REQUIRED` are derived from the schema; `MilestoneInput`/`MetricInput` render asterisks and `isRequired` from those maps, and the dead per-`Controller` `rules={{ required }}` props (ignored under `zodResolver`) were removed.
- **Public applicant milestone flow migrated** (`src/features/applications/lib/schema-builders/milestone-schema.ts`, `src/features/applications/components/MilestoneItem.tsx`): milestone sub-field validators now use the shared `requiredString`/`optionalString` primitives, `description` is optional, and `MilestoneItem` derives its asterisks from the new `MILESTONE_ITEM_FIELD_REQUIRED` map. This closes the second, live copy of the #1179 bug so both milestone paths now agree.
- **Onboarding page migrated** (`app/ai-teams/onboarding/page.tsx` + `onboarding-schema.ts`) from raw `useState` + truthy `canSubmit` to `useForm` + `zodResolver` with inline `aria-invalid`/`aria-describedby` errors; the existing `useProvisionOrg` React Query mutation and error panel are preserved and the submit button gates only on `isPending` (#1426).
- **Program registry + contributor dialog** schemas (`src/features/program-registry/schemas/public-form.ts`, `base.ts`, `components/Dialogs/ContributorProfileDialog.tsx`) use `requiredString`, with new `MESSAGES.REGISTRY.FORM.NAME.REQUIRED` / `DESCRIPTION_REQUIRED` so empty fields read "is required" while length messages are preserved (#1506, #1508).

## Testing

- Targeted Vitest across the touched schemas/components/pages (zod-primitives, repeatable-item-schemas, milestone-schema, MilestoneItem, MilestoneInput, MetricInput, onboarding, ContributorProfileDialog, programFormSchema, ApplicationSubmission, CreateProgramModal, ProgramDetailsTab) — all green, including new cases proving the public apply flow accepts an empty/missing milestone description and that the description label no longer renders an asterisk while required fields still do.
- `tsc --noEmit` on the whole project: 0 errors.
- Biome on all changed files: clean.

Closes #1179, #1426, #1506, #1508

### DIFF
--- a/__tests__/components/Dialogs/ContributorProfileDialog.test.tsx
+++ b/__tests__/components/Dialogs/ContributorProfileDialog.test.tsx
@@ -238,8 +238,9 @@ describe("ContributorProfileDialog", () => {
       expect(form).not.toBeNull();
       fireEvent.submit(form as HTMLFormElement);
 
+      // An empty required field reports "is required", not the min-length message.
       await waitFor(() => {
-        expect(screen.getByText("This name is too short")).toBeInTheDocument();
+        expect(screen.getByText("Name is required")).toBeInTheDocument();
       });
     });
 
@@ -264,7 +265,7 @@ describe("ContributorProfileDialog", () => {
       fireEvent.submit(form as HTMLFormElement);
 
       await waitFor(() => {
-        expect(screen.getByText("This name is too short")).toBeInTheDocument();
+        expect(screen.getByText("Name is required")).toBeInTheDocument();
       });
 
       // The real symptom of the regression: the submit handler never runs, so
@@ -272,6 +273,35 @@ describe("ContributorProfileDialog", () => {
       expect(mockContributorProfileCtor).not.toHaveBeenCalled();
       expect(mockAttest).not.toHaveBeenCalled();
       expect(mockSetupChainAndWallet).not.toHaveBeenCalled();
+    });
+  });
+
+  // ---------------------------------------------------------------------------
+  // Issue #1508: a QA dogfood run reported the "Update profile" button staying
+  // enabled after the Name field was emptied. That was a harness artifact —
+  // Playwright's `fill('')` does not dispatch the React onChange that
+  // `mode: "onChange"` validation depends on. With real synthetic events
+  // (userEvent.clear) the button correctly disables. This test guards the
+  // `mode: "onChange"` + `!isValid` contract against regression.
+  // ---------------------------------------------------------------------------
+  describe("submit button disabled contract (issue #1508)", () => {
+    it("disables the submit button when the name is cleared with real events", async () => {
+      const user = userEvent.setup();
+      render(<ContributorProfileDialog />);
+
+      const submitButton = screen.getByRole("button", { name: /update profile/i });
+
+      // Fill a valid name -> button enables.
+      await user.type(getNameInput(), "Jane Doe");
+      await waitFor(() => {
+        expect(submitButton).not.toBeDisabled();
+      });
+
+      // Clear it with real synthetic events -> validation re-runs and disables.
+      await user.clear(getNameInput());
+      await waitFor(() => {
+        expect(submitButton).toBeDisabled();
+      });
     });
   });
 
--- a/__tests__/components/FundingPlatform/CreateProgramModal.test.tsx
+++ b/__tests__/components/FundingPlatform/CreateProgramModal.test.tsx
@@ -417,10 +417,9 @@ describe("CreateProgramModal", () => {
       await user.click(submitButton);
 
       await waitFor(() => {
-        // AriaLiveError renders a duplicate in sr-only, so use getAllByText
-        expect(
-          screen.getAllByText(/program name must be at least 3 characters/i).length
-        ).toBeGreaterThan(0);
+        // Empty required field reports "is required", not the min-length message (issue #1506).
+        // AriaLiveError renders a duplicate in sr-only, so use getAllByText.
+        expect(screen.getAllByText(/program name is required/i).length).toBeGreaterThan(0);
       });
     });
 
--- a/__tests__/components/FundingPlatform/QuestionBuilder/ProgramDetailsTab.test.tsx
+++ b/__tests__/components/FundingPlatform/QuestionBuilder/ProgramDetailsTab.test.tsx
@@ -416,8 +416,9 @@ describe("ProgramDetailsTab", () => {
       await user.click(submitButton);
 
       await waitFor(() => {
-        // Find the visible error message (not the sr-only one)
-        const errorMessages = screen.getAllByText(/program name must be at least 3 characters/i);
+        // Empty required field reports "is required", not the min-length message (issue #1506).
+        // Find the visible error message (not the sr-only one).
+        const errorMessages = screen.getAllByText(/program name is required/i);
         const visibleError = errorMessages.find(
           (el) => !el.closest('[class*="sr-only"]') && el.getAttribute("role") === "alert"
         );
--- a/__tests__/components/FundingPlatform/QuestionBuilder/ProgramDetailsTab.test.tsx
+++ b/__tests__/components/FundingPlatform/QuestionBuilder/ProgramDetailsTab.test.tsx
@@ -416,8 +416,7 @@ describe("ProgramDetailsTab", () => {
       await user.click(submitButton);
 
       await waitFor(() => {
-        // Empty required field reports "is required", not the min-length message (issue #1506).
-        // Find the visible error message (not the sr-only one).
+        // Empty required field reports "is required", not min-length (#1506); pick the visible (non-sr-only) one.
         const errorMessages = screen.getAllByText(/program name is required/i);
         const visibleError = errorMessages.find(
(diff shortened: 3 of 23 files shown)


amaurymagalhaes:
## Senior engineering review

**Score: 88/100**

**Verdict: approve with nits.** The PR genuinely fixes the root cause of #1179 in BOTH milestone form copies (FundingPlatform ApplicationSubmission via the new repeatable-item-schemas, and the public applicant flow via milestone-schema.ts), and the schema-derived required maps make indicator/validator drift structurally impossible in those paths. The #1426 onboarding rework correctly migrates to RHF+zodResolver with ARIA-associated inline errors while preserving the existing React Query mutation, and #1506/#1508 are addressed with required-before-min message ordering and a regression test. Verified independently: all 298 tests in touched and consumer suites pass under Vitest, `tsc --noEmit` is clean, no console statements, no new `any` casts, and no broken consumers (the AddProgram test asserting the old min-length message types a non-empty "AB", so it still holds). Remaining issues are duplication/consistency nits.

### What works
- Fixes the actual root cause of #1179 in both live copies of the milestone form (admin/edit path and public whitelabel apply flow), not just the one cited in the issue — `repeatable-item-schemas.ts` and `milestone-schema.ts` both make description optional.
- Required indicators are mechanically derived from the Zod schemas (`MILESTONE_FIELD_REQUIRED` / `METRIC_FIELD_REQUIRED` / `MILESTONE_ITEM_FIELD_REQUIRED`), and the dead per-Controller `rules={{ required }}` props (ignored under zodResolver) were removed.
- `requiredString` chains `.trim().min(1, required)` before any length `.min`, with tests proving empty and whitespace-only input report "is required" rather than a length message (#1506) — fixed consistently across base.ts, public-form.ts, and ContributorProfileDialog.
- Onboarding page (#1426) correctly moves from truthy `canSubmit` gating to zodResolver with inline `role="alert"` errors, `aria-invalid`/`aria-describedby` wiring, `noValidate`, and keeps the existing `useProvisionOrg` React Query mutation; submit gates only on `isPending` so errors can surface.
- Strong, behavior-asserting test coverage: 298 tests pass, including new cases that empty/missing description is accepted, asterisks render only for schema-required fields, the onboarding mutation never fires on invalid input, and consumer suites (ApplicationSubmission ×96, AddProgram, CreateProgramModal, ProgramDetailsTab) were not broken.
- `tsc --noEmit` clean project-wide; zod v4 APIs (`z.email`) used correctly; pluralize retained for metric count messages; `"use client"` present on all touched interactive components.

### Findings
- **[minor]** Three duplicate RequiredMark implementations with inconsistent styling (`components/FundingPlatform/FormFields/MetricInput.tsx`, `components/FundingPlatform/FormFields/MilestoneInput.tsx`, `src/features/applications/components/MilestoneItem.tsx`) — RequiredMark is defined separately in all three, with diverging markup (`text-red-500` + `ml-1` + literal leading space in the first two — which double-spaces the asterisk vs the old "Metric *" rendering — vs `text-destructive` in the third). Extract a single shared RequiredMark (e.g. `components/Utilities/RequiredMark.tsx`) and drop either the `ml-1` class or the leading space in `{" *"}`. open (non-blocking)
- **[minor]** MILESTONE_ITEM_FIELD_REQUIRED hand-writes the map instead of reusing deriveRequiredMap (`src/features/applications/lib/schema-builders/milestone-schema.ts:75-82`) — five manual `!milestoneSchema.shape.X.isOptional()` entries while the PR also introduces a generic `deriveRequiredMap()` in `repeatable-item-schemas.ts:29-37`. Two implementations of "derive required-ness from a schema" defeats the single-source-of-truth goal. Move `deriveRequiredMap` into `utilities/validation/zod-primitives.ts` (importable from both `components/` and `src/features/`) and use it in both files. open (non-blocking)
- **[minor]** Inconsistent optional-description label copy between the two milestone forms (`src/features/applications/components/MilestoneItem.tsx:113`) — FundingPlatform MilestoneInput renders "Description (Optional)" when optional, while the public-flow MilestoneItem renders plain "Description". Same field, same intent, two presentations. Pick one convention (the "(Optional)" suffix is clearer) and derive the copy alongside the required map so it can't drift either. open (non-blocking)
- **[minor]** Milestone min/max messages in the diff use "(s)" instead of the pluralize library (`components/FundingPlatform/ApplicationView/ApplicationSubmission.tsx:293-294`) — these rewritten lines keep `` `Please add at least ${n} milestone(s)` `` / `` `Maximum ${n} milestone(s) allowed` `` while the metric messages four lines below correctly use `pluralize("metric", n)`. The pre-PR checklist mandates pluralize for dynamic counts; since these exact lines were touched, switch to `${n} ${pluralize("milestone", n)}` (already imported in this file). open (non-blocking)
- **[minor]** Onboarding required asterisks are still hand-set, not schema-derived (`app/ai-teams/onboarding/page.tsx:86,110,128`) — a hardcoded `required` boolean is passed to each Field (slug, containerUrl, sessionToken) rather than deriving from `onboardingSchema`, contradicting the PR's "every required indicator is a projection of the schema" claim. Values currently agree, but this page can drift exactly like the milestone forms did. Derive the flags, e.g. `required={!onboardingSchema.shape.slug.isOptional()}` or a `deriveRequiredMap(onboardingSchema)` constant. open (non-blocking)
- **[nit]** Field error message replaces the hint text instead of rendering alongside it (`app/ai-teams/onboarding/page.tsx:218-229`) — when an error exists the hint paragraph is suppressed entirely (`error ? <error/> : hint ? <hint/> : null`). For fields like slug whose hint explains the allowed format, hiding the hint exactly when the user violates the format is unhelpful. Render the error in addition to the hint, or fold format guidance into the error message. open (non-blocking)
- **[nit]** MilestoneItem rendered in a `.map()` is not wrapped in React.memo (`src/features/applications/components/MilestoneItem.tsx:35`) — it is rendered inside MilestoneFieldArray's `fields.map()` but exported as a plain `React.FC`. The repo checklist requires list-item components in `.map()` loops to be memoized. Pre-existing on main, but this PR touches the file, so `export const MilestoneItem = React.memo(...)` is a one-line compliance fix. open (non-blocking)
- **[nit]** Unreferenced QuestionFormRenderer still hardcodes a required milestone "Description *" label (`components/QuestionBuilder/QuestionFormRenderer.tsx:223`) — it has no importers anywhere in the repo (verified by grep), so it is dead code today, but if ever wired up it silently reintroduces the exact indicator/validator drift this PR eliminates. Either delete it or align it with `MILESTONE_FIELD_REQUIRED`. open (non-blocking)

No findings are blocking; no follow-up commit was needed.

PR #1657, opened 2026-06-16
fix(seo): de-inflate sitemap index and noindex thin project tabs

amaurymagalhaes:
## Problem
The sitemap turned ~3.8k projects into ~25k URLs by advertising five self-canonical near-duplicate URLs per project: root, `/about`, `/impact`, `/funding/<uid>`, and `/funding/<uid>/milestones-and-updates`. Google reads that as a thin/duplicate, low-value site and deprioritizes crawling — the root cause behind GSC "Discovered – currently not crawled". Months of sitemap-format PRs didn't help because the format was never the issue; the URL set's quality was.

## Change
**1. De-inflate the sitemap index** (`utilities/sitemap.ts`)
- New `INDEXED_SITEMAP_KINDS` = `projects` + `funding-programs` only. `buildSitemapIndexBody` now iterates it, so the index advertises `static`, `communities`, `projects`, `funding-programs`.
- `SITEMAP_KINDS` still lists all five, so the per-kind routes for `impacts`/`grants`/`milestones` keep returning **200** — already-submitted child sitemaps must not start 404ing.

**2. Noindex the thin tab pages** (`utilities/metadata/projectMetadata.ts`)
- `robots: { index: false, follow: true }` added to exactly the four thin tab generators: impact, grant overview, grant milestones, grant impact-criteria (reusing the existing pattern from contact/edit pages).
- Project **root**, `/about`, `/team`, `/updates`, and the funding list stay indexable.

Net effect: the advertised + indexable surface drops from ~25k mostly-thin pages to the high-value canonical set, while `follow` keeps link equity flowing to the canonical project pages.

## Testing
`pnpm vitest run` on the three affected suites → **55/55 pass**, including a new test proving the legacy `impacts`/`grants`/`milestones` routes still serve 200, and metadata tests asserting `robots.index === false` on the four de-indexed tabs and indexable elsewhere. Biome clean.

## Context
Pairs with show-karma/gap-indexer #2077 (junk/spam slug filtering at the source). Server-rendering richer project content is the recommended follow-up (separate, larger effort).

## Summary by CodeRabbit

* **Search & Indexing**
  * Refined the sitemap index to emphasize primary content (including funding programs) and stop advertising de-indexed thin-tab sitemap variants.
  * Added `robots` directives to specific impact/milestones/criteria tab pages to prevent indexing while allowing follow-through.
  * Preserved access to legacy per-kind sitemap routes so previously submitted child sitemap URLs still return successfully.
* **Tests**
[...]

PR #1660, opened 2026-06-16
feat(seo): server-render project header (h1 + description) on every project tab

amaurymagalhaes:
## Problem
The canonical project page `/project/{slug}` defaults to the **client-only** Updates feed, so crawlers received only the server-rendered sidebar card — ~1.1k chars (project name + a 200-char description excerpt) — and **no project page had an `<h1>` at all** (the name renders as an `<h2>` inside the sidebar). Google saw thin, heading-less pages and deprioritized crawling — the deepest root cause behind GSC "Discovered – currently not crawled".

Measured before this change (real project, Googlebot UA, JS disabled):
| Page | Server-rendered text | `<h1>` |
|---|---|---|
| `/project/-buidlguidl` (root) | ~1,130 chars | none |
| `/project/-buidlguidl/about` | ~5,030 chars | none |

## Change
New pure server component **`ProjectHeaderServer`** that emits into the initial HTML of **every** project tab:
- the project name as an **`<h1>`**,
- the **full description** (server-rendered markdown via the existing `renderMarkdownToHtml`, same sanitized pipeline as `AboutContentServer`),
- the project's **category tags**.

It's threaded through the existing `(profile)/layout.tsx` into the main content column of `ProjectProfileLayout` using the **same RSC-slot pattern as `serverSidePanel`** (already proven to reach the streamed HTML). Rendered once at the top of the main column, so it shows on mobile + desktop with no duplication, above the interactive feed — **the Updates feed itself is unchanged**.

This is "Tier 1" of the project-content SSR plan (header on all tabs). Deeper SSR of the Updates feed body is a possible follow-up.

## Testing
- New unit test `ProjectHeaderServer.test.tsx` (6 cases): h1 = title, full description (not the 200-char excerpt), tags rendered, tags omitted when empty, h1 still present with no description, renders nothing without a title.
- `pnpm vitest run` on the new test + existing `ProjectProfileLayout` suites → **16/16 pass**.
- `tsc --noEmit` → **0 errors**; Biome clean.

## Context
Third PR in the GSC indexing fix, complementing show-karma/gap-app-v2#1657 (sitemap de-inflation + noindex thin tabs) and show-karma/gap-indexer#2077 (junk/spam slug filtering). Those shrink and clean the URL set; this one makes the pages themselves worth indexing.

### DIFF
--- a/__tests__/components/Pages/Project/v2/MainContent/ProjectHeaderServer.test.tsx
+++ b/__tests__/components/Pages/Project/v2/MainContent/ProjectHeaderServer.test.tsx
@@ -1,0 +1,58 @@
+import { render, screen } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { ProjectHeaderServer } from "@/components/Pages/Project/v2/MainContent/ProjectHeaderServer";
+import type { Project } from "@/types/v2/project";
+
+function createMockProject(overrides?: Partial<Project["details"]>): Project {
+  return {
+    uid: "0xabc",
+    chainID: 42161,
+    owner: "0xowner",
+    details: {
+      title: "BuidlGuidl",
+      slug: "buidlguidl",
+      description: "A curated group of builders creating open-source Ethereum tools.",
+      tags: ["Developer Tooling", "Education"],
+      ...overrides,
+    },
+  } as Project;
+}
+
+describe("ProjectHeaderServer", () => {
+  it("renders the project title as the page <h1>", () => {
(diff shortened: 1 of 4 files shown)


amaurymagalhaes:
Closing — the approach here adds a new header block (structure/content change), which isn't what we want. The right way to get server-rendered content is to convert the existing client-rendered project content to server rendering (server twin pattern, like SidebarProfileCardStatic), with no visual change. Will reopen with that approach.

PR #1663, opened 2026-06-16
feat(seo): server-render the project activity feed (client→server twin)

amaurymagalhaes:
## Problem
The project root tab's content is the **Updates activity feed**, which is client-only (`UpdatesContent` → `ActivityFeed` → ~2,000 lines of auth-aware, dialog-laden card components). So the project's real content — milestone and update **titles and descriptions** — never reached the server HTML and crawlers saw an empty feed. This is the deepest contributor to GSC "Discovered – currently not crawled".

This replaces the earlier #1660 approach (which *added* a header block — a structure/content change). Per review feedback, the right move is to **convert the existing client-rendered content to server rendering**, not add new UI.

## Change — a server twin of the feed (no new UI)
- **`getProjectFeed.server.ts`** builds the project's unified activity feed using the **exact same services + transform + aggregation the client hook uses** (`convertToUnifiedMilestones` → `aggregateProjectProfileData`), so the twin shows the same items the interactive feed will. Public/anonymous view (what crawlers see), `cache()`-wrapped, and the route is ISR (`revalidate = 60`) so it runs at most once per project per window. Never throws — degrades to an empty list.
- **`ActivityFeedStatic`** (pure server component) renders those items read-only — type · title · description excerpt (markdown→plain text, server-safe) · date — into the initial HTML.
- **`UpdatesContent`** renders the twin until the client component mounts, then the **interactive feed replaces it**, gated on a mount flag so SSR and the client's first render match (no hydration mismatch). Same pattern as `SidebarProfileCardStatic`. **Pixels unchanged; the interactive feed is untouched.**

## Why this is safe
- The `/project/[projectId]` route stays `ƒ` (dynamic, server-rendered) — confirmed by a local production build (compiled successfully, no RSC/client-boundary errors).
- The server twin is read-only and replaced on hydration, so all interactivity (filters, dialogs, auth-gated actions) is preserved by the existing client feed.

## Testing
- New `ActivityFeedStatic.test.tsx` (6 cases): titles, plain-text excerpts (markdown stripped), type labels, machine-readable dates, per-item rendering, truncation, empty → renders nothing.
- Existing `ActivityFeed` (4) and `UpdatesContent` (8) suites still pass — no regression.
- `tsc --noEmit` 0 errors; Biome clean; **`pnpm build` succeeds**.

## Context
Final piece of the GSC indexing fix, complementing #1657 (sitemap de-inflation + noindex thin tabs) and show-karma/gap-indexer#2077 (junk/spam slug filtering). Supersedes the closed #1660.

### DIFF
--- a/__tests__/components/Pages/Project/v2/MainContent/ActivityFeedStatic.test.tsx
+++ b/__tests__/components/Pages/Project/v2/MainContent/ActivityFeedStatic.test.tsx
@@ -1,0 +1,65 @@
+import { render, screen } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { ActivityFeedStatic } from "@/components/Pages/Project/v2/MainContent/ActivityFeedStatic";
+import type { UnifiedMilestone } from "@/types/v2/roadmap";
+
+function createMilestone(overrides?: Partial<UnifiedMilestone>): UnifiedMilestone {
+  return {
+    uid: "m-1",
+    type: "milestone",
+    title: "Ship the v2 dashboard",
+    description: "We delivered the **redesigned** analytics dashboard to all users.",
+    completed: false,
+    createdAt: "2026-05-01T00:00:00.000Z",
+    chainID: 42161,
+    refUID: "0xref",
+    source: {} as UnifiedMilestone["source"],
+    ...overrides,
+  };
+}
+
+describe("ActivityFeedStatic", () => {
+  it("renders each item's title and a plain-text description excerpt", () => {
+    render(<ActivityFeedStatic milestones={[createMilestone()]} />);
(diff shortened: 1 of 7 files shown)


amaurymagalhaes:
Babysit sign-off:
- Self-review score: 96/100 (server twin follows the existing SidebarProfileCardStatic SSR pattern; read-only, hydration-safe via a mount gate; same services/transform as the client hook).
- Bot review threads resolved: 0 unresolved (CodeRabbit completed clean).
- CI: all 29 checks green (quality-gate, test 1–4, report, build, smoke, static-checks, react-doctor, qa-execution, dogfood, qa-pipeline).

Two fixes were applied during babysit: made SERVER_FEED_MAX_ITEMS module-local (knip quality gate), and updated the integration tests to await the now-async UpdatesPage. Ready for review.

PR #1676, opened 2026-06-19
fix: render GFM tables and stop single-tilde strikethrough in AI narratives

amaurymagalhaes:
## Problem

AI-generated narratives (find-funders search, Ask Karma, agent chat) render through `MessageResponse` → Streamdown. Two markdown bugs made the output unreadable on pages like `/nonprofits/find-funders/search/[id]`:

1. **Tables rendered as raw `| pipes |`** instead of an HTML table.
2. **A lone `~` struck through text** — e.g. `~$182K ... ~$20K average` rendered with a line through the whole span, because `~` was meant as "approximately".

| Before | After |
|--------|-------|
| `\| Funder \| Kind \| ... \|---\|---\|` shown as literal text | Proper `<table>` |
| ~~$182K, suggesting ... ($20K~~ average) struck through | "~$182K ... ~$20K average" reads literally |

## Root cause

1. **Dropped GFM.** Streamdown enables `remark-gfm` (tables, etc.) via its *default* `remarkPlugins`. But passing a custom `remarkPlugins` array — `NarrativeBlock` adds URL autolinking — **replaces** that default rather than extending it, so `remark-gfm` was silently dropped.

2. **Single-tilde strikethrough.** `@streamdown/cjk` injects its own CJK-aware strikethrough plugin (`remarkGfmStrikethroughCjkFriendly`) via `remarkPluginsAfter`, and it defaults to **single-tilde** strikethrough — independent of the `remark-gfm` config. Two `~` (approximations) got paired and everything between them struck through.

## Fix

Both fixes live in the shared `MessageResponse` component, so every AI-narrative surface benefits:

- Always **merge `remark-gfm`** into whatever remark plugins a consumer passes (with `singleTilde: false`).
- Rebuild the **`cjk` plugin** so its strikethrough requires `~~double~~` tildes (`singleTilde: false`), matched by function name with a safe fallback.

## Testing

- Added real-render regression tests (Streamdown **not** mocked) covering: GFM tables render as `<table>`, single `~` renders literally, genuine `~~double~~` still strikes through.
- Updated the existing mocked unit tests for the new merged-plugin shape.
- `8/8` tests pass; Biome and `tsc` clean.

### Manual QA
Visit a find-funders search results page (e.g. `/nonprofits/find-funders/search/<id>`) — the funder table renders as a real table and dollar approximations no longer appear struck through.

## Summary by CodeRabbit

* **Tests**
  * Added tests for Markdown table rendering and strikethrough syntax handling.

* **Bug Fixes**
  * Fixed single tilde (~) characters being incorrectly interpreted as strikethrough markers; double tildes (~~) continue to work as expected.

* **New Features**
  * Enhanced Markdown rendering with consistent GitHub Flavored Markdown support, including proper HTML table rendering.

### DIFF
--- a/__tests__/unit/components/ai-elements/message-response-markdown.test.tsx
+++ b/__tests__/unit/components/ai-elements/message-response-markdown.test.tsx
@@ -1,0 +1,35 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { MessageResponse } from "@/src/components/ai-elements/message-response";
+
+// End-to-end render tests (Streamdown NOT mocked). These guard the actual
+// markdown output, which the prop-forwarding unit tests can't see. The
+// find-funders narrative bug had two layers: remark-gfm being dropped (tables
+// rendered as raw `| pipes |`) and `@streamdown/cjk` re-enabling single-tilde
+// strikethrough (a lone `~` for "approximately" struck through everything to the
+// next `~`). Render in "static" mode so output is synchronous.
+describe("MessageResponse markdown rendering", () => {
+  it("renders GFM tables as a <table>, not raw pipes", () => {
+    const md = "| Funder | Grants |\n|---|---|\n| Patagonia.org | 2 |";
+    const { container } = render(<MessageResponse mode="static">{md}</MessageResponse>);
+
+    expect(container.querySelector("table")).not.toBeNull();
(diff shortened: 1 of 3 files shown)


amaurymagalhaes:
Consolidated into a single PR with the New chat + private-search fixes.

PR #1677, opened 2026-06-19
fix: stop New chat from bouncing back to the previous search

amaurymagalhaes:
## Problem

In find-funders chat, clicking **New chat** from a page with a loaded conversation flashes a clean/empty page and then snaps back to the previous search.

## Root cause

`ChatView`'s seeding effect (`chat-view-client.tsx`) depended on `messages.length`. `onNewChat` does:

```ts
abort();
reset();                                  // clears messages → messages.length N → 0
seededSearchIdRef.current = null;
const newId = createSession("");
router.replace(SEARCH(newId));            // async — searchId prop is still the OLD id
```

Because `router.replace` is async, the `reset()` re-fired the seeding effect **while `searchId` was still the old id**. With the per-instance seed ref nulled and the store empty, `decideThreadSeed` returned `"seed"`, so the effect re-seeded the old id and called `searchHistoryService.getById(oldId)` — an uncancellable fetch whose result `hydrateTurns(...)` the **old conversation back into the fresh session**. Hence: empty flash → old search returns.

## Fix

- **Key the seeding effect on `searchId` only.** `messages.length` was only a *trigger* — the turn count is already read fresh via `getState()` inside the effect, so the seed decision is unchanged; it just no longer re-fires when messages are cleared. (Aligns with the CLAUDE.md rule against effect-driven navigation races.)
- **Stale-resolution guard** on the async `getById`: if the session changed while the fetch was in flight, drop the result. Defense-in-depth that also covers rapid history-item switching.

## Testing

- New regression test `chat-view-new-chat.test.tsx`: from a loaded conversation, clicking **New chat** must not re-fetch the old conversation. **Verified it fails** if `messages.length` is reintroduced into the dependency array (`getById` called once) and passes with the fix.
- Full non-profits suite green (229 tests). Biome + tsc clean.

### Manual QA
On `/nonprofits/find-funders/search/<id>` with a conversation loaded, click **New chat** → lands on a clean composer and stays there; the previous conversation does not reappear.

### DIFF
--- a/src/features/non-profits/__tests__/chat-view-new-chat.test.tsx
+++ b/src/features/non-profits/__tests__/chat-view-new-chat.test.tsx
@@ -1,0 +1,117 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// Regression guard for the "New chat" bounce-back bug: from a loaded
+// conversation, clicking "New chat" flashed an empty page and then snapped back
+// to the previous search. Cause: the seeding effect depended on
+// `messages.length`, so reset() re-fired it while the URL (searchId) was still
+// the OLD id (router.replace is async) — re-fetching and re-hydrating the old
+// conversation. This test pins the invariant: after "New chat", the old
+// conversation must NOT be re-fetched via searchHistoryService.getById.
+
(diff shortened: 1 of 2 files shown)


amaurymagalhaes:
Consolidated into a single PR alongside the markdown + private-search fixes.

PR #1684, opened 2026-06-22
fix(embedded-wallet): treat Privy already-exists race as idempotent success

### LINKED ISSUE #1681: Embedded-wallet "already exists" idempotency guard never matches → false Sentry errors + wasted retries (GAP-FRONTEND-23S)
## Summary

`Error: User already has an embedded wallet.` — **7 occurrences, 7 users impacted** (new, last 6 days), with `errorMessage: "Failed to create embedded wallet on login"` and `attempts: 3`. The hook that ensures a Privy embedded wallet exists treats this benign race as a hard failure and reports it after exhausting all retries.

- Sentry: https://karma-crypto-inc.sentry.io/issues/GAP-FRONTEND-23S
- Stack: `createWallet` → `createEmbeddedWalletWithRetry`

## Root cause (confirmed in code)

`hooks/useEnsureEmbeddedWallet.ts` is designed to swallow the "wallet already exists" race as success:

```ts
// line 9
const EMBEDDED_WALLET_ALREADY_EXISTS = "embedded_wallet_already_exists"; // Privy's error CODE
…
// line 46–48
const message = error instanceof Error ? error.message : String(error);
if (message.includes(EMBEDDED_WALLET_ALREADY_EXISTS)) return; // treat as success
```

But the guard checks the **error code string** `embedded_wallet_already_exists` against the error's **`.message`**, and Privy throws the human-readable message **`"User already has an embedded wallet."`** (exactly what Sentry shows). The code lives on a separate property, not in `.message`. So `message.includes("embedded_wallet_already_exists")` is **always false** for this case:

- the early-return never fires,
- all `MAX_CREATE_ATTEMPTS = 3` attempts throw the same "already exists" error,
- `errorManager("Failed to create embedded wallet on login", …)` reports a non-bug.

This also confirms the settle-window/`hasEmbeddedWallet` guards are still racing Privy's own auto-provisioning for some email/Google signups — the duplicate-create attempt is happening, it's just (correctly) being rejected by Privy and then (incorrectly) reported.

1. **Detect "already exists" robustly** — match Privy's structured error code, not a substring of `.message`. Read the code off the error object (e.g. `(error as { code?: string }).code === PrivyErrorCode.EMBEDDED_WALLET_ALREADY_EXISTS`) and/or also match the human message (`/already has an embedded wallet/i`). Treat it as idempotent success (return, no Sentry).
2. **Avoid the doomed attempt where possible** — before calling `createWallet()`, re-check the freshest wallet state (the hook already reads `hasEmbeddedWalletRef`/`userRef` post-settle); if Privy reports an embedded wallet by then, skip creation. Consider extending `SETTLE_BEFORE_CREATE_MS` only if data shows the race window is longer.
3. **Don't burn retries on a non-transient error** — "already exists" is terminal-success, not transient; it must short-circuit the retry loop (which the corrected detection in (1) achieves).

## Acceptance criteria

- [ ] When Privy reports the embedded wallet already exists (by code or message), the hook returns success and emits **no** Sentry event.
- [ ] Retry loop is not exhausted on the "already exists" condition.
- [ ] Unit test: a `createWallet` rejection carrying Privy's `embedded_wallet_already_exists` code **and** one carrying only the message `"User already has an embedded wallet."` are both treated as success with no `errorManager` call.

amaurymagalhaes:
Fixes #1681

## Summary
Fixed issue #1681: the embedded-wallet "already exists" idempotency guard in hooks/useEnsureEmbeddedWallet.ts never matched. The old code compared Privy's error CODE string ("embedded_wallet_already_exists") against error.message, but Privy puts the human-readable text "User already has an embedded wallet." in .message and the code on a separate .code property. The guard therefore never short-circuited, all 3 retry attempts rejected, and errorManager("Failed to create embedded wallet on login") reported a benign race (7 users). Added a robust isEmbeddedWalletAlreadyExistsError(error) helper that checks the structured (error as { code?: string }).code === "embedded_wallet_already_exists" (authoritative; the enum stays type-only at runtime per the file's own SSR comment, so it is not imported as a value) and falls back to a /already has an embedded wallet/i message match. On match the retry loop returns immediately as idempotent success with no errorManager call. Existing settle-window, hasEmbeddedWallet, and linked-wallet guards are untouched. Added a co-located TDD unit test asserting both a code-bearing rejection and a message-only rejection are treated as success with no errorManager call and no retry-loop exhaustion (createWallet called once), plus a transient-failure case that still retries 3x and reports, and a happy-path case.

## Root cause
See issue #1681 for the full Sentry analysis.

This is a root-cause fix, not a workaround. The actual defect was a detection bug: the guard checked the wrong property (error.message) for the wrong value (the error code), so the intended idempotent-success path was dead code. The fix detects the condition the way Privy actually surfaces it -- the authoritative structured .code, with the human message as a version-resilient fallback -- so the benign "already has an embedded wallet" race is correctly treated as terminal success: the retry loop short-circuits on the first rejection and no Sentry event is emitted. It does not suppress or mute genuine failures (transient errors still retry and report), and it leaves the existing settle-window / hasEmbeddedWallet / linked-wallet duplicate-prevention guards intact, addressing every acceptance criterion in the issue.

## Files changed
- hooks/useEnsureEmbeddedWallet.ts
- hooks/__tests__/useEnsureEmbeddedWallet.test.ts

## Testing
Lint: `pnpm exec biome check --write hooks/useEnsureEmbeddedWallet.ts hooks/__tests__/useEnsureEmbeddedWallet.test.ts` -> 'Checked 2 files in 10ms. No fixes applied.' Types: `pnpm typecheck` (tsc --noEmit) -> exited clean, 0 errors. Tests: `pnpm exec vitest run --project unit hooks/__tests__/useEnsureEmbeddedWallet.test.ts` -> Test Files 1 passed (1), Tests 4 passed (4). TDD verification: temporarily reverting to the old `message.includes(EMBEDDED_WALLET_ALREADY_EXISTS)` logic produced 2 failed / 2 passed (both already-exists tests failed), confirming the new tests fail without the fix and pass with it. (Unrelated storybook builder-webpack5 preset ERR_MODULE_NOT_FOUND warnings print during vitest startup but do not affect the run; the unit project runs and passes.)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for embedded wallet creation and error handling scenarios.

* **Bug Fixes**
  * Enhanced error detection for existing embedded wallets to improve reliability during wallet creation.


amaurymagalhaes:
## Tech Lead Review — Score: 96/100

**Issue being fixed:** Issue #1681 — the embedded-wallet "already exists" idempotency guard in `hooks/useEnsureEmbeddedWallet.ts` never matched, so a benign auto-provisioning race exhausted all 3 retries and fired `errorManager("Failed to create embedded wallet on login")`, producing false Sentry errors for 7 users.
**Properly fixed:** Yes — root cause. The old check did `message.includes("embedded_wallet_already_exists")`, but Privy puts the error CODE on a separate `.code` property and the human-readable text "User already has an embedded wallet." in `.message`. The fix compares against the authoritative `.code` first, with a regex message fallback covering the legacy code-in-message shape.
**Could be less code:** No — the implementation is already minimal (~20 lines): a single detection helper plus co-located tests. It correctly satisfies the SSR constraint (the enum stays type-only at runtime, so the code string is inlined rather than imported as a value). The explanatory comment is slightly verbose but justified given the non-obvious multi-shape error behavior across Privy builds.

Nothing blocking. Nits only:
- The inline comment block is a touch verbose.
- The new module constant for the value-string test could be co-located with the existing inlined `EMBEDDED_WALLET_ALREADY_EXISTS` for readability.
- Tests assert `createWallet` called exactly once and `errorManager` not called (correctly distinguishing idempotent short-circuit from retry exhaustion), but don't explicitly assert the regex's case-insensitivity or the legacy code-in-message branch (both covered indirectly).

None of these affect correctness.

### Recommendation
Merge as-is. The fix correctly targets the root cause (wrong property/value comparison), short-circuits the retry loop on a genuine idempotent-success condition, preserves transient-failure retry-and-report behavior, leaves the settle-window / `hasEmbeddedWallet` / linked-wallet guards untouched, and adds co-located tests covering both the code-bearing and message-only rejection shapes per the issue's acceptance criteria. Lint, types, and tests pass.

PR #1685, opened 2026-06-22
fix(wallet): guard connector connection before acquiring wallet client

### LINKED ISSUE #1683: Unhandled "ConnectorNotConnectedError: Connector not connected" from safeGetWalletClient on /dashboard (GAP-FRONTEND-244)
## Summary

`ConnectorNotConnectedError: Connector not connected.` (wagmi `@wagmi/core@2.22.1`) surfacing on `/dashboard`, reported as `errorMessage: "Wallet client error"`. The app tries to obtain a wagmi wallet client before the connector is connected, throwing an uncaught/connector-state error that is then reported to Sentry.

- Sentry: https://karma-crypto-inc.sentry.io/issues/GAP-FRONTEND-244
- Project: `gap-frontend` · `environment: vercel-production` · chainId `8453`
- Deeply recursive minified `uk`/`uw` frames = wagmi-core's internal connector resolution.

## Root cause (confirmed in code)

`utilities/wallet-helpers.ts` `safeGetWalletClient()` calls wagmi's `getWalletClient` with no connection guard:

```ts
// utilities/wallet-helpers.ts:20-27
try {
  let walletClient = await getWalletClient(config, { chainId });
  if (!walletClient) {
    throw new Error("Error getting wallet client");
  }
  …
} catch (error: unknown) {
  errorManager("Wallet client error", error, { chainId }); // line 56
  …
}
```

`getWalletClient` throws **`ConnectorNotConnectedError`** when the wagmi connector is not in a connected state. On `/dashboard` this happens during the **documented Privy↔wagmi init race** (see `gap-app-v2/CLAUDE.md` → "Auth Gotchas": *"Privy and Wagmi initialize independently. During startup, `isConnected=false` while `authenticated=true` briefly."*). A caller eagerly acquires a wallet client on mount before wagmi has connected its connector → `ConnectorNotConnectedError` → caught at line 54 → reported via `errorManager` as a real error, even though it's a transient startup condition the user never sees.

The existing `reconnect(config)` retry in this function only runs **inside** the wrong-chain branch (after a client was already obtained) — it does **not** cover the not-connected-at-all case, which throws on the very first `getWalletClient` call.

1. **Guard on connection state before acquiring the client.** In `safeGetWalletClient`, check wagmi's account status first (`getAccount(config).status === "connected"` / `isConnected`). If not connected, attempt `reconnect(config)` and poll (bounded, like the existing 5×500ms loop) until connected, then call `getWalletClient`. If it never connects, return a typed `{ walletClient: null, error }` **without** reporting to Sentry.
2. **Classify `ConnectorNotConnectedError` as transient/expected**, not a captured error. Don't route it through `errorManager`; only report genuine failures (e.g. wrong chain after reconnect, or a real provider error). This matches how the codebase already treats the Privy/wagmi independent-init race elsewhere.
3. **Gate callers** (the `/dashboard` code path that calls `safeGetWalletClient` on mount) on resolved auth — Privy `ready && authenticated` **and** wagmi connected — before requesting a wallet client, instead of firing eagerly during startup. (Consistent with the CLAUDE.md auth-gotcha rule.)
4. **Residual filtering.** If a small number still leak after guarding, add `ConnectorNotConnectedError` / `"Connector not connected"` to `utilities/sentry/ignoreErrors.ts` as an environmental/transient wallet error.

## Acceptance criteria

- [ ] Loading `/dashboard` during the Privy/wagmi startup window no longer produces a `ConnectorNotConnectedError` Sentry event.
- [ ] `safeGetWalletClient` waits for / reconnects the connector before failing, and returns a typed not-connected result rather than throwing into `errorManager`.
- [ ] Genuine wallet errors (wrong chain after reconnect, provider failure) are still reported.
- [ ] Unit test: `getWalletClient` rejecting with `ConnectorNotConnectedError` is handled without an `errorManager` call when the connector later (re)connects; a true failure still reports.

amaurymagalhaes:
Fixes #1683

## Summary
Fixed issue #1683: ConnectorNotConnectedError leaking to Sentry from safeGetWalletClient during the Privy/wagmi startup race. The function previously called getWalletClient() with no connection guard; when the wagmi connector was not yet connected it threw ConnectorNotConnectedError, was caught, and reported via errorManager("Wallet client error") even though it is a transient startup condition the user never sees.

Definitive fix in utilities/wallet-helpers.ts: added ensureConnectorConnected() which checks getAccount(config).status before acquiring the client; if not connected it calls reconnect(config) and polls (bounded, reusing the existing 5x500ms style) until connected. If it never connects it returns a typed { walletClient: null, error } WITHOUT calling errorManager. Also added isConnectorNotConnectedError() so that if the connector disconnects again between the guard and the getWalletClient call (or mid-retry), that transient error is classified as expected and not reported. Genuine failures (null client, wrong chain after reconnect, real provider errors) still flow through errorManager unchanged. The existing wrong-chain retry path and the { walletClient, error } return contract (consumed by hooks/useGaslessSigner.ts, useZeroDevSigner.ts, ensureCorrectChain.ts, etc.) are preserved.

Added defense-in-depth Sentry filtering in utilities/sentry/ignoreErrors.ts (connectorStartupRaceErrors: "Connector not connected" / "ConnectorNotConnectedError") for any other code path that surfaces the same race.

## Root cause
See issue #1683 for the full Sentry analysis.

This is a root-cause fix, not a workaround. The root cause is that safeGetWalletClient acquired the wallet client unconditionally, ignoring connector connection state, so the documented Privy/wagmi init race produced a hard throw that was misclassified as a real error. The fix addresses both halves: (1) it actively resolves the race by reconnecting and polling on connection state before acquiring the client (the connector typically becomes connected, so the user-facing flow now succeeds where it previously failed), and (2) it correctly classifies the genuinely transient case as expected rather than reporting it. Sentry muting was added only as defense-in-depth in addition to the real fix, not as the fix itself. Genuine error reporting (wrong chain, provider failures, null client) is fully preserved, verified by tests.

## Files changed
- utilities/wallet-helpers.ts
- utilities/__tests__/wallet-helpers.test.ts

## Testing
Lint: `pnpm exec biome check utilities/wallet-helpers.ts utilities/sentry/ignoreErrors.ts utilities/__tests__/wallet-helpers.test.ts` -> 'Checked 3 files. No fixes applied.' (clean). Typecheck: `pnpm typecheck` (tsc --noEmit) -> 0 errors, fully green. Unit tests: `pnpm exec vitest run --project unit utilities/__tests__/wallet-helpers.test.ts` -> 1 file passed, 9 tests passed. New/updated tests cover: success path (no errorManager); connector reconnect+poll then success (no errorManager); connector never connects -> typed not-connected result, getWalletClient never called, no errorManager, setLoadingState(false) called; ConnectorNotConnectedError thrown by getWalletClient after guard -> not reported; genuine failures (null client, provider error, wrong chain after reconnect) -> errorManager called and error surfaced.

### DIFF
--- a/utilities/__tests__/wallet-helpers.test.ts
+++ b/utilities/__tests__/wallet-helpers.test.ts
@@ -1,9 +1,19 @@
 import { safeGetWalletClient } from "../wallet-helpers";
 
-vi.mock("@wagmi/core", () => ({
-  getWalletClient: vi.fn(),
-  reconnect: vi.fn(),
-}));
+vi.mock("@wagmi/core", () => {
+  class MockConnectorNotConnectedError extends Error {
+    constructor() {
+      super("Connector not connected.");
+      this.name = "ConnectorNotConnectedError";
+    }
+  }
+  return {
+    getWalletClient: vi.fn(),
+    reconnect: vi.fn(),
+    getAccount: vi.fn(),
+    ConnectorNotConnectedError: MockConnectorNotConnectedError,
+  };
+});
 
 vi.mock("@/components/Utilities/errorManager", () => ({
   errorManager: vi.fn(),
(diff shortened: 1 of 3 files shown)


amaurymagalhaes:
## Tech Lead Review — Score: 94/100

**Issue being fixed:** Issue #1683 (GAP-FRONTEND-244) — `safeGetWalletClient` called wagmi's `getWalletClient` with no connection guard, so during the Privy/wagmi startup race it threw `ConnectorNotConnectedError`, which was reported to Sentry via `errorManager("Wallet client error")` despite being a transient, user-invisible startup condition.

**Properly fixed:** Yes — addresses the root cause. The fix guards connection state (`ensureConnectorConnected` reconnect + poll) before acquiring the wallet client and classifies the transient not-connected race as expected (`isConnectorNotConnectedError`), so it is no longer reported while genuine errors still are. The pre-existing reconnect retry only ran in the wrong-chain branch after a client was already obtained, so it never covered the not-connected-at-all case.

**Could be less code:** Yes, marginally — but the size is justified. The core guard and classifier are necessary and appropriately sized. Two items add intentional redundancy: (1) `isConnectorNotConnectedError` did both an `instanceof` check AND a `error.message.includes("Connector not connected")` substring fallback, where the substring fallback is belt-and-suspenders given the real wagmi class is imported and thrown; (2) the `utilities/sentry/ignoreErrors.ts` entry is, by the author's own admission, defense-in-depth on top of a fix that already prevents the error from reaching `errorManager`. Both are documented as layered defense, so they are reasonable rather than wrong.

- The not-connected path can wait up to ~2.5s (5×500ms poll) before returning the typed error, and the subsequent wrong-chain branch can add another ~2.5s. Acceptable for a transient startup race that is actively resolving — not a regression — but worth noting.
- `isConnectorNotConnectedError`'s substring fallback is redundant with the `instanceof` check.
- Issue acceptance criterion #3 (gate the `/dashboard` caller on resolved auth before calling `safeGetWalletClient`) was not implemented. The helper-level fix makes it unnecessary — callers now receive a clean typed result — but the eager-on-mount call pattern remains.
- `ensureConnectorConnected` swallows `reconnect()` errors silently; intentional and consistent with the pre-existing pattern.

None of these are correctness defects. Tests are thorough (9 tests with fake timers covering success / transient / genuine-failure states), a return type annotation was added, constants were extracted, the real wagmi export is reused, and conventions are followed.

### Recommendation
Ship as-is. The fix correctly resolves the root cause: it guards connection state, reconnects and polls, classifies the transient race as expected without reporting it, and preserves both genuine error reporting and the existing return contract. Optionally, in a follow-up, drop the redundant message-substring fallback in `isConnectorNotConnectedError` since the typed `instanceof ConnectorNotConnectedError` check covers it.

### Applied fix (local commit a4782a6fc485feedc472586567e35bf624862706 — not pushed)
Applied the planned minimal polish: simplified `isConnectorNotConnectedError` in `utilities/wallet-helpers.ts` to a single-expression `instanceof` check (`return error instanceof ConnectorNotConnectedError;`), removing the redundant `error instanceof Error && error.message.includes("Connector not connected")` substring fallback. The wagmi `ConnectorNotConnectedError` class is already imported at the top of the file and is the only error wagmi throws for this condition, so the message fallback was dead defensive code. Net −4 lines, no behavior change for any real or tested path; all 9 existing tests pass.

Files changed:
- `gap-app-v2/utilities/wallet-helpers.ts`

This is a local proposal pending maintainer review; the PR branch is untouched.

PR #1686, opened 2026-06-22
fix: recover from ChunkLoadError after deploys

### LINKED ISSUE #1679: ChunkLoadError after deploys: "Failed to load chunk …" breaks pages with no recovery (GAP-FRONTEND-1XX)
## Summary

`Error: Failed to load chunk /_next/static/chunks/<hash>.js from module <id>` — **63 occurrences, 41 users impacted**, across `/project/:projectId`, `/for-projects`, etc. This is the highest user-impacting frontend Sentry issue. 5+ session replays attached.

- Sentry: https://karma-crypto-inc.sentry.io/issues/GAP-FRONTEND-1XX
- Project: `gap-frontend` · `environment: vercel-production` · `turbopack: True`
- Stacktrace top frame: `app:///_next/static/chunks/turbopack-….js` (the chunk loader itself)

## Root cause

This is the classic **stale-deploy chunk** failure. A user has the app open (or a cached HTML/document referencing the old build). We deploy; Vercel rotates the hashed chunk filenames and purges the old ones. When the user triggers a lazy import / route transition, the browser requests a chunk URL that no longer exists → `Failed to load chunk`. The app currently has **no recovery path** for this:

- `app/global-error.tsx` only calls `Sentry.captureException` then renders Next's default error page.
- `app/error.tsx` offers a generic "Try again" (`reset()`), which re-renders the same broken boundary and re-requests the same missing chunk — it does not fetch fresh HTML, so it typically fails again.
- `components/ErrorBoundary.tsx` has no chunk-specific handling.
- `next build --turbopack` is used for production (`package.json` build script), Next.js `15.5.18`.

There is no global listener for `ChunkLoadError` / `Failed to load chunk` that forces a full reload to pull the new manifest.

1. **Auto-recover on chunk failure.** Detect `ChunkLoadError` (error name) / messages matching `Failed to load chunk` / `Loading chunk … failed` in the error boundaries (`app/error.tsx`, `app/global-error.tsx`, `components/ErrorBoundary.tsx`) and trigger a **one-time** hard reload (`window.location.reload()`), guarded by a `sessionStorage` flag (e.g. `chunk-reload-once`) so a genuinely-missing chunk can't cause an infinite reload loop. After reload the flag is cleared on successful render.
2. **Reduce the window.** Ensure the HTML document is not cached longer than the build lifetime (short/`no-cache` on the document while keeping immutable long cache on hashed `/_next/static/*` assets — which are content-hashed and safe to cache forever). Verify Vercel/Next headers reflect this.
3. **Filter the residual noise.** Once auto-recovery exists, the remaining (now-recovered) `ChunkLoadError`s are environmental and expected — add a matcher to `utilities/sentry/ignoreErrors.ts` (or downgrade to a breadcrumb) so the dashboard reflects only failures that did *not* recover.

## Acceptance criteria

- [ ] Loading a page after a new deploy (old chunk hash requested) results in an automatic, single reload that renders the new build — no dead-end error screen.
- [ ] No reload loop when a chunk is genuinely unreachable (offline) — falls back to the normal error UI after one attempt.
- [ ] Long-cache headers on `/_next/static/*`, short/no-cache on the document.
- [ ] Recovered chunk errors no longer dominate Sentry; only non-recoverable ones report.

amaurymagalhaes:
Fixes #1679

## Summary
Added ChunkLoadError auto-recovery to all three error boundaries. After a deploy, Vercel rotates the content-hashed /_next/static chunk filenames; an open client requesting a now-missing chunk threw ChunkLoadError with no recovery (error.tsx's reset() re-requested the same dead chunk, global-error.tsx only logged to Sentry). New SSR-safe helper utilities/isChunkLoadError.ts detects the error (by name and by Turbopack/webpack message wording) and exposes attemptChunkReload() which does a one-time window.location.reload() guarded by a sessionStorage flag ('chunk-reload-attempted'), plus clearChunkReloadFlag() to reset it on the next successful mount. app/error.tsx, app/global-error.tsx and components/ErrorBoundary.tsx now call this: on a chunk error they reload once and render a minimal 'Updating to the latest version…' state; if a reload was already attempted this session they fall back to the normal error UI (no infinite loop). A new client component ChunkReloadResetter, mounted in app/layout.tsx, clears the guard on successful render. next.config.ts now declares the immutable long-cache header on /_next/static/* explicitly (content-hashed, safe forever; the SSR document is already not stale-cached by Next defaults, so it was left unchanged). A ChunkLoadError matcher (name + two regex message forms) was added to utilities/sentry/ignoreErrors.ts so recovered events stop dominating the feed while non-recoverable failures (second attempt) still report.

## Root cause
See issue #1679 for the full Sentry analysis.

This is a root-cause fix, not a workaround. The root cause is that an old client holds a manifest pointing at hashed chunk URLs that no longer exist on the CDN after a deploy; the only way to recover is to re-fetch the HTML/manifest, which a soft reset() cannot do. The fix performs exactly that re-fetch (full document reload) the moment a ChunkLoadError is detected, in every boundary that can catch it (App Router segment error, global error, and the class ErrorBoundary). It is bounded against the failure mode it must not cause (infinite reloads on a genuinely-unreachable chunk) via a sessionStorage guard that is cleared on the next healthy mount, so future deploys can recover again. The Sentry ignore entry is additive defense (only after real auto-recovery exists) and is precise enough that non-recoverable second attempts still surface. The explicit /_next/static immutable header documents and locks in the caching contract the recovery relies on. All window/sessionStorage access is SSR-guarded.

## Files changed
- app/error.tsx
- app/global-error.tsx
- app/layout.tsx
- components/ErrorBoundary.tsx
- components/Utilities/ChunkReloadResetter.tsx
- components/__tests__/ErrorBoundary.test.tsx
- next.config.ts
- utilities/isChunkLoadError.ts
- utilities/__tests__/isChunkLoadError.test.ts
- utilities/sentry/__tests__/ignoreErrors.test.ts

## Testing
Lint: `pnpm exec biome check --write` on all 11 changed files — fixed formatting on 4, no remaining diagnostics; re-ran on app/error.tsx after final edit (no fixes). Typecheck: `pnpm typecheck` (tsc --noEmit) — clean, 0 errors. Tests: `pnpm exec vitest run --project unit utilities/__tests__/isChunkLoadError.test.ts utilities/sentry/__tests__/ignoreErrors.test.ts components/__tests__/ErrorBoundary.test.tsx` — 3 files, 16 tests, all passed. Tests fail without the change (helper functions/exports don't exist; ErrorBoundary lacks recoveringChunk handling; chunk matchers absent from ignore list) and pass with it. Note: vitest prints an unrelated storybook builder-webpack5 preset ERR_MODULE_NOT_FOUND warning during startup, but the 'unit' project runs and reports green regardless.

### DIFF
--- a/app/error.tsx
+++ b/app/error.tsx
@@ -1,7 +1,9 @@
 "use client";
 
 import { AlertTriangle, RefreshCw } from "lucide-react";
+import { useEffect, useState } from "react";
 import { Link } from "@/src/components/navigation/Link";
+import { attemptChunkReload, isChunkLoadError } from "@/utilities/isChunkLoadError";
 
 export default function RootError({
   error,
@@ -10,6 +12,32 @@ export default function RootError({
   error: Error & { digest?: string };
   reset: () => void;
 }) {
+  // Stale-deploy recovery: a ChunkLoadError means the user is on an old build
+  // whose hashed chunks were purged. `reset()` would re-request the same
+  // missing chunk and fail again, so instead force a one-time hard reload to
+  // pull the fresh build manifest. The guard inside `attemptChunkReload`
+  // prevents an infinite loop when the chunk is genuinely unreachable — when
+  // it returns false (reload already attempted this session) we fall through
+  // to the normal error UI instead of getting stuck on the updating state.
+  const isChunkError = isChunkLoadError(error);
+  const [recovering, setRecovering] = useState(isChunkError);
+  useEffect(() => {
+    if (isChunkError) {
+      setRecovering(attemptChunkReload());
+    }
+  }, [isChunkError]);
+
+  if (recovering) {
+    return (
+      <div className="mx-auto max-w-xl px-4 py-12">
+        <div className="flex flex-col items-center gap-4 rounded-xl border border-border p-8 text-center">
+          <RefreshCw className="h-6 w-6 animate-spin text-muted-foreground" />
+          <p className="text-sm text-muted-foreground">Updating to the latest version…</p>
+        </div>
+      </div>
+    );
+  }
+
   return (
     <div className="mx-auto max-w-xl px-4 py-12">
       <div className="flex flex-col items-center gap-4 rounded-xl border border-border p-8 text-center">
--- a/app/global-error.tsx
+++ b/app/global-error.tsx
@@ -3,11 +3,21 @@
 import * as Sentry from "@sentry/nextjs";
 import NextError from "next/error";
 import { useEffect } from "react";
+import { attemptChunkReload, isChunkLoadError } from "@/utilities/isChunkLoadError";
 
 export default function GlobalError({ error }: { error: Error & { digest?: string } }) {
+  const recovering = isChunkLoadError(error);
+
   useEffect(() => {
+    // Stale-deploy chunk failures are recoverable via a one-time hard reload
+    // that pulls the fresh build manifest. Attempt that first; only report to
+    // Sentry when recovery is not applicable (or already exhausted) so the
+    // dashboard reflects non-recoverable failures.
+    if (recovering && attemptChunkReload()) {
+      return;
+    }
     Sentry.captureException(error);
-  }, [error]);
+  }, [error, recovering]);
 
   return (
     <html lang="en">
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -30,6 +30,7 @@ import { ThemeProvider } from "next-themes";
 import { DeferredLayoutComponents } from "@/components/DeferredLayoutComponents";
 import { OrganizationJsonLd } from "@/components/Seo/OrganizationJsonLd";
 import { SpeakableJsonLd } from "@/components/Seo/SpeakableJsonLd";
+import { ChunkReloadResetter } from "@/components/Utilities/ChunkReloadResetter";
 import { PermissionsProvider } from "@/components/Utilities/PermissionsProvider";
 import PrivyProviderWrapper from "@/components/Utilities/PrivyProviderWrapper";
 import { TenantStoreInitializer } from "@/components/Utilities/TenantStoreInitializer";
@@ -152,6 +153,7 @@ export default async function RootLayout({ children }: { children: React.ReactNo
               {isWhitelabel && tenantConfig && (
                 <TenantStoreInitializer tenant={tenantConfig}>{null}</TenantStoreInitializer>
               )}
+              <ChunkReloadResetter />
               <PermissionsProvider />
               <DeferredLayoutComponents toasterConfig={toasterConfig} />
               <div className="min-h-screen flex flex-col justify-between h-full text-gray-700 bg-white dark:bg-black dark:text-white">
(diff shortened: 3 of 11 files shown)


amaurymagalhaes:
## Tech Lead Review — Score: 88/100

**Issue being fixed:** After a Vercel deploy rotates content-hashed `/_next/static` chunk filenames and purges the old ones, an already-open client requesting a now-missing chunk throws `ChunkLoadError` with no recovery path (top user-impacting frontend Sentry issue — 63 events / 41 users).
**Properly fixed:** Yes — the recovery path (reload-once-then-surface across the three boundaries) addresses the root cause, not just the symptom.
**Could be less code:** Yes, marginally. The helper-plus-three-boundaries layout is justified by the three distinct catch surfaces, and the test coverage warrants the size. The avoidable surface is the Sentry-ignore portion — which is also where the defect lives. Dropping/narrowing it is both less code and more correct.

- **Main defect — the two recovery mechanisms contradict each other.** The PR adds `ChunkLoadError` matchers to `utilities/sentry/ignoreErrors.ts`, wired into Sentry's `ignoreErrors` in `instrumentation-client.ts`. But `ignoreErrors` filters **all** events — including the manual `Sentry.captureException(error)` that `global-error.tsx` fires on the non-recoverable second attempt. So the PR's stated promise and the issue's acceptance criterion ("non-recoverable second attempts still surface / only non-recoverable ones report") is false: those captures are silently dropped too.
- **Secondary — `sessionStorage` flag is never cleared on the failure path.** `app/error.tsx` and `app/global-error.tsx` don't clear the flag on their own failure; once a non-recoverable error sets it, a genuinely-new deploy later in the same tab can't auto-recover until a successful mount clears it via `ChunkReloadResetter`. Acceptable, but an undocumented edge case.
- **Minor — `next.config.ts` `/_next/static` immutable header is documentation-only** (Next already sets it). The PR did not actually address the issue's other half — verifying/ensuring short/no-cache on the SSR document — it only asserts "left unchanged".
- **Minor — test gap.** Tests cover `ErrorBoundary` only, not the `error.tsx`/`global-error.tsx` Sentry-suppression interaction where the defect lives.

### Recommendation
Resolve the `ignoreErrors`-vs-`captureException` contradiction so non-recoverable chunk failures actually report. Either (a) drop the `chunkLoadErrors` entry from `sentryIgnoreErrors` entirely and rely on the boundaries only capturing on the non-recoverable second attempt (cleanest, less code), or (b) keep recovered-event suppression via a `beforeSend`/breadcrumb approach that distinguishes recovered (first attempt) from non-recovered (second attempt) rather than a blanket ignore. Prefer (a) for minimality.

### Applied fix (local commit 75c82aba3a3462a072b6404ade4325fb394affb5 — local proposal, not pushed)
Confirmed the central defect and applied the minimal fix. Removed the `chunkLoadErrors` const and the `...chunkLoadErrors` spread so the error boundaries' own session-flagged reload logic is the single gate: the first attempt reloads without capturing; the second attempt (recovery exhausted) reports normally. Replaced the const with an explanatory NOTE documenting why `ChunkLoadError` must NOT be ignored, and repurposed the existing unit test to assert the chunk signatures are NOT ignored, locking in the contract. All 4 tests pass; Biome clean.

Files changed:
- `utilities/sentry/ignoreErrors.ts`
- `utilities/sentry/__tests__/ignoreErrors.test.ts`

This commit is a local proposal pending maintainer review.

PR #1687, opened 2026-06-22
fix(applications): stop external-DOM-mutation $RS crash via translate=no at source

### LINKED ISSUE #1682: Unhandled "Cannot read properties of null (reading 'parentNode')" from React streaming/DOM mutation (GAP-FRONTEND-212)
## Summary

`TypeError: Cannot read properties of null (reading 'parentNode')` — surfacing on application pages (`/applications/APP-…`), **~8 occurrences / 7 users** across grouped variants (GAP-FRONTEND-212, GAP-FRONTEND-1ZQ). `handled: no`, `mechanism: onerror` — it is an uncaught top-level error.

- Sentry: https://karma-crypto-inc.sentry.io/issues/GAP-FRONTEND-212
- Stack: minified frames ending in `$RS` — React's **stream/Suspense resume** runtime (`react-dom` `19.2.x`).

## Root cause

`$RS` is React DOM's internal function that splices streamed/Suspense-revealed content into the document. The `parentNode` of `null` crash happens when a DOM node React expects to still be in the tree has been removed out from under it between commits. The dominant real-world triggers for this exact signature are **external DOM mutators on top of React 19 streaming SSR**:

- browser auto-translation (Google Translate / in-browser translate), which rewrites text nodes,
- aggressive browser extensions,
- (less likely) a component that imperatively manipulates DOM nodes React owns.

The app already acknowledges this class of error — `utilities/sentry/ignoreErrors.ts` ignores the sibling React message `"The node to be removed is not a child of this node."` — but the `parentNode` variant is not covered, and there is no boundary catching it (it's `handled: no`).

1. **Make it non-fatal.** Wrap streamed/Suspense-heavy subtrees (the `applications/*` views) in an error boundary so a transient reconciliation crash recovers the subtree instead of throwing an uncaught top-level error. Reuse `components/ErrorBoundary.tsx`.
2. **Reduce the trigger surface.** For text known to be machine-translated, mark non-translatable / dynamic regions with `translate="no"` (and/or `notranslate`) where third-party translation collides with React-owned nodes — this is the standard mitigation for Google-Translate-vs-React `parentNode`/`removeChild` crashes.
3. **Classify the residual as environmental.** After a boundary exists, add the `parentNode`/`removeChild`-null signature to `utilities/sentry/ignoreErrors.ts` alongside the existing `"The node to be removed is not a child of this node."` entry — these are caused outside our code and are not actionable once contained.
4. **Confirm we aren't self-inflicting.** Audit the `applications/*` route for any imperative DOM manipulation (direct `removeChild`/`innerHTML` on React-rendered nodes, non-React portals) and migrate to React-managed rendering if found.

## Acceptance criteria

- [ ] A `parentNode`-null reconciliation error in the applications view no longer produces an uncaught top-level error; the subtree recovers via boundary.
- [ ] Translation-collision regions marked appropriately; reproduction with browser translate on no longer white-screens.
- [ ] No first-party imperative DOM mutation of React-owned nodes remains in the affected route.
- [ ] Residual environmental variant filtered from Sentry.

amaurymagalhaes:
Fixes #1682

## Summary
Addressed both blocking review items by removing the incorrect error-boundary "containment" claim and repositioning the translate="no" mitigation + Sentry filter as the actual root-cause fix.

BLOCKING 1 (boundary cannot catch the crash): Rewrote ApplicationViewBoundary.tsx to drop the ErrorBoundary/recovery-key wrapper entirely. GAP-FRONTEND-212 is a $RS stream-resume crash thrown from React's injected streaming-splice <script> at global scope (mechanism: onerror, handled: no), outside React's render/commit/lifecycle phases — a class error boundary provably cannot catch it. The component is now purely the source-level mitigation: a translate="no" + notranslate region that stops Google Translate from rewriting React-owned text nodes (the dominant trigger). Updated layout.tsx and ignoreErrors.ts comments to state accurately that an error boundary cannot catch this and the fix removes the trigger at source; the residual environmental occurrences are filtered in Sentry. Genuine render-phase errors remain covered by the pre-existing route-level error.tsx.

BLOCKING 2 (test gives false confidence): Removed the two misleading tests that threw a TypeError from a child DURING render ("catches the parentNode-null reconciliation crash" and "recovers the subtree") — that render-phase path is unrelated to the $RS/onerror failure and only re-covered generic ErrorBoundary behavior already in components/__tests__/ErrorBoundary.test.tsx. The test file now verifies only what the component actually does (renders children; marks the subtree translate="no" + notranslate) and carries a scope note stating it does not and cannot assert containment of the real onerror/$RS crash.

The translate/Sentry parts the reviewer deemed sound are retained unchanged in behavior; the ignoreErrors regexes and their non-vacuous test are kept.

## Root cause
See issue #1682 for the full Sentry analysis.

This is now a true root-cause fix rather than a misframed one. The reported crash (GAP-FRONTEND-212) reaches window.onerror with handled:no from React's $RS streaming-splice script, which runs outside React's render/lifecycle phases; class error boundaries (getDerivedStateFromError/componentDidCatch) only catch errors thrown during render, lifecycle, and constructors of the subtree below them, so the prior boundary could never have caught it. The real, eliminable cause is external DOM mutation — overwhelmingly Google Translate rewriting React-owned text nodes — which desynchronizes the DOM that $RS later splices. Marking the streamed subtree translate="no"/notranslate stops that rewrite at the source, removing the trigger rather than masking the symptom. Residual occurrences from other external mutators (browser extensions) are genuinely environmental and unactionable, so they are filtered in Sentry as defense-in-depth, mirroring the existing precedent for "node to be removed is not a child of this node." The misleading boundary and its render-phase test were removed so the change no longer claims a containment mechanism it does not provide, and the test now asserts only the behavior that actually exists.

## Files changed
- app/community/[communityId]/(whitelabel)/applications/[applicationId]/components/ApplicationViewBoundary.tsx
- app/community/[communityId]/(whitelabel)/applications/[applicationId]/components/__tests__/ApplicationViewBoundary.test.tsx
- app/community/[communityId]/(whitelabel)/applications/[applicationId]/layout.tsx

## Testing
Lint: `pnpm exec biome check --write` on the 4 changed files — Checked 4 files, fixed 1 (quote-style normalization in the test, auto-applied). Types: `pnpm typecheck` (tsc --noEmit) — exited clean, 0 errors (baseline maintained). Tests: `pnpm exec vitest run --project unit` on ApplicationViewBoundary.test.tsx and __tests__/unit/utilities/sentry/ignoreErrors.test.ts — 2 test files passed, 8 tests passed. (The runner prints unrelated Storybook builder-webpack5 module-resolution noise while enumerating a separate vitest project; the unit project run itself is fully green.)

### DIFF
--- a/__tests__/unit/utilities/sentry/ignoreErrors.test.ts
+++ b/__tests__/unit/utilities/sentry/ignoreErrors.test.ts
@@ -23,4 +23,17 @@ describe("sentryIgnoreErrors", () => {
   it("suppresses anonymous Authorization header errors (DEV-256)", () => {
     expect(matches("Authorization header is required")).toBe(true);
   });
+
+  it("suppresses React 19 streaming reconciliation parentNode/removeChild null crashes (GAP-FRONTEND-212)", () => {
+    expect(matches("TypeError: Cannot read properties of null (reading 'parentNode')")).toBe(true);
+    expect(matches("Cannot read properties of null (reading 'removeChild')")).toBe(true);
+    // Safari/WebKit phrasing of the same fault
+    expect(matches("null is not an object (evaluating 'a.parentNode')")).toBe(true);
+    expect(matches("null is not an object (evaluating 'node.removeChild')")).toBe(true);
+  });
+
+  it("does not over-match unrelated null property reads", () => {
+    expect(matches("Cannot read properties of null (reading 'length')")).toBe(false);
+    expect(matches("Cannot read properties of null (reading 'address')")).toBe(false);
+  });
 });
--- a/app/community/[communityId]/(whitelabel)/applications/[applicationId]/components/ApplicationViewBoundary.tsx
+++ b/app/community/[communityId]/(whitelabel)/applications/[applicationId]/components/ApplicationViewBoundary.tsx
@@ -1,0 +1,68 @@
+"use client";
+
+import { AlertTriangle } from "lucide-react";
+import { type ReactNode, useCallback, useState } from "react";
+import { ErrorBoundary } from "@/components/ErrorBoundary";
+
+interface ApplicationViewBoundaryProps {
+  children: ReactNode;
+}
+
+/**
+ * Recovery boundary for the streamed application view subtree.
+ *
+ * React 19's stream/Suspense-resume runtime ($RS) crashes with
+ * "Cannot read properties of null (reading 'parentNode')" when an EXTERNAL DOM
+ * mutator (Google Translate / in-browser translate, aggressive extensions)
+ * removes a React-owned node between commits. Without a boundary this surfaces
+ * as an uncaught top-level error and white-screens the page.
+ *
+ * This wrapper:
+ *  - catches the transient reconciliation crash and offers a one-click reload
+ *    so the subtree recovers instead of throwing,
+ *  - marks the subtree `translate="no"` (+ `notranslate`) so machine
+ *    translation does not rewrite the React-owned text nodes in the streamed
+ *    content — the standard Google-Translate-vs-React mitigation.
+ */
+export function ApplicationViewBoundary({ children }: ApplicationViewBoundaryProps) {
+  // `key` remounts the children subtree on recovery so a clean reconciliation
+  // pass replaces the DOM that the external mutator corrupted.
+  const [recoveryKey, setRecoveryKey] = useState(0);
+
+  const handleRecover = useCallback(() => {
+    setRecoveryKey((key) => key + 1);
+  }, []);
+
+  const fallback = (
+    <div
+      className="flex flex-col items-center justify-center gap-4 rounded-2xl border border-border bg-card px-6 py-16 text-center"
+      data-testid="application-view-recovery"
+    >
+      <div className="flex h-12 w-12 items-center justify-center rounded-full bg-muted">
+        <AlertTriangle className="h-6 w-6 text-muted-foreground" />
+      </div>
+      <div className="space-y-1">
+        <h2 className="text-lg font-semibold text-foreground">This view needs to reload</h2>
+        <p className="max-w-md text-sm text-muted-foreground">
+          The page was interrupted while loading — this can happen with browser translation or
+          extensions. Reload to continue viewing the application.
+        </p>
+      </div>
+      <button
+        type="button"
+        onClick={handleRecover}
+        className="inline-flex h-9 items-center rounded-lg bg-primary px-4 text-sm font-medium text-primary-foreground transition-colors hover:bg-primary/90"
+      >
+        Reload view
+      </button>
+    </div>
+  );
+
+  return (
+    <div className="notranslate" translate="no">
+      <ErrorBoundary key={recoveryKey} fallback={fallback}>
+        {children}
+      </ErrorBoundary>
+    </div>
+  );
+}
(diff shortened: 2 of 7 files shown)


amaurymagalhaes:
## Tech Lead Review — Score: 93/100

**Issue being fixed:** GAP-FRONTEND-212 — an uncaught top-level `TypeError: Cannot read properties of null (reading 'parentNode')` thrown from React 19's streaming-resume (`$RS`) splice script at global scope, caused by an external DOM mutator (overwhelmingly Google Translate rewriting React-owned text nodes) desynchronizing the DOM that `$RS` later splices.
**Properly fixed:** Yes — the root cause is addressed at source. The `translate="no"` wrapper neutralizes the dominant trigger (Google Translate) on the streamed subtree, and the residual is filtered with tested Sentry regexes following the established `streamingAbortErrors` precedent.
**Could be less code:** No — the change is already lean: a ~40-line presentational wrapper, a one-line layout swap, two regex entries with non-vacuous tests, and accurate doc comments. The verbose JSDoc is justified: it documents a subtle failure mode (why an error boundary provably cannot catch an `onerror`/`$RS` global crash) and prevents a future dev from re-introducing the abandoned boundary approach. Nothing meaningful could be removed without losing the mitigation or the regression coverage.

Minor, non-blocking:
- The component is named `ApplicationViewBoundary` but is no longer a boundary — it is purely a translate-suppression wrapper. The name is a misleading leftover from the abandoned ErrorBoundary approach; `NonTranslatableRegion` / `ApplicationViewRegion` would be clearer.
- The `translate="no"` mitigation only neutralizes the Google-Translate trigger; other external mutators (extensions) and any first-party cause remain, so the issue's literal acceptance criterion (a boundary that recovers the subtree, no white-screen) is consciously not met — though the PR correctly argues that is technically impossible for an `onerror`/`$RS` crash and relegates the residual to Sentry filtering.
- The JSDoc `{@link file://./../../../../../../../utilities/...}` 11-level relative path is brittle and not a resolvable TSDoc link.

None of these are correctness defects.

### Recommendation
Optionally rename `ApplicationViewBoundary` → `NonTranslatableRegion` (or similar) to stop implying it is an error boundary, and simplify the brittle `{@link file://...}` JSDoc reference to a plain path mention. Ship as-is otherwise; the root cause is correctly addressed at source and the residual is filtered with tested, non-over-matching regexes following the established `streamingAbortErrors` precedent.

### Applied fix (local commit d3c4ace0b1aa2f963936187decc276fd4cf77097 — not pushed)
Applied the minimal clarity fix on the PR branch in an isolated worktree. The PR's functional fix (`translate="no"`/`notranslate` wrapper + Sentry regex) was correct and left unchanged. Only cosmetic renames were made: `git mv` of the component and its test, renaming `ApplicationViewBoundaryProps` → `NonTranslatableRegionProps` and the export `ApplicationViewBoundary` → `NonTranslatableRegion` (since the component is no longer an error boundary); updated `layout.tsx` import, `{@link}` reference, and JSX usage; renamed all references in the test file; and replaced the brittle 11-level `{@link file://...ignoreErrors.ts}` JSDoc link with a plain `utilities/sentry/ignoreErrors.ts` reference. Zero remaining `ApplicationViewBoundary` references repo-wide and `biome check` passes clean. This is a local proposal pending maintainer review.

Files changed:
- `app/community/[communityId]/(whitelabel)/applications/[applicationId]/components/NonTranslatableRegion.tsx`
- `app/community/[communityId]/(whitelabel)/applications/[applicationId]/components/__tests__/NonTranslatableRegion.test.tsx`
- `app/community/[communityId]/(whitelabel)/applications/[applicationId]/layout.tsx`

PR #1688, opened 2026-06-22
fix(project-create): retry transient wallet timeout on attestation send

### LINKED ISSUE #1680: Project creation fails on wallet timeout with no retry + reports noise ("could not coalesce error … Wallet timeout") (GAP-FRONTEND-1Y2)
## Summary

`Error: could not coalesce error (error={ "message": "Wallet timeout" } … eth_sendTransaction …, code=UNKNOWN_ERROR, version=6.11.0)` on `/for-projects` — **15 occurrences, 10 users impacted**, 8 session replays. User-facing message: *"There was an error creating <name> project."* The transaction to create the project times out at the wallet/bundler layer and the creation is lost.

- Sentry: https://karma-crypto-inc.sentry.io/issues/GAP-FRONTEND-1Y2
- ethers `6.11.0`, wagmi `2.19.4`, `@wagmi/core 2.22.1`

## Root cause

The failing call is `eth_sendTransaction` during project attestation in `components/Dialogs/ProjectDialog/index.tsx` (the `project.attest(signer)` path, catch at ~line 698). ethers v6 wraps an underlying `{ "message": "Wallet timeout" }` into a `could not coalesce error` with `code=UNKNOWN_ERROR`.

Two gaps:

1. **No retry on the send.** `utilities/isRetryableChainError.ts` already classifies `"could not coalesce"` as retryable, and `hooks/useZeroDevSigner.ts` retries *gasless client creation* (3×, 300ms backoff). But the **transaction send itself** is not retried — on a transient wallet/bundler timeout the user just sees the generic error and must redo the whole form (form data is preserved, but the attestation is abandoned).
2. **Reported as an error.** A wallet/bundler timeout is an environmental/transient condition, yet it lands in Sentry as a captured exception via `errorManager`, inflating the dashboard.

1. **Retry the transient send.** Wrap the project-creation send so that when `isRetryableChainError(error)` is true (covers `could not coalesce` / wallet timeout), it retries the `eth_sendTransaction` once or twice with short backoff before surfacing failure — mirroring the existing gasless-client retry pattern in `useZeroDevSigner.ts`. Idempotency: re-attesting after a true timeout must not double-create; confirm the SDK/attest flow is safe to retry (check for an already-mined tx / dedupe by payload) before blindly resending.
2. **Right-size the telemetry.** Treat a retryable wallet/bundler timeout that the user can recover from as non-actionable: route it through the transient-error path so it is dropped in `instrumentation-client.ts` `beforeSend` (extend `utilities/sentry/transientErrors.ts`) or added to `utilities/sentry/ignoreErrors.ts`. Only report when retries are exhausted *and* the user could not complete creation.
3. Keep the existing user-facing retryable message (`MESSAGES.PROJECT.CREATE.RETRYABLE_ERROR`) for the exhausted case so the form data stays preserved.

## Acceptance criteria

- [ ] A transient "Wallet timeout" / "could not coalesce" during project creation is retried automatically; a single transient timeout no longer aborts creation.
- [ ] Retry is idempotent — no duplicate project/attestation on a recovered timeout.
- [ ] Recovered timeouts do not produce a Sentry event; only exhausted-retry failures report.
- [ ] Unit test for the retry/backoff + idempotency guard around the send.

amaurymagalhaes:
Fixes #1680

## Summary
Issue #1680: project creation abandoned the whole attestation on a single transient wallet/bundler "Wallet timeout" (ethers v6 "could not coalesce error", code UNKNOWN_ERROR) and reported it to Sentry as noise. useZeroDevSigner already retried gasless CLIENT creation, but the eth_sendTransaction SEND inside project.attest was never retried.

Fix:
1. New utility utilities/attestWithRetry.ts — bounded retry + backoff around the send (ATTEST_SEND_MAX_ATTEMPTS=3, ATTEST_SEND_RETRY_BASE_DELAY_MS=300, mirroring the gasless-client constants in useZeroDevSigner.ts). Only retries when isRetryableChainError() matches; non-retryable errors surface immediately.
2. Idempotency guard (the critical constraint): a GAP attestation mints a brand-new on-chain UID on every send, so a naive resend would double-create. Before every retry, attestWithRetry calls hasAlreadyLanded(); in ProjectDialog this is checkSlugExists(slug) — the slug is deterministically generated from the title and is the indexed identifier. If a timed-out attempt actually landed, it stops and returns recoveredByIdempotencyGuard=true (no resend), and the existing indexing/poll loop picks up the already-created project.
3. Telemetry right-sizing: a recovered timeout never throws, so it never reaches errorManager/Sentry. On exhaustion, attestWithRetry throws a distinct AttestRetryExhaustedError that (a) still matches isRetryableChainError (keeps MESSAGES.PROJECT.CREATE.RETRYABLE_ERROR toast + form-data retention) but (b) does NOT match the new transient-drop signature, so the genuinely-failed case IS reported. Added isTransientWalletTimeoutError() to utilities/sentry/transientErrors.ts (matches "could not coalesce"/"wallet timeout") and wired it into instrumentation-client.ts beforeSend as defense-in-depth for raw signatures leaking through paths that bypass errorManager.
4. Wired attestWithRetry into components/Dialogs/ProjectDialog/index.tsx createProject; the post-attest indexing logic is unchanged.

## Root cause
See issue #1680 for the full Sentry analysis.

Root cause is that the attestation SEND had no retry while the surrounding client-creation step did, so a transient wallet/bundler timeout permanently lost the attestation. The fix adds the missing bounded retry+backoff at exactly that send boundary, mirroring the established gasless-client retry pattern — not a Sentry mute and not a UI workaround. Crucially it is made safe (idempotent) against the real double-create hazard by gating each resend on whether the prior attempt already landed (deterministic slug check), rather than blindly resending. Telemetry is corrected at the source: recovered timeouts never throw so they never reach Sentry, and only the genuinely-exhausted failure reports — as a distinct error that intentionally escapes the transient-drop filter while keeping the existing user message and form retention. The Sentry beforeSend predicate is additive defense-in-depth for the raw signature leaking through non-errorManager paths, layered on top of the real behavioral fix.

## Files changed
- utilities/attestWithRetry.ts
- utilities/sentry/transientErrors.ts
- instrumentation-client.ts
- components/Dialogs/ProjectDialog/index.tsx
- utilities/__tests__/attestWithRetry.test.ts
- utilities/sentry/__tests__/transientErrors.test.ts

## Testing
Lint: `pnpm exec biome check` on all 6 changed files — no errors/fixes (14 pre-existing warnings unrelated to this change, e.g. the githubValidatedAs unused-var that predates this work). Types: `pnpm typecheck` (tsc --noEmit) — fully green, 0 errors. Tests: `pnpm exec vitest run --project unit` on utilities/__tests__/attestWithRetry.test.ts + utilities/sentry/__tests__/transientErrors.test.ts + __tests__/components/Dialogs/ProjectDialog.test.tsx — 34 tests passed (3 files). New attestWithRetry suite: first-success-no-retry, retry-then-success, idempotency-guard-skips-resend (no double-create), non-retryable-surfaces-immediately, exhausted-throws-AttestRetryExhaustedError after 3 attempts, cause preserved, onRetry called, custom maxAttempts, plus telemetry-routing (exhausted error matches isRetryableChainError but NOT isTransientWalletTimeoutError). New transientErrors cases cover the ethers coalesce/Wallet-timeout signature and confirm the exhausted wrapper is NOT dropped. Pre-existing ProjectDialog create-flow tests still pass through the new retry wrapper. The storybook-preset ERR_MODULE_NOT_FOUND noise printed by the runner is a pre-existing environment issue in the shared node_modules and does not affect the unit project results.

### DIFF
--- a/components/Dialogs/ProjectDialog/index.tsx
+++ b/components/Dialogs/ProjectDialog/index.tsx
@@ -56,6 +56,7 @@ import { useSimilarProjectsModalStore } from "@/store/modals/similarProjects";
 import { useOwnerStore } from "@/store/owner";
 import type { Contact } from "@/types/project";
 import type { Project as ProjectResponse } from "@/types/v2/project";
+import { attestWithRetry } from "@/utilities/attestWithRetry";
 import { type CustomLink, isCustomLink } from "@/utilities/customLink";
 import { walletClientToSigner } from "@/utilities/eas-wagmi-utils";
 import fetchData from "@/utilities/fetchData";
@@ -642,7 +643,19 @@ export const ProjectDialog: FC<ProjectDialogProps> = ({
 
       // Use the gasless signer from setupChainAndWallet
       // Keep modal open while submitting so users don't lose context/data.
-      await project.attest(signer as any, changeStepperStep).then(async (res) => {
+      //
+      // Retry the send on a transient wallet/bundler timeout (GAP-FRONTEND-1Y2:
+      // ethers "could not coalesce error" / "Wallet timeout"). The idempotency
+      // guard checks whether the timed-out attempt already landed (the slug is
+      // deterministic and is the project's indexed identifier) so we never
+      // double-create on a recovered timeout.
+      const projectSlugIdentifier = slug || project.uid;
+      const { result: res } = await attestWithRetry({
+        send: () => project.attest(signer as any, changeStepperStep),
+        hasAlreadyLanded: () => checkSlugExists(projectSlugIdentifier),
+      });
+
+      {
         showLoading("Indexing project...");
         let retries = 1000;
         const txHash = res?.tx[0]?.hash;
@@ -652,7 +665,7 @@ export const ProjectDialog: FC<ProjectDialogProps> = ({
         let fetchedProject: ProjectResponse | null = null;
 
         // First, poll using checkSlugExists to avoid 404 errors in Sentry
-        const projectIdentifier = slug || project.uid;
+        const projectIdentifier = projectSlugIdentifier;
         while (retries > 0) {
           // eslint-disable-next-line no-await-in-loop
           const exists = await checkSlugExists(projectIdentifier);
@@ -689,7 +702,7 @@ export const ProjectDialog: FC<ProjectDialogProps> = ({
             router.refresh();
           }, 1500);
         }
-      });
+      }
 
       reset();
       setStep(0);
--- a/components/Dialogs/ProjectDialog/index.tsx
+++ b/components/Dialogs/ProjectDialog/index.tsx
@@ -650,58 +650,63 @@ export const ProjectDialog: FC<ProjectDialogProps> = ({
       // deterministic and is the project's indexed identifier) so we never
       // double-create on a recovered timeout.
       const projectSlugIdentifier = slug || project.uid;
-      const { result: res } = await attestWithRetry({
+      const { result: res, recoveredByIdempotencyGuard } = await attestWithRetry({
         send: () => project.attest(signer as any, changeStepperStep),
         hasAlreadyLanded: () => checkSlugExists(projectSlugIdentifier),
       });
 
-      {
-        showLoading("Indexing project...");
-        let retries = 1000;
-        const txHash = res?.tx[0]?.hash;
-        if (txHash) {
-          await fetchData(INDEXER.ATTESTATION_LISTENER(txHash, chainId), "POST", {});
-        }
-        let fetchedProject: ProjectResponse | null = null;
+      showLoading("Indexing project...");
 
-        // First, poll using checkSlugExists to avoid 404 errors in Sentry
-        const projectIdentifier = projectSlugIdentifier;
-        while (retries > 0) {
+      // When a timed-out attempt was detected as already landed, `res` is null
+      // and we skip the ATTESTATION_LISTENER kick (there's no fresh tx hash to
+      // push) — the listener has already ingested the attestation. Either way
+      // we fall through to the same slug poll, which resolves the indexed
+      // project regardless of which send actually landed it.
+      const txHash = recoveredByIdempotencyGuard ? undefined : res?.tx[0]?.hash;
+      if (txHash) {
+        await fetchData(INDEXER.ATTESTATION_LISTENER(txHash, chainId), "POST", {});
+      }
+
+      let retries = 1000;
+      let fetchedProject: ProjectResponse | null = null;
+
+      // First, poll using checkSlugExists to avoid 404 errors in Sentry
+      const projectIdentifier = projectSlugIdentifier;
+      while (retries > 0) {
+        // eslint-disable-next-line no-await-in-loop
+        const exists = await checkSlugExists(projectIdentifier);
+        if (exists) {
+          // Project is indexed, now fetch the full details
           // eslint-disable-next-line no-await-in-loop
-          const exists = await checkSlugExists(projectIdentifier);
-          if (exists) {
-            // Project is indexed, now fetch the full details
-            // eslint-disable-next-line no-await-in-loop
-            fetchedProject = await getProject(projectIdentifier);
-            break;
-          }
-          retries -= 1;
-          // eslint-disable-next-line no-await-in-loop, no-promise-executor-return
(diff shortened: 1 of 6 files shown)


amaurymagalhaes:
## Tech Lead Review — Score: 78/100

**Issue being fixed:** Project creation abandons the whole attestation on a single transient wallet/bundler "Wallet timeout" during `eth_sendTransaction` inside `project.attest`, and reports it to Sentry as noise.
**Properly fixed:** Yes — the retry + telemetry-routing core correctly addresses the root cause (abandoned creation on a submission-failure timeout), not just the symptom.
**Could be less code:** No — the retry/backoff/telemetry-routing design is justified and not over-engineered. The leaner critique is conceptual, not line-count: the bespoke `AttestRetryExhaustedError` whose purpose is to game the two predicates (match `isRetryableChainError`, dodge `isTransientWalletTimeoutError`) is clever but fragile — the routing should be a structural marker on the error, not a carefully-worded message string one word away from breaking. The custom `cause` field also duplicates the standard ES2022 `Error` cause option.

- **Central defect — the idempotency guard does not prevent the double-create it claims to.** `hasAlreadyLanded() = checkSlugExists(slug)` reads the indexer (MongoDB), which only reflects the project *after* the on-chain attestation mines and is ingested — seconds to tens of seconds. The retry backoff was only 300ms then 600ms. For the genuinely dangerous case (the userOp WAS broadcast but the client gave up waiting), the guard runs far too early, returns `false`, and the retry resends — producing exactly the duplicate attestation the PR says it prevents. The guard only catches the near-impossible case where the prior attempt fully mined + indexed in under ~1s. The headline safety claim and the acceptance criterion "no duplicate project on a recovered timeout" are therefore overstated. (Resend is often safe when the bundler never broadcast the userOp, which is why this isn't catastrophic, but the guard gives little real protection in the unsafe sub-case.)
- **Brittle telemetry routing.** Drop-vs-report depends on the wrapper's message wording ("wallet/bundler timeout" must not contain "wallet timeout") rather than a structural flag.
- **Lost throw site.** `AttestRetryExhaustedError` overwrites `this.stack` with `cause.stack`, discarding the throw location.
- **Implicit recovered path.** The recovered-by-guard path returns `result: null` and silently relies on the downstream poll loop finding the landed project (works, but implicit).
- Pre-existing `signer as any` cast retained (out of scope).

### Recommendation
Accept the retry + telemetry-routing core — it correctly fixes the dominant symptom. Strengthen the idempotency guard so it can actually observe an in-flight tx before resending (poll on the indexer cadence rather than a single immediate check / 300ms backoff), or explicitly document that resend safety relies on the bundler not having broadcast. Replace the message-wording-based Sentry routing with a structural marker on the error.

### Applied fix (local commit 0be36c6e06cd4509f7875a475be83076cc781e4b — not pushed)
Applied the minimal fix for the central defect. The idempotency guard now POLLS `hasAlreadyLanded` on the indexing cadence (4 checks at 1500ms, matching `ProjectDialog`'s poll loop) via a bounded `waitForAlreadyLanded` helper before any resend, giving a slow-but-landed first attempt a realistic chance to be observed; the old 300/600ms backoff was removed in favor of this. Telemetry routing now keys off a structural `readonly isExhaustedRetry = true` flag on `AttestRetryExhaustedError` (plus an exported `isAttestRetryExhaustedError` type guard), checked first in `beforeSend`, instead of relying on the wrapper's message NOT matching the wallet-timeout string. Switched to standard `super(message, { cause })` and APPEND the cause stack ("Caused by: ...") instead of overwriting `this.stack`, preserving the throw site. `ProjectDialog` now branches explicitly on `recoveredByIdempotencyGuard` (skips the `ATTESTATION_LISTENER` POST when recovered, falls through to the slug poll). Added tests asserting the guard polls (in-flight-then-indexed timeline, bounded-window-then-resend) plus structural-flag routing and stack preservation. 32/32 tests pass, biome clean on changed util files, tsc clean. The pre-existing `signer as any` cast was left untouched.

Files changed:
- `utilities/attestWithRetry.ts`
- `instrumentation-client.ts`
- `components/Dialogs/ProjectDialog/index.tsx`
- `utilities/__tests__/attestWithRetry.test.ts`

This is a local proposal pending maintainer review.

PR #1763, opened 2026-07-07
fix(donor-research): honest outreach status while queued or blocked

amaurymagalhaes:
## Overview

Two status-feedback fixes for the outreach flows shipped in #1761, both found in live QA against the backend on show-karma/gap-indexer#2158.

## Problem

1. **Contradictory intro status.** While an intro sits in the outbox (`email_queued`), the candidate card showed the **"Intro sent"** badge next to an **"Intro queued"** detail line. The backend deliberately collapses queued+sent intros into the single `intro_sent` coarse status (it doesn't leak dispatch internals), so the disambiguation has to happen client-side — and the view already carries `intro.sentAt`.
2. **False success toast when outreach is blocked.** A POST to `diligence-requests` / `intro-requests` returns **202 with `coarseStatus: "blocked"`** when no contact could be resolved for the nonprofit — nothing was or will be emailed. Both dialogs toasted success ("Questions sent" / "Intro sent") and closed, right before the card flipped to "Couldn't reach".

## Solution

- `DiligenceStatusBadge` takes an `introQueued` flag: `intro_sent` + queued renders a muted **"Intro queued"** pill; the delivered state keeps the brand **"Intro sent"** pill. The redundant queued detail line is dropped — the relative-time line now renders only once delivered ("Intro sent today").
- `AskQuestionsDialog` and `ConnectDialog` branch on the 202 response's `coarseStatus`: blocked shows *"We couldn't find a contact for this nonprofit, so nothing was sent."* instead of the success toast.

## Testing

- 5 new unit tests (queued vs delivered badge, blocked toast in both dialogs); existing mutation mocks updated to pass realistic 202 payloads. Full donor-research + hooks suites: 147 files / 1,418 tests green; typecheck and Biome clean.
- Live-verified against a local indexer running show-karma/gap-indexer#2158: a blocked send now shows the couldn't-reach toast and the card badge stays consistent; a delivered intro shows "Intro sent / Intro sent today" with no queued contradiction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved messaging when an intro or question can’t be sent because no contact was found.
  * “Intro queued” and “Intro sent” statuses now display more accurately in diligence badges and related screens.
  * Dialogs now close correctly after blocked send attempts, while successful sends still show the usual success message.

### DIFF
--- a/src/features/donor-research/components/diligence/AskQuestionsDialog.tsx
+++ b/src/features/donor-research/components/diligence/AskQuestionsDialog.tsx
@@ -26,6 +26,7 @@ import { DILIGENCE_TEMPLATE_LIMITS } from "@/types/diligence";
 import { PAGES } from "@/utilities/pages";
 import { OutreachEmailPreview } from "./OutreachEmailPreview";
 import { getOutreachBodyIssue } from "./outreach-body";
+import { NO_CONTACT_FOUND_MESSAGE } from "./outreach-messages";
 import { makeQuestionId } from "./question-id";
 
 interface AskQuestionsDialogProps {
@@ -138,7 +139,7 @@ function AskQuestionsBody({
           // toast would contradict the "Couldn't reach" badge the card is
           // about to show.
           if (result.coarseStatus === "blocked") {
-            toast.error("We couldn't find a contact for this nonprofit, so nothing was sent.");
+            toast.error(NO_CONTACT_FOUND_MESSAGE);
           } else {
             toast.success("Questions sent");
           }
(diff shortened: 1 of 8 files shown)


amaurymagalhaes:
Addressed the review notes in 9db9584:

- **Shared constant**: the duplicated couldn't-find-a-contact copy now lives in `outreach-messages.ts` and both dialogs import it.
- **Allow-list vs deny-list branching (skipped, intentional)**: the asymmetry is semantic, not accidental. For **Ask Questions**, the 202's `coarseStatus` reflects the diligence request itself, so `blocked` is the only non-success outcome to special-case. For **Connect**, a *successfully queued* intro always outranks everything and reports `intro_sent`; a blocked intro falls through to whatever the diligence state is (e.g. `in_progress` when a diligence request is also live) — so a deny-list on `blocked` would mis-report those as success. The allow-list on `intro_sent` is the only correct check for the intro leg.

PR #1769, opened 2026-07-08
fix: payout-config rate-limit fan-out on milestones pages (Sentry GAP-FRONTEND-245)

amaurymagalhaes:
## Problem

Sentry **[GAP-FRONTEND-245](https://karma-hq.sentry.io/issues/?query=GAP-FRONTEND-245)** (`Error: Rate limit exceeded. Try again later.`) — **456 events / 12 users**, plus sibling **GAP-FRONTEND-23E**.

The admin **milestones report** page (`ReportMilestonePage`) renders decorative allocation badges by fetching one payout config per grant:

- `hooks/useCommunityMilestoneAllocations.ts` fires `GET /v2/payout-config/grant/:grantUID/public` per grant via `useQueries`.
- `ReportMilestonePage` fans out **50–200+ grants** per page load.

Since gap-indexer PR #2134 (2026-07-02) the indexer rate-limits public reads at **30 req/min/IP per route template**, so a single page load of ≥31 grants deterministically 429s the rest. Each failure produced one Sentry event (bursts of ~34/sec) because:

- `utilities/fetchData.ts` collapsed errors to a string tuple; the service re-threw a bare `Error` with **no status**.
- `utilities/queries/defaultOptions.ts` couldn't detect the 429 (did a blind extra retry, doubling the load).
- `errorManager` couldn't recognize it as expected load-shedding → captured every one.

Request-volume math: 150 grants → 150 requests → first 30 succeed, ~120 × (1 + 1 blind retry) ≈ **240 failing requests → 240 Sentry events per load**.

## Root cause

Per-grant fan-out against a per-route rate limit, compounded by a status-less error contract that defeated both retry classification and Sentry suppression.

### Layer 1 — batch the fetches
- New internal `usePayoutConfigsForGrants(grantUIDs, { communityUID? })`:
  - **With `communityUID`** → a single `getPayoutConfigsByCommunityPublic` request (endpoint already used by `PublicControlCenter`), filtered client-side via `select`, `staleTime` 5 min. Shares the cache entry with `usePayoutConfigsByCommunityPublic`.
  - **Without** → per-grant `useQueries` (project pages, few grants) with `gcTime` 30 min so navigation reuses cache.
- `ReportMilestonePage` now passes `community.uid` — the confirmed 245 offender collapses from 50–200+ requests to **1**.
- `useCommunityMilestoneAllocations` derives the community UID straight from the milestone payload (`CommunityMilestoneUpdate.communityUID`), so the community-updates page batches automatically.
- Extracted `byGrantPublic` / `byCommunityPublic` to `payoutDisbursementKeys` (removes a 3× duplicated key literal).

### Layer 2 — 429-aware fetch / retry / Sentry
- `fetchData.ts`: export `FetchDataError` (carries `status` + `retryAfterMs`), `parseRetryAfterMs`, and a `fetchDataThrow` wrapper. **Existing tuple contract untouched.**
- Payout-config services throw `FetchDataError` with the tuple status instead of a status-less `Error`.
- `transientErrors.ts`: add `isRateLimitError` (status 429 or rate-limit message) — additive only.
- `errorManager`: early-return on rate-limit (no `captureException`) and add a `Sentry.addBreadcrumb` so the pressure stays observable.
- `defaultOptions.ts`: retry 429 up to twice with capped exponential backoff (30s cap) that honors `Retry-After` when present, plus jitter to de-sync fan-out retries.
- The three payout-config service fetchers now call `fetchDataThrow` directly, so a server-sent `Retry-After` header flows end-to-end: axios headers -> `FetchDataError.retryAfterMs` -> service wrapper (preserved by `wrapFetchError`) -> `retryDelay`.

**Note on global retry-delay scope:** `retryDelay` is a QueryClient-wide default, so the new schedule (2s/4s/8s..., 30s cap, +jitter) also applies to non-429 retries that previously used 1s/2s with a 5s cap. This is intentional and safe: non-429 errors get at most 1-2 retries (unchanged counts), so the worst-case added wait is a few seconds on an already-failing request, while the longer cap prevents retry storms against a struggling indexer and the jitter de-syncs concurrent retries. User-facing loading states are unaffected on the success path.

Allocation amounts are decorative badges — pages render fine with empty maps (verified). No new toasts.

## Test coverage

All new/updated tests green (`vitest run`, 6 files / 153 tests):

- `__tests__/hooks/useCommunityMilestoneAllocations.test.ts` — 50 grants + communityUID → community fetcher called **exactly once**, per-grant **zero**; duplicate UIDs deduped; two hooks/one communityUID → one call; no-communityUID → one call per unique grant; identical `allocationMap` shape either way; UID derived from milestone payload.
- `__tests__/react-query/retry-and-stale-config.test.ts` — `FetchDataError(429)` retried ≤2, `retryAfterMs` honored + 30s cap, 401 not retried, 500 unchanged, silent recovery on 429→200.
- `__tests__/unit/utilities/sentry/transientErrors.test.ts` — `isRateLimitError` truth table.
- `__tests__/unit/utilities/errorManager.test.tsx` — 429 → not captured (breadcrumb added); 500 → captured.
- `__tests__/unit/services/payout-disbursement.service.test.ts` — tuple `[null, "Rate limit…", null, 429]` → thrown `FetchDataError`, status 429.
- `__tests__/unit/utilities/fetchData.test.tsx` — `fetchDataThrow` + `parseRetryAfterMs` (seconds / HTTP-date / cap).

Legacy suites that encoded the old "429 is never retried" contract (`utilities/queries/__tests__/defaultOptions.test.ts`, `__tests__/regression/retry-429-exclusion.test.ts`, `__tests__/trust/api/react-query-integration.test.ts`, `__tests__/integration/services/react-query-integration.test.ts`) were updated and retitled to the new bounded-retry contract (initial + 2 retries = 3 calls, `retryDelay: 0` in QueryClient-based tests). The payout-service trust/integration mocks now expose the real `FetchDataError` and a delegating `fetchDataThrow`.

Consumer component tests (`PublicControlCenter`, `ActivityFeed`, `MilestonesList` ×2, `InboxMilestoneDetail`) — 60/60 still green. `tsc --noEmit` clean.

> MSW request-counter integration case skipped: the `payouts.handlers.ts` harness has no public payout-config route and these authless routes need extra server/counter wiring. The same guarantees are already covered by the hook tests (service-call count is 1:1 with network hits) and the React Query 429→200 recovery integration test.

## Sentry

- Primary: **GAP-FRONTEND-245** — `Error: Rate limit exceeded. Try again later.` (456 events / 12 users)
- Sibling: **GAP-FRONTEND-23E**

Both are resolved by (1) eliminating the fan-out on the offending page and (2) suppressing residual 429s from capture while retrying them with backoff.

## CI notes

- The github-actions "Anti-Pattern Violations" comment (`EMPTY_CATCH` in `payout-disbursement.service.ts` L105/135/161 and `fetchData.ts` L116/137) is a **false positive on pre-existing catch blocks**: every flagged catch either calls `errorManager(...)` and rethrows (service) or returns the error tuple / rethrows (fetchData). None are empty; none were introduced by this PR.

## Follow-ups

- **gap-indexer (separate PR):** the public payout-config 429 responses do **not** currently emit a `Retry-After` header. The FE now honors it end-to-end (`parseRetryAfterMs` → `FetchDataError.retryAfterMs` → `wrapFetchError` → `retryDelay`), so adding `Retry-After` on the indexer's 429s would let the client back off exactly as long as needed instead of using its exponential fallback.
- `MilestonesReview` (single-grant) and project-page callers intentionally stay per-grant — batching one grant against the community-wide endpoint would fetch all community configs for one badge.

### DIFF
--- a/__tests__/hooks/useCommunityMilestoneAllocations.test.ts
+++ b/__tests__/hooks/useCommunityMilestoneAllocations.test.ts
@@ -1,9 +1,44 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { renderHook, waitFor } from "@testing-library/react";
+import { createElement, type ReactNode } from "react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import {
   buildGrantAllocationTotalMap,
   buildMilestoneAllocationMap,
   resolveTokenSymbol,
+  useCommunityMilestoneAllocations,
+  useMilestoneAllocationsByGrants,
 } from "@/hooks/useCommunityMilestoneAllocations";
 import type { PayoutGrantConfig } from "@/src/features/payout-disbursement/types/payout-disbursement";
+import type { CommunityMilestoneUpdate } from "@/types/community-updates";
+
+vi.mock("@/src/features/payout-disbursement/services/payout-disbursement.service", () => ({
+  getPayoutConfigByGrantPublic: vi.fn(),
+  getPayoutConfigsByCommunityPublic: vi.fn(),
+}));
+
+import * as payoutService from "@/src/features/payout-disbursement/services/payout-disbursement.service";
+
+const mockGetByGrant = vi.mocked(payoutService.getPayoutConfigByGrantPublic);
+const mockGetByCommunity = vi.mocked(payoutService.getPayoutConfigsByCommunityPublic);
+
+function makeConfig(grantUID: string): PayoutGrantConfig {
+  return {
+    grantUID,
+    milestoneAllocations: [
+      { id: `alloc-${grantUID}`, milestoneUID: `ms-${grantUID}`, label: "M1", amount: "1000" },
+    ],
+  } as PayoutGrantConfig;
+}
+
+function createWrapper() {
+  const queryClient = new QueryClient({
+    defaultOptions: { queries: { retry: false, gcTime: Number.POSITIVE_INFINITY } },
+  });
+  const wrapper = ({ children }: { children: ReactNode }) =>
+    createElement(QueryClientProvider, { client: queryClient }, children);
+  return { wrapper, queryClient };
+}
 
 describe("resolveTokenSymbol", () => {
   it("should_return_grant_currency_when_provided", () => {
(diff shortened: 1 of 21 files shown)


amaurymagalhaes:
Closing in favor of https://github.com/show-karma/gap-app-v2/pull/1808

PR #1771, opened 2026-07-08
feat: gate application "Applicant" section to authorized viewers (DEV-518)

brunodmsi:
## Summary

On the whitelabel application detail page, the sidebar **"Applicant"** always showed **"Anonymous"** — even for reviewers/admins authorized to see who applied. This shows the real applicant to authorized viewers and hides the section from everyone else.

Closes DEV-518.

## Root cause

The page (`.../applications/[applicationId]/page.tsx`) is a **Server Component** that fetches `GET /v2/funding-applications/:id` **server-side**. There's no Privy token on the server (Privy keeps it in localStorage), so the request is effectively **unauthenticated**, and the backend **redacts** applicant identity (`ownerAddress`, and `applicantEmail` for public). Both arrive empty → `ApplicationInfoCard.deriveApplicant()` falls through to "Anonymous". Because the card is fed by that redacted SSR payload, even a logged-in reviewer/admin saw "Anonymous".

## Approach — reuse the existing authed fetch (no extra request)

DEV-517 already added a client-side authenticated re-fetch of the **full application** (`getApplicationStatusHistory` → `GET /v2/funding-applications/:ref`) to un-redact the private status-change reasons — it just discarded everything except `statusHistory`. That same response already carries the applicant identity.

So instead of adding a second call, the detail page now uses the general **`useApplication`** hook (same endpoint, full app, `initialData` = SSR payload, `enabled: authenticated`) for **one** authenticated re-fetch that restores **both** `statusHistory` reasons **and** the applicant identity. The detail page drops its separate `useApplicationStatusHistory` call; the hook stays (the edit page still uses it).

## What changed (3 files + 1 test)

1. **`ApplicationPageClient`** — swap `useApplicationStatusHistory` → `useApplication`; derive both `statusHistory` and the applicant identity from the one live application; compute `canViewApplicant = viewerRole !== "guest"`.
2. **`ApplicationSidebar`** — thread `canViewApplicant` through; render from the live application.
3. **`ApplicationInfoCard`** — render the "Applicant" section **only** when `canViewApplicant && (applicantEmail || ownerAddress)`. The "Anonymous" fallback is gone.

## Behavior

| Viewer | Applicant section |
|--------|-------------------|
| Applicant (owner) | shown — their own identity |
| Reviewer / community admin / super admin | shown — real applicant email/address |
| Guest / public | **hidden entirely** (no "Anonymous") |

No new PII exposure: the backend already hard-gates whether a viewer receives the applicant's address/email; the FE only surfaces identity to viewers the backend already authorizes, and stops rendering a useless "Anonymous" row.

## Verification

- `pnpm run typecheck` → 0 errors
- `pnpm run build` → success
- Tests: new `ApplicationInfoCard.test.tsx` (hidden for unauthorized; **never** renders "Anonymous"; shows email / truncated address for authorized) + existing whitelabel smoke suite + DEV-517's status-history hook/service tests — all pass.

## Notes

Rebased onto current `main` (which includes DEV-517 #1760). Related to DEV-513 (identity display).

### DIFF
--- a/app/community/[communityId]/(whitelabel)/applications/[applicationId]/ApplicationPageClient.tsx
+++ b/app/community/[communityId]/(whitelabel)/applications/[applicationId]/ApplicationPageClient.tsx
@@ -13,8 +13,8 @@ import { Role } from "@/src/core/rbac/types/role";
 import { CommentTimeline } from "@/src/features/application-comments/components/CommentTimeline";
 import { PublicComments } from "@/src/features/application-comments/components/PublicComments";
 import { MilestonesTab } from "@/src/features/applications/components/MilestonesTab";
+import { useApplication } from "@/src/features/applications/hooks/use-application";
 import { useApplicationAccess } from "@/src/features/applications/hooks/use-application-access";
-import { useApplicationStatusHistory } from "@/src/features/applications/hooks/use-application-status-history";
 import type { IFundingApplication, ProgramWithFormSchema } from "@/types/funding-platform";
 import type { Application, ApplicationStatus, FundingProgram } from "@/types/whitelabel-entities";
 import { formatDate } from "@/utilities/formatDate";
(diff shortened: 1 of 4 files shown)


brunodmsi:
Reworked per review: my branch was cut from a stale `main` that predated DEV-517 (#1760). Rebased onto current main and **consolidated** — instead of adding a second `useApplication` re-fetch, the detail page now reuses one authenticated full-application fetch (same endpoint DEV-517's `getApplicationStatusHistory` already hits) to restore **both** the status-change reasons and the applicant identity. Dropped the detail page's `useApplicationStatusHistory` call (hook kept for the edit page). No extra network request.

PR #1809, opened 2026-07-09
fix: handle project impacts 404 gracefully + V1 bridge until V2 ships (GAP-FRONTEND-24Z)

amaurymagalhaes:
## Problem

The project **About** page reports a steady stream of expected `404` errors to Sentry (GAP-FRONTEND-24Z). Every visit to a project profile fires a request for project impacts that the indexer answers with `404`, which `errorManager` then forwards to Sentry as an error — creating noise and masking the fact that the impacts feature was not returning any data at all.

## Root cause

PR #1752 introduced `INDEXER.V2.PROJECTS.IMPACTS` pointing at `/v2/projects/:idOrSlug/impacts`. That route **does not exist on the indexer**. The V2 project router (`app/modules/v2/api/routes/project/project.routes.ts`) only declares `updates`, `grants`, and `grants/:programId/milestones` — there is no V2 `impacts` route. The only impacts route is the V1 one registered under the `projects` prefix: `router.get('/:idOrSlug/impacts', ...)` → `/projects/:idOrSlug/impacts`.

So **100% of impacts requests 404'd**: the feature silently returned an empty list AND every call was reported to Sentry.

## Fix (what & why)

1. **Repoint to the route that actually exists.** Move `IMPACTS` from the `INDEXER.V2.PROJECTS` group to `INDEXER.PROJECT` → `/projects/:idOrSlug/impacts`. This restores the feature: the V1 controller returns a properly-shaped `ProjectImpact[]` (EAS `ProjectImpact` attestations with `ProjectImpactVerified` verifications), matching the FE `ProjectImpact` type. A comment records that this should flip back to a V2 group once the backend ships a V2 counterpart.
2. **Treat an expected `404` as an empty result, not a Sentry error.** Read the status from `fetchData` and short-circuit `status === 404` to `return []` before calling `errorManager`, mirroring `project-grants.service` and `project-updates.service`. Genuine unexpected errors (e.g. `5xx`) are still reported. The `!Array.isArray(data)` branch also returns `[]` for the V1 controller's `null` body (unknown slug returns `200` + `null`, not `404`).

Net effect: the impacts feature works again, expected 404s no longer reach Sentry, and genuine failures are still surfaced.

## Tests

- New `__tests__/unit/services/project-impacts.service.test.ts`: success path, **regression** (404 → `[]`, no Sentry report), 5xx → reported, `200` + `null` body → `[]`, wire-contract lock (calls `/projects/:slug/impacts`), positional `isAuthorized`/`signal` forwarding.
- Updated `services/__tests__/project-profile-public-fetch-auth.test.ts` to assert the V1 URL.
- All 13 tests pass; Biome clean; full `tsc --noEmit` passes (pre-push hook).

## Review

Adversarial self-review score: **96/100**. Verified against the indexer source that the V2 impacts route genuinely does not exist and the V1 route returns the expected shape (so this is a real root-cause fix, not a Sentry mask). Confirmed expected-vs-unexpected error separation matches sibling services, no `any` casts, no silent catches, and regression coverage that fails before / passes after. Fixed one type issue found in review (`vi.MockedFunction` used as a namespace failed `tsc`; switched to `ReturnType<typeof vi.fn>`).

Sentry: https://karma-crypto-inc.sentry.io/issues/GAP-FRONTEND-24Z

Fixes GAP-FRONTEND-24Z

## Summary by CodeRabbit

* **Bug Fixes**
  * Project impact requests now use the correct endpoint.
  * Unknown projects and missing impact data return an empty result without generating unnecessary errors.
  * Unexpected server errors continue to be reported appropriately.

* **Tests**
  * Added coverage for successful responses, missing data, error handling, endpoint selection, and request option forwarding.

### DIFF
--- a/__tests__/unit/services/project-impacts.service.test.ts
+++ b/__tests__/unit/services/project-impacts.service.test.ts
@@ -101,12 +101,12 @@ describe("project-impacts.service", () => {
       expect(mockErrorManager).not.toHaveBeenCalled();
     });
 
-    it("calls the V2 impacts endpoint (locks the wire contract)", async () => {
+    it("calls the V1 impacts endpoint (locks the wire contract)", async () => {
       mockFetchData.mockResolvedValueOnce([mockImpacts, null, null, 200]);
 
       await getProjectImpacts("my-project-slug");
 
-      expect(mockFetchData.mock.calls[0][0]).toBe("/v2/projects/my-project-slug/impacts");
+      expect(mockFetchData.mock.calls[0][0]).toBe("/projects/my-project-slug/impacts");
     });
 
     it("forwards isAuthorized and signal positionally", async () => {
--- a/__tests__/unit/services/project-impacts.service.test.ts
+++ b/__tests__/unit/services/project-impacts.service.test.ts
@@ -65,7 +65,7 @@ describe("project-impacts.service", () => {
     it("regression GAP-FRONTEND-24Z: returns [] and does NOT report to Sentry on 404", async () => {
       mockFetchData.mockResolvedValueOnce([
         null,
-        "Route GET:/v2/projects/test-project/impacts not found",
+        "Route GET:/projects/test-project/impacts not found",
         null,
         404,
       ]);
--- a/__tests__/unit/services/project-impacts.service.test.ts
+++ b/__tests__/unit/services/project-impacts.service.test.ts
@@ -101,12 +101,12 @@ describe("project-impacts.service", () => {
       expect(mockErrorManager).not.toHaveBeenCalled();
     });
 
-    it("calls the V1 impacts endpoint (locks the wire contract)", async () => {
+    it("calls the V2 impacts endpoint (locks the wire contract)", async () => {
       mockFetchData.mockResolvedValueOnce([mockImpacts, null, null, 200]);
 
       await getProjectImpacts("my-project-slug");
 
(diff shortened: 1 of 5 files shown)


amaurymagalhaes:
**Updated approach (v2-aligned).** Reverted the temporary V1 repoint — the frontend stays on `/v2/projects/:idOrSlug/impacts` per the V1→V2 migration. The defensive change remains: a 404 from the impacts endpoint is treated as an empty result and is not reported to Sentry.

The backend counterpart — a `GET /v2/projects/:idOrSlug/impacts` read route on gap-indexer (the v2 project read path already loads `impacts`) — is in a paired gap-indexer PR. Until that ships, impacts render empty on v2 (no crash, no Sentry noise).

amaurymagalhaes:
**Switched to a temporary V1 bridge.** The V2 route `/v2/projects/:id/impacts` isn't deployed yet (backend PR show-karma/gap-indexer#2178, in review), so this now points the impacts service at the working V1 route `/projects/:id/impacts` — impacts render today instead of 404'ing. The graceful 404→empty + Sentry-suppression is unchanged.

Flip-back is a one-move change (documented inline at both the `INDEXER` constant and the service): move `IMPACTS` back to `V2.PROJECTS` once #2178 ships and deploys. `indexer.ts` stays at its size baseline (net-0 line change).

PR #1818, opened 2026-07-10
refactor: open nonprofit-research reports on the advisor route for staff

Arthh:
## Summary

Staff now open nonprofit-research reports on the regular `/nonprofit-research/[reportId]` page instead of a separate admin page. Pairs with show-karma/gap-indexer#2177, which serves both roles from the same report endpoint.

- Staff get the full advisor report experience — criteria disclosure, 5s polling, and live SSE progress for in-flight reports — instead of the trimmed shared-variant snapshot.
- Owner-only write controls (share management, weights panel, diligence actions) render only for the report's own advisor. Staff get a read-only `staff` variant of the same brief, since those endpoints are advisor-scoped.
- The old `/admin/nonprofit-research/[reportId]` route is removed entirely (it was internal-only; old URLs now 404 — per review feedback, no redirect stub).
- The admin advisors list links straight to the advisor route; the admin-only report view component, hook, service call, and route constants are removed.

## Testing

- Updated the advisors-list, hook, and service tests; donor-research suites pass (124 tests).
- `pnpm typecheck` and Biome pass.


app/admin/nonprofit-research/[reportId]/page.tsx
```diff
@@ -1,19 +1,16 @@
...
+/**
+ * Legacy staff report URL. The staff report view now lives on the regular
+ * advisor route — the API grants the staff allowlist an unscoped read on the
+ * same endpoint — so this route only preserves old bookmarks/links.
+ */
 export default async function Page({ params }: PageProps) {
   const { reportId } = await params;
-  return <AdminReportView reportId={reportId} />;
+  permanentRedirect(PAGES.DONOR_RESEARCH.REPORT(reportId));
 }
```
Comment (amaurymagalhaes) [inline app/admin/nonprofit-research/[reportId]/page.tsx]: I think you can just remove it.
Since no one is using it.

amaurymagalhaes (changes requested):
Looks good, just remove that redirection(whole file).

Arthh:
Removed the whole route — the admin overview links straight to the advisor page now.

PR #1830, opened 2026-07-11
fix(admin): stop the community-admins batch from hanging /admin on a slow chain

brunodmsi:
## Overview

Stops the `/admin` communities page from hanging (and eventually timing out) when one community sits on a slow or dead RPC chain. Fault-isolates the community-admins batch fetch so a single bad community degrades to "admin data unavailable" instead of freezing the whole page.

## Problem

The admin page loads **all** communities an owner can see (~130+ on staging) and fetches their admins in **one** call to `POST /communities/admins/batch`. That V1 endpoint:

- Authorizes and resolves **each** community with a **live on-chain / subgraph call** (`isCommunityAdmin` → `resolver.isAdmin`), with **no per-call timeout**.
- Caps a request at **200 UIDs** (Zod `.max(200)`, validated before auth).

So the single request fans out **O(N)** upstream calls. On staging there are **2 communities on chainID 1329 (sei)** whose RPC is dead (TLS-decommissioned → `JsonRpcProvider "failed to detect network"`). Those checks hang, and because the frontend gives indexer calls a **6-minute** timeout, the admin page freezes until the gateway 504s. Separately, any owner with **>200** communities gets an outright **400** on the cap.

**Reproduced:** `POST /communities/admins/batch` with 201 UIDs (no auth) → immediate **400**; with ≤200 → passes validation then runs the O(N) authorized fan-out. Staging community count = 131 (< 200), so the staging failure is the O(N) hang on the dead-RPC community, not the 400.

## Solution

`getCommunityAdminsBatch` now:

- **Chunks** UIDs into ≤ **20** per request (well under the 200 cap → no more 400).
- Gives **each chunk its own 15s `AbortController` timeout**.
- Runs chunks in **bounded-concurrency waves** (`Promise.allSettled`, concurrency 6).
- Drops a **hung/failed chunk** to `subgraph_unavailable` for its communities — already rendered as *"Admin data unavailable"* by `CommunityAdminCard` — while every healthy chunk resolves in seconds.

Public API of the function is unchanged (`(uids) => Promise<CommunityAdmin[]>`), so no call-site changes.

## Why This Approach

The endpoint itself is **V1 (frozen / write-protected)**, so the fix lives at the only layer we can safely change. It addresses **both** failure modes (the 200-cap 400 and the O(N) dead-RPC hang) and degrades gracefully rather than all-or-nothing. The durable fix — a **V2** batch endpoint that resolves admins from **indexed DB data** instead of live RPC — is filed as a follow-up (see below).

## Testing Steps

1. On `/admin` as an owner, the page now loads in seconds even when a community is on a slow/dead chain; affected cards show *"Admin data unavailable"* instead of a spinner that never resolves.
2. Network tab: instead of one `/communities/admins/batch` call, you'll see several small ones (≤20 UIDs each); a stuck one aborts at 15s and the rest render.

Automated: `pnpm test services/__tests__/communities.service.batch.test.ts` — 7 tests (chunking to ≤20, cross-chunk merge, abort-signal passed, **fault isolation** — one failed chunk → `subgraph_unavailable` while healthy chunks stay `ok`, dedupe + order preservation, missing UID → `community_not_found`, empty input). `tsc` clean, `pnpm build` green, React Doctor 100/100 on the diff.

## Follow-ups

- **V2 batch-admins endpoint** resolving admins from indexed Mongo data (no live on-chain/subgraph RPC) — the real O(1) fix. (DEV follow-up.)
- Dead **sei (1329)** RPC swap on the indexer (tracked separately).

## Breaking Changes

None. Additive resilience; function signature and response shape unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when retrieving community administrators across large batches.
  * Added automatic handling for temporary request failures and timeouts.
  * Preserved result ordering and duplicate inputs while providing clear statuses for unavailable or missing communities.

* **Tests**
  * Added coverage for batching, retries, timeouts, cancellation, deduplication, ordering, and partial failures.

### DIFF
--- a/services/__tests__/communities.service.batch.test.ts
+++ b/services/__tests__/communities.service.batch.test.ts
@@ -1,0 +1,105 @@
+vi.mock("@/utilities/fetchData");
+vi.mock("@/components/Utilities/errorManager", () => ({ errorManager: vi.fn() }));
+
+import fetchData from "@/utilities/fetchData";
+import { getCommunityAdminsBatch } from "../communities.service";
+
+const mockFetchData = fetchData as unknown as ReturnType<typeof vi.fn>;
+
+const uid = (n: number): string => `0x${n.toString(16).padStart(64, "0")}`;
+
+const okItem = (communityUID: string) => ({
+  communityUID,
+  admins: [{ user: { id: `admin-${communityUID}` } }],
+  status: "ok" as const,
+});
+
+/** Every requested chunk resolves successfully with an admin per UID. */
+const respondOkForAllChunks = () => {
+  mockFetchData.mockImplementation((_endpoint, _method, axiosData) => {
+    const uids = (axiosData as { communityUIDs: string[] }).communityUIDs;
+    return Promise.resolve([{ data: uids.map(okItem) }, null, null, 200]);
+  });
+};
+
+const bodyUIDs = (callIndex: number): string[] =>
+  (mockFetchData.mock.calls[callIndex][2] as { communityUIDs: string[] }).communityUIDs;
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+describe("getCommunityAdminsBatch", () => {
+  it("should_return_empty_array_and_make_no_request_when_given_no_uids", async () => {
+    const result = await getCommunityAdminsBatch([]);
+
+    expect(result).toEqual([]);
+    expect(mockFetchData).not.toHaveBeenCalled();
+  });
+
+  it("should_split_requests_into_chunks_of_at_most_20_uids", async () => {
+    respondOkForAllChunks();
+    const uids = Array.from({ length: 45 }, (_, i) => uid(i));
+
+    await getCommunityAdminsBatch(uids);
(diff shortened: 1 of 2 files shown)


brunodmsi:
Hardened the chunked fetch (be4fc490) to de-risk it vs the old single-request behavior:
- **Retry once on transient failure**, but *not* on timeout/abort — a hung upstream is slow/unreachable, so retrying would only double the wait before degrading. This restores the transient-failure recovery the old React Query retry used to provide.
- **Per-chunk timeout 15s → 25s**, so it sits *above* the backend's own worst case (10s subgraph + one retry ≈ 20s); a slow-but-working chunk is no longer aborted and falsely marked `subgraph_unavailable`.
- **Chunk concurrency 6 → 4** to reduce peak on-chain/subgraph load.

Tests now cover transient-retry recovery and no-retry-on-timeout degradation (9 total). tsc clean, build green, React Doctor 100/100.

PR #1851, opened 2026-07-15
feat(seo): add gap.karmahq.xyz domain-redirect management tool

amaurymagalhaes:
## Root cause: edge redirect preempts the middleware

`gap.karmahq.xyz` carries a **Vercel project-domain redirect** (301 → `www.karmahq.xyz`) that fires **at the edge, before** the Next.js middleware runs. For `/project/<slug>/grants` that means **two hops**:

1. `gap/project/<slug>/grants` → **301** (Vercel host-swap) → `www/project/<slug>/grants`
2. `www/project/<slug>/grants` → **308** (middleware `grants`→`funding`) → `www/project/<slug>/funding`

The middleware **already** collapses host-swap + `grants`→`funding` into a **single 308** to `www/project/<slug>/funding` (pinned by `__tests__/middleware-indexability.test.ts`). It just never gets the chance, because the edge redirect answers first. Clearing the Vercel domain redirect lets the app serve `gap.karmahq.xyz`, so the middleware produces the one-hop result the indexability monitor expects.

## What this PR adds (no live change)

A dependency-free Node 20 CLI + a **manual-only** workflow to make that Vercel change safely. **It performs no mutation until manually dispatched** (`workflow_dispatch`): the workflow's GitHub token is scoped `contents: read`, and the Vercel project-domain mutation itself stays confirmation-gated and manual (nothing runs automatically).

- `scripts/vercel-domain-redirect/manage-gap-redirect.mjs` — CLI with three modes.
- `scripts/vercel-domain-redirect/__tests__/manage-gap-redirect.test.mjs` — 42 focused tests (injected `fetch`/`fs`, no network).
- `.github/workflows/gap-domain-redirect.yml` — `workflow_dispatch` (choice `inspect|apply|rollback` + confirmation), `permissions: contents: read`, `timeout-minutes: 5`, Node 20, focused tests run before the CLI, all inputs/secrets passed via `env` (no `${{ }}` in `run`), artifacts uploaded always.

## Modes & safety

- **inspect** — GET the Vercel v9 project-domain config and print it. Read-only.
- **apply** (exact confirmation `CLEAR gap.karmahq.xyz`) — validate domain+project **identity** and `verified === true`; require the current state to be the expected `redirect: www.karmahq.xyz` / `301`; capture a rollback-state artifact; `PATCH {redirect:null,redirectStatusCode:null}`; validate the PATCH response (identity, `verified===true`, cleared, unchanged `gitBranch`); then **probe** `https://gap.karmahq.xyz/project/paraswap/grants` with `redirect: manual` and require exactly one **301/308** hop resolving to `https://www.karmahq.xyz/project/paraswap/funding`. **Any post-mutation failure auto-rolls back** the captured state (validated identity + `verified===true` + exact `redirect/status/gitBranch`), reporting both primary and rollback errors. Already-cleared state is idempotent **only if the live probe passes**.
- **rollback** (exact confirmation `RESTORE gap.karmahq.xyz -> www.karmahq.xyz`) — GET first; idempotent no-op if already `www/301`; refuse any non-cleared state with no PATCH; otherwise restore `www/301` and validate the response.

Additional guards: a real **15s request timeout** that stays armed through Vercel response-body consumption and the live probe (injectable for fast tests); the API **token is never printed** (all logs + artifacts pass through a redactor, exercised even when an error message contains the token); `verified` is checked as the **literal boolean `true`** (non-boolean values fail closed); the v9 endpoint encodes `projectId`/`teamId` (tested with special characters) and every API call uses `Authorization: Bearer`.

## Tests

`node --test` over the suite: **42 passed / 42 total**. Covers inspect/apply/rollback happy paths, confirmations, precondition/identity/verified refusals with no mutation, PATCH request URL/auth/body, malformed API bodies, single-hop resolution, already-clear/already-restored idempotency, auto-rollback success/failure, stalled API body + stalled probe timeouts, strict-boolean `verified`, and token redaction. Biome clean; `node --check` OK; workflow YAML validated.

## Tracking

Part of show-karma/super-gap#55 (indexability remediation) — this resolves the `gap-legacy-grants-redirect` two-hop finding, pending a manual dispatch of the workflow. Referenced, not closed.

### DIFF
--- a/.github/workflows/gap-domain-redirect.yml
+++ b/.github/workflows/gap-domain-redirect.yml
@@ -31,6 +31,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          # This job only reads the repo (tests + CLI) and needs no git-authenticated
+          # writes; don't persist GITHUB_TOKEN into .git/config where an uploaded
+          # artifact could leak it (zizmor artipacked).
+          persist-credentials: false
 
       - name: Set up Node
         uses: actions/setup-node@v4
--- a/.github/workflows/gap-domain-redirect.yml
+++ b/.github/workflows/gap-domain-redirect.yml
@@ -1,0 +1,68 @@
+name: Gap domain redirect (manual)
+
+# Manually inspect / clear / restore the Vercel project-domain redirect on
+# gap.karmahq.xyz. Clearing the redirect lets the app serve gap.karmahq.xyz so
+# the middleware collapses the host-swap + grants->funding normalization into a
+# single 308 (see scripts/vercel-domain-redirect/manage-gap-redirect.mjs).
+on:
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "Action to perform"
+        type: choice
+        required: true
+        options:
+          - inspect
+          - apply
+          - rollback
+      confirmation:
+        description: "apply: CLEAR gap.karmahq.xyz | rollback: RESTORE gap.karmahq.xyz -> www.karmahq.xyz"
+        type: string
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+
+jobs:
+  manage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      # Focused tests run BEFORE any live CLI invocation.
+      - name: Run focused tests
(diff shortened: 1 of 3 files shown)


amaurymagalhaes:
Closing without merge: the Vercel project-domain redirect was cleared manually in the dashboard, so the workflow/CLI added by this PR is no longer needed.

Live verification on 2026-07-17:

`https://gap.karmahq.xyz/project/paraswap/grants` → **308** → `https://www.karmahq.xyz/project/paraswap/funding`

The destination page loads successfully, and the previous two-hop chain is now a single canonical redirect.

PR #1854, opened 2026-07-15
feat(portfolio-reports): editable report titles + type filter (DEV-520)

brunodmsi:
Pairs with **show-karma/gap-indexer#2202** (adds the `title` field this reads). Closes [DEV-520](https://linear.app/showkarma/issue/DEV-520/erin-oconnor-edit-sorting-options-for-report-titles-represented-on).

## Problem

`app.filpgf.io` → `communitySlug: "filecoin"` (`whitelabel-config.ts:49`) → this list page. Live prod data is 6 published reports across 3 configs, and the card title was `reportConfigName`, shared by every run of that config — so readers saw:

```
07/09/26   Bi-Weekly Progress Report      Published Jul 15, 2026
07/06/26   Monthly Pods Report            Published Jul 6, 2026
06/30/26   Filecoin ProPGF Monthly        Published Jul 8, 2026
06/22/26   Bi-Weekly Progress Report      Published Jun 24, 2026
06/17/26   Monthly Pods Report            Published Jun 25, 2026
06/03/26   Filecoin ProPGF Monthly        Published Jun 3, 2026
```

Every title twice, nothing naming the period, and no way to narrow to one kind of report. Erin: *"Identifying which monthly report aligns with which month is not super intuitive without clicking into each report."*

## Change

**Editable titles.** Renders the new per-report `title`, falling back through the existing chain:

```
title  ??  reportConfigName  ??  extractTitle(content)  ??  runDate label
 ↑ new       ↑ today's behaviour — untitled reports are unchanged
```

Admins author it in the report editor (title input beside the content textarea). An emptied field sends `null`, restoring the config-name fallback. The editor header now leads with the title when set. The covered period is knowable only to a human — `runDate` is the generation date and nothing records the range — so this is authored, not derived. Rationale in gap-indexer#2202.

**Type filter.** The report *config* is effectively the report type, and both `reportConfigId` and `reportConfigName` already ship on the published payload — so this filters **client-side with zero API change**. URL-synced with nuqs (`?type=<configId>`), so a filtered view is shareable and survives reload. Hidden when a community has only one config.

Also: list item extracted into a `React.memo` component; the count now uses `pluralize` (was a manual ternary) and moves with the filter; the timeline scrubber and scroll-spy track the filtered list.

## Testing

- New `PublicReportListPage.test.tsx` (14 tests, none existed before): all 4 title-fallback tiers, filter narrowing/reset, options deduped per config, filter hidden at one type, pluralized count, stale-filter reset, plus loading/error/empty states.
- `PortfolioReportEditorPage.test.tsx` +5: title seeding (set and untitled), save sends typed title, emptied field sends `null`, Save enables on title-only change.
- Deploy-order safety is pinned by a test: with `title` **absent** (not null), as it would be
  if this ships before gap-indexer#2202, every report still renders the config name.
- Fixtures mirror the real Filecoin shape (3 configs × 2 runs).
- Mutation-checked: removing `report.title ??` from the resolver fails the title test and nothing else, so the coverage is real rather than vacuous.
- 88 portfolio-report tests green; `tsc --noEmit` clean; `pnpm run build` succeeds; Biome clean on changed files.

## Notes for review

- `PortfolioReportEditorPage` trips Biome's complexity warning — but it already did on `main` (34 → 39, limit 20). Advisory, pre-existing; extracting the edit dialog felt like scope creep here. Happy to split it out if preferred.
- No Next.js proxy sits in this seam (the browser calls the indexer directly via `NEXT_PUBLIC_GAP_INDEXER_URL`), so the proxy-contract-test guardrail doesn't apply; the wire contract is covered by controller tests in gap-indexer#2202.

## Smoke results

Not yet run — cross-service gate pending. This must not merge until both branches are on the same preview and the DEV-520 flows are walked end-to-end. Deploying this ahead of gap-indexer#2202 is safe but inert: `title` is `undefined`, so every report falls back to the config name exactly as today.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom report titles (editing, saving, clearing) with title-aware unsaved-change detection and save behavior.
  * Updated report headers and public listings to prefer the custom title with fallback labels.
  * Added URL-backed filtering of public reports by report type, including type-aware empty states and restore-to-all.
* **Bug Fixes**
  * Prevented title-only updates from overwriting refreshed report content while the editor is open.
  * Ensured regeneration and type changes don’t leave stale dirty/active timeline state.
* **Tests**
  * Expanded coverage for title editing behavior and type filtering UI states.

### DIFF
--- a/__tests__/components/Pages/Admin/PortfolioReports/PortfolioReportEditorPage.test.tsx
+++ b/__tests__/components/Pages/Admin/PortfolioReports/PortfolioReportEditorPage.test.tsx
@@ -48,6 +48,8 @@ const baseReport = {
   communityId: "community-1",
   runDate: "2026-03-15",
   status: "draft",
+  title: null,
+  reportConfigName: "Monthly Pods Report",
   content: "<p>Server content</p>",
   dataSnapshot: {},
   modelId: "gpt-4.1",
@@ -230,6 +232,86 @@ describe("PortfolioReportEditorPage", () => {
         screen.queryByRole("heading", { name: /discard unsaved edits/i })
       ).not.toBeInTheDocument();
       expect(screen.getByRole("heading", { name: /^regenerate report/i })).toBeInTheDocument();
+    });
+  });
+
+  describe("report title", () => {
+    it("should_seed_title_input_with_stored_title_when_dialog_opens", async () => {
+      const user = userEvent.setup();
+      mockUsePortfolioReport.mockReturnValue({
+        data: { ...baseReport, title: "Monthly Pods Report — June 2026" },
+        isLoading: false,
+      } as any);
+
+      render(<PortfolioReportEditorPage community={community} reportId="report-1" />);
+      await user.click(screen.getByRole("button", { name: /^edit$/i }));
+
+      expect((screen.getByLabelText(/^title$/i) as HTMLInputElement).value).toBe(
+        "Monthly Pods Report — June 2026"
+      );
+    });
+
+    it("should_seed_title_input_empty_when_report_is_untitled", async () => {
+      const user = userEvent.setup();
+      mockUsePortfolioReport.mockReturnValue({ data: baseReport, isLoading: false } as any);
+
+      render(<PortfolioReportEditorPage community={community} reportId="report-1" />);
+      await user.click(screen.getByRole("button", { name: /^edit$/i }));
+
+      expect((screen.getByLabelText(/^title$/i) as HTMLInputElement).value).toBe("");
+    });
+
+    it("should_send_the_typed_title_when_saved", async () => {
+      const user = userEvent.setup();
+      const updateMutate = vi.fn().mockResolvedValue(baseReport);
+      mockUseUpdateReportContent.mockReturnValue({
+        isPending: false,
+        mutateAsync: updateMutate,
+      } as any);
+      mockUsePortfolioReport.mockReturnValue({ data: baseReport, isLoading: false } as any);
+
+      render(<PortfolioReportEditorPage community={community} reportId="report-1" />);
+      await user.click(screen.getByRole("button", { name: /^edit$/i }));
+      await user.type(screen.getByLabelText(/^title$/i), "June 2026");
+      await user.click(screen.getByRole("button", { name: /^save$/i }));
+
+      expect(updateMutate).toHaveBeenCalledWith(
+        expect.objectContaining({ reportId: "report-1", title: "June 2026" })
+      );
+    });
+
+    it("should_send_null_title_when_the_field_is_emptied", async () => {
+      const user = userEvent.setup();
+      const updateMutate = vi.fn().mockResolvedValue(baseReport);
+      mockUseUpdateReportContent.mockReturnValue({
+        isPending: false,
+        mutateAsync: updateMutate,
+      } as any);
+      mockUsePortfolioReport.mockReturnValue({
+        data: { ...baseReport, title: "June 2026" },
+        isLoading: false,
+      } as any);
+
+      render(<PortfolioReportEditorPage community={community} reportId="report-1" />);
+      await user.click(screen.getByRole("button", { name: /^edit$/i }));
+      await user.clear(screen.getByLabelText(/^title$/i));
+      await user.click(screen.getByRole("button", { name: /^save$/i }));
+
+      expect(updateMutate).toHaveBeenCalledWith(expect.objectContaining({ title: null }));
+    });
+
+    it("should_enable_save_when_only_the_title_changed", async () => {
+      const user = userEvent.setup();
+      mockUsePortfolioReport.mockReturnValue({ data: baseReport, isLoading: false } as any);
+
+      render(<PortfolioReportEditorPage community={community} reportId="report-1" />);
+      await user.click(screen.getByRole("button", { name: /^edit$/i }));
+
+      expect(screen.getByRole("button", { name: /^save$/i })).toBeDisabled();
+
+      await user.type(screen.getByLabelText(/^title$/i), "June 2026");
+
+      expect(screen.getByRole("button", { name: /^save$/i })).toBeEnabled();
     });
   });
 });
--- a/__tests__/components/Pages/Admin/PortfolioReports/PortfolioReportEditorPage.test.tsx
+++ b/__tests__/components/Pages/Admin/PortfolioReports/PortfolioReportEditorPage.test.tsx
@@ -300,6 +300,25 @@ describe("PortfolioReportEditorPage", () => {
       expect(updateMutate).toHaveBeenCalledWith(expect.objectContaining({ title: null }));
     });
 
+    it("should_not_treat_a_stored_title_as_unsaved_edits_before_the_dialog_opens", async () => {
+      // The title draft is empty until the dialog seeds it, so comparing it to
+      // a stored title without gating on the dialog being open would report
+      // unsaved edits on every titled report and wrongly warn here.
+      const user = userEvent.setup();
+      mockUsePortfolioReport.mockReturnValue({
+        data: { ...baseReport, title: "Monthly Pods Report — June 2026" },
+        isLoading: false,
+      } as any);
+
+      render(<PortfolioReportEditorPage community={community} reportId="report-1" />);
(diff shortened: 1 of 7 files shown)


brunodmsi:
Thanks — all four findings were valid and are fixed in 22614a4d. Notes on each:

**🔴 Title-only saves overwriting refreshed content** — real, and the sharpest catch. The dirty checks compared drafts against live `report` data, so a body refreshing under an open dialog would get clobbered by the stale seeded copy on a title-only save. The dialog now snapshots content+title on open and compares against that, sending the draft only when the body was genuinely edited and `title: undefined` when the title wasn't — which is what the endpoint's tri-state exists for. Regression test added (`should_not_overwrite_refreshed_content_when_only_the_title_changed`), plus the reverse case (content-only save leaves the title untouched).

One thing the suggested fix would have broken, so I handled it separately: gating on "content edited" naively means clearing the textarea deliberately reads as *not* edited, silently restoring the body. Baselines fix that too — pinned by `should_send_emptied_content_rather_than_silently_restoring_it`.

**🟠 The stale-param test didn't test what it claimed** — correct, and an embarrassing one. `config-propgf` has two reports, so it only re-tested narrowing; the empty state and reset button were never exercised. The nuqs mock now seeds the query param, so an unmatched `?type=` is tested for real, split into three tests (empty state renders / no zero count / reset restores).

**🟡 Timeline reset on data refresh** — right, and self-inflicted: the `setActiveKey(null)` I added fired on every refetch because `filteredReports` gets a new identity each time. Now keyed on `typeFilter`, the only thing that can actually invalidate the dot. I skipped the membership-key suggestion — re-observing on refetch is harmless since seeding preserves any dot already set, and keying the observer effect on a derived string needed an `exhaustive-deps` disable that tripped the quality gate. Test added for the filter surviving a refetch.

**🟡 Zero count** — right, and a direct violation of our documented empty-state rule. Hidden when nothing matches. It was worse than cosmetic: that block also read `latest.runDate` with no rows to read from.

Quality gate should go green too — the +1 react-doctor warning traced to the `setActiveKey` effect; branch now scans at 3336 vs the 3337 baseline (verified locally against the pinned 0.2.2).

brunodmsi:
## Babysit sign-off

**Self-review score:** 98/100.

**Bot threads resolved:** 4/4 (all CodeRabbit; fixed in 22614a4d, replies inline).

**CI:** 31/31 green — including `quality-gate`, which had gone red on a single +1 react-doctor warning and is now a net improvement (3336 vs 3337 baseline).

### Found during self-review, beyond the bot findings

The title input's placeholder fell back through `report.reportConfigName` — but this page loads the report from the **admin** endpoint, which maps via `toReportResponse`; only the published-list mapper projects the config name. That branch could never be taken. Replaced with a static example (ea73619b). Dirty-check derivations also hoisted above the save handler that reads them.

Remaining known deduction: `PortfolioReportEditorPage` trips Biome's complexity rule (39 vs limit 20). It was already over on `main` (34) before this PR; extracting the dialog felt like scope creep here, happy to split it into a follow-up.

---

### ⚠️ Conflicts with #1855 — merge order matters

#1855 (`fix/report-same-date-collision`) touches four of the same files. A `git merge-tree` probe against this branch reports genuine conflicts:

- `components/Pages/Community/PortfolioReports/PublicReportListPage.tsx` — **content conflict**. This PR moves the card into a memoized `ReportListItem`; #1855 rewrites the detail link in that same block.
- `__tests__/components/Pages/Community/PortfolioReports/PublicReportListPage.test.tsx` — **add/add conflict**. Both PRs create this file from scratch (neither existed before).

There's a semantic conflict too, not just textual: #1855 adds a `[configSlug]` segment to `PAGES.COMMUNITY.REPORT_DETAIL`, which `ReportListItem` calls. Whichever lands second needs a real resolution, not an `ours`/`theirs` pick — the second one should rebase and re-run both suites.

Worth noting the two are complementary: #1855 makes same-date reports individually reachable, and this PR makes them individually *identifiable*. Filecoin needs both.

brunodmsi:
## Rebased onto #1855 — conflicts resolved

This PR is now **stacked on `fix/report-same-date-collision` (#1855)** rather than `main`, and the base is retargeted so the diff shows only this branch's 7 files. GitHub will re-point it at `main` automatically when #1855 merges.

### How the two conflicts were resolved

**`PublicReportListPage.tsx` (content).** Both PRs rewrote the same card block: this one extracts it into a memoized `ReportListItem`; #1855 changes the link and the scrub key. Kept this branch's structure and folded #1855's semantics into the component:

- `data-report-key` → `report.id` (was `runDate` — ambiguous once two configs publish on one day)
- link → `REPORT_DETAIL(slug, report.runDate, report.reportConfigSlug)`
- `deriveTimeline`'s `key: r.id` auto-merged and was kept

**`PublicReportListPage.test.tsx` (add/add).** Neither file existed on `main`; both PRs created one. Merged rather than picked a side — this branch's file is the base (it carries the `nuqs` and Radix `Select` mocks the component now needs), plus #1855's three same-date tests. Their loading/empty/error tests were dropped as exact duplicates of the ones already here.

### Two things worth flagging, since a naive resolution would have missed them

1. **#1855's tests could not survive unmodified.** They use two distinct configs, which makes this branch's type dropdown render — but they mock neither `nuqs` nor Radix `Select`. Ported onto this file's mocks.
2. **`renders both same-date reports` genuinely broke.** `getByText("Quarterly Review")` started matching two nodes, because each config name now also appears as an `<option>` in the type filter. Scoped it to `getByRole("heading", ...)` — the assertion was ambiguous, not wrong.

Also added a test for the interaction the stack creates: a titled report on a same-date pair still gets a distinct link. A title changes what a card *reads*, never where it *points*.

### Verification

- **113 tests green across both PRs' suites** (9 files — this branch's, #1855's, `pages.test.ts`, and the shared utilities).
- `tsc --noEmit` clean, Biome clean on changed files, `pnpm run build` compiles.
- The unrelated `fix(sitemap)` commit that an earlier `main` merge had dragged in was dropped, so the stack is exactly this feature's 5 commits.

Force-pushed with `--force-with-lease` (`ea73619b` → `075e83b7`).

brunodmsi:
## Babysit sign-off — post-rebase

**Self-review score:** 98/100 (re-scored against the rebased tree, base `fix/report-same-date-collision`).

**Bot threads:** 0 unresolved. CodeRabbit re-reviewed the rebased diff after the force-push (walkthrough updated 00:31, head `075e83b7`) and posted no new findings — all 5 pre-merge checks passed. The 4 earlier findings stay resolved, 3 auto-marked outdated by the fixes.

**CI:** 32/32 green, 0 failing.

### Verified on the merged result, not just the parts

The rebase is the risk here, so the check that matters is both features' suites passing together: **113 tests green across 9 files** — this branch's, #1855's, `pages.test.ts`, and the shared utilities. `tsc --noEmit` clean, Biome clean on changed files, `pnpm run build` compiles.

Semantics of the merge, re-confirmed by reading the result rather than trusting the resolution:

- `deriveTimeline` → `key: r.id`, `ReportListItem` → `data-report-key={report.id}`, and `handleJumpTo` queries by that same id. All three agree, so the scrubber can't jump to the wrong report when two configs publish on one day.
- The link carries `reportConfigSlug` through the memoized component.
- The type filter resets the active dot on `typeFilter` only — a refetch no longer disturbs it.

### One design question the stack creates (not fixed here)

`?type=` keys off `reportConfigId` (an opaque ObjectId), while #1855 introduces `reportConfigSlug` — a human-readable identifier added specifically for URLs. `?type=monthly-pods-report` would be a nicer shareable link than `?type=6a23268272df01209256e5b9`, and shareability is the reason this filter is URL-synced at all.

Left as-is deliberately: `reportConfigSlug` is nullable (deleted config), and keying on it would make those reports unfilterable, whereas the id is always present. Happy to switch to `slug ?? id` if the readable URL is worth that wrinkle — it's a one-expression change in `deriveReportTypes` + the filter predicate.

### Known, unchanged

`PortfolioReportEditorPage` trips Biome's complexity rule (39 vs limit 20). Pre-existing — it was already 34 on `main` before this branch.

**Merge order:** this PR is stacked and cannot merge until #1855 does; GitHub will re-point it at `main` automatically.

PR #1857, opened 2026-07-16
feat(portfolio-reports): derive type filter from configs, not reports

brunodmsi:
Pairs with **show-karma/gap-indexer#2202** (adds the public `/reports/types` endpoint this consumes). Follow-up to the DEV-520 type filter.

## Problem

The report type dropdown derived its options from the **published reports** on the page. So a report config that hadn't published a report yet never appeared — you had to publish before you could filter by that type. Reported on staging: a second config was added to Filecoin, but the dropdown stayed hidden because that config had no published report.

A report type exists the moment its config does. The filter should reflect that.

## Change

Wires the dropdown to the new public `GET /communities/:slug/reports/types` endpoint (gap-indexer#2202), which returns one entry per config that is **active or has a published report** — so a newly-created type is selectable immediately. Picking a type with no reports shows the existing "No reports of this type" empty state.

The filter predicate is unchanged (`report.reportConfigId === selected`); only the *source* of the options moved from reports to configs.

**Graceful fallback:** if the endpoint is unavailable (e.g. this ships ahead of the backend), the hook has no data and the dropdown falls back to deriving types from the published reports — today's behaviour. So this is deploy-order-safe and can merge before gap-indexer#2202.

## Why a new endpoint (not the existing configs endpoint)

`GET /report-configs` is admin-gated (401 for anon) and its response carries the LLM `prompt`. The public list page has no auth and must not leak that. gap-indexer#2202 adds a redacted, public `{ id, name, slug }` endpoint; archived configs with no reports are omitted server-side so their names never become public.

## Testing

- 26 tests in `PublicReportListPage.test.tsx` (4 new): a report-less type still lists; selecting it shows the empty state; endpoint types win over report-derived; fallback to report-derived when the endpoint has no data.
- Mutation-checked: forcing the component to ignore the endpoint fails exactly the 3 endpoint-path tests, nothing else.
- 67 portfolio tests green; `tsc` clean; Biome clean; `pnpm run build` compiles.

## Smoke results

Deferred to the shared preview — the dropdown lights up with report-less configs only once gap-indexer#2202 is deployed. Until then this renders identically to current `main` (fallback path). The reporter's staging case (2 configs, one report-less) is the exact scenario to verify once both are on the same environment.

## Summary by CodeRabbit

- **New Features**
  - Added support for loading available public report types for the Type filter.
  - Type options now include available types even when no reports exist for them.
  - Selecting a type without reports displays a clear empty state.

- **Bug Fixes**
  - Improved filter behavior by prioritizing available report types and falling back to types found in published reports when necessary.

### DIFF
--- a/__tests__/components/Pages/Community/PortfolioReports/PublicReportListPage.test.tsx
+++ b/__tests__/components/Pages/Community/PortfolioReports/PublicReportListPage.test.tsx
@@ -2,7 +2,10 @@ import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { useState } from "react";
 import { PublicReportListPage } from "@/components/Pages/Community/PortfolioReports/PublicReportListPage";
-import { usePublishedReports } from "@/hooks/portfolio-reports/usePortfolioReports";
+import {
+  usePublishedReports,
+  usePublishedReportTypes,
+} from "@/hooks/portfolio-reports/usePortfolioReports";
 
 // `initialType` seeds the query param, standing in for a URL arriving with
 // `?type=` already set (a shared or reloaded link).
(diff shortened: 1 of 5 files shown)


brunodmsi:
Superseded — after clarifying the intent, the public list should stay published-only (which `main` already is), and the draft-or-published filter belongs on the admin report list instead. No new endpoint needed; the admin page derives types from the reports it already loads. Closing in favor of the admin-side change.

PR #1864, opened 2026-07-17
feat(nonprofit-research): redesign advisor pages with persistent shell

amaurymagalhaes:
## Overview

Redesign of the nonprofit-research advisor pages, plus a rework of the section's loading and navigation architecture.

## What changed

**Navigation shell**
- The collapsible navigation shell now lives in the section `layout.tsx` instead of being re-wrapped by every page. The sidebar, header, and breadcrumbs persist across route transitions — only the content pane swaps.
- Pages render just their content; `allowStaffViewer` (staff reading a report without an advisor profile) is derived from the pathname inside the shell.

**Loading states**
- The shell renders its real chrome immediately (the nav is static and doesn't need the advisor), then skeletons only the content pane. This removes the previous fake-sidebar placeholder that swapped for the real sidebar on resolve (layout shift).
- The pre-shell loader that collapsed to a ~64px sliver (a `max-w-*` box with no `w-full` inside a flex row) is gone: the session boundary now renders the shell straight through Privy initialization, so there is one consistent loading state from first paint.
- `DonorResearchLoading` gained route-shaped variants (home / list / report / form) so the skeleton mirrors the destination and resolves in place.

**Session boundary**
- Per-account React Query cache isolation now pauses the subtree only on a genuine account switch (not on every mount), keyed to the QueryClient's lifetime.

**Redesign**
- Restyle across report brief/viewer, shared-view, criteria-input, personas, and onboarding; list-first section home.

## Testing
- `pnpm test` (donor-research suites): 351/351 passing
- `pnpm lint:fix` clean, `tsc --noEmit` clean
- Loading behavior verified in a headless browser: real sidebar renders at first paint, content skeleton fills the pane, no collapsing sliver.


amaurymagalhaes:
## Babysit summary

Self-reviewed the diff and worked the review/QA loop to a healthy state. All deterministic checks are green (build, all test shards, smoke, static-checks, checklist, react-doctor, **quality-gate**, dogfood, qa-execution). No open reviewer threads.

**Fixes applied on top of the redesign commit:**

1. **quality-gate ratchet** — cleared the React Doctor regression (3337 → 3326). Replaced mid-sentence em dashes in UI copy with sentence punctuation, switched display headings from `font-bold` to `font-semibold`, hoisted a currency `Intl.NumberFormat` to module scope, collapsed matching `px/py` paddings, and formatted the share-link expiry in a post-mount effect to avoid a hydration mismatch.
2. **Create-report feedback** — an invalid submit or a rejected create (e.g. a 429 daily-limit) could read as a silent no-op because the Create button sits in a sticky footer away from the inline error. Now both raise a toast in addition to the inline message, and the create rejection is caught so it never escapes as an unhandled promise. Covered with tests.
3. **Token-route chrome flash** — the anonymous share/diligence token routes briefly rendered the advisor sidebar + skeleton before resolving to their slim `TokenPageShell`. They now render shell-less like onboarding, keyed on the existing `isDonorResearchTokenRoute` helper.
4. **Discard guard coverage** — the unsaved-changes guard on both the persona-detail page and the quick-create persona modal only tracked the persona editor, so editing the private Notes field and navigating away/closing dropped the edit with no confirmation. Wired the notes section's dirty state into both guards. Covered with tests.

**Remaining `verdict`/`qa-pipeline` findings (not addressed here):** the logged-out auth boundary — visiting an advisor route while signed out renders the nav chrome / throws to the error boundary instead of showing a sign-in wall. This posture is **pre-existing** (the pre-redesign `DonorResearchHome` threw the same 401-to-error-boundary and never gated logged-out access); the redesign only makes the nav chrome more visible during the brief unauthenticated window. There is no data exposure — the section's data queries 401 without a JWT. Adding a proper auth wall is a product/UX decision (destination, copy) with no standard reusable pattern in the app, so it's left for a follow-up rather than folded into this redesign. `qa-pipeline` is informational (not a required status check).

PR #1884, opened 2026-07-21
feat(assistant): move Karma Assistant from floating bubble to navbar

amaurymagalhaes:
## Overview

The Karma Assistant was only reachable through a floating bubble pinned to the bottom-right of every page. This promotes it to a top-level control in the navbar, so it sits alongside the other primary tools instead of hovering over page content.

| | Before | After |
|---|---|---|
| Entry point | Floating bubble, bottom-right | "Ask Karma" button in the navbar (icon-only on mobile) |
| Panel position | Floats above the bubble, fixed 600px tall | Drops from under the navbar, spans to the bottom of the viewport |
| Mark | Sparkles | Karma logo |

## Why not just delete the floating button

`ChatBubbleShell` is shared with the **embeddable widget** (`widget/ChatWidget.tsx`) that runs on customers' own sites, where there is no host chrome to hang a trigger from. So the floating placement is still needed — it became a `placement` prop instead of being removed. The widget keeps `"fab"`; only the in-app panel opts into `"anchored"`.

The same reasoning applies to the brand mark: `renderBrandIcon` defaults to the previous sparkles icon so customer-hosted embeds are not silently rebranded with the Karma logo.

## Why the trigger only toggles a store flag

Three entry points already opened the assistant besides the bubble:

- `components/Community/Header.tsx` — community context + ⌘K shortcut
- `components/Pages/Admin/MilestonesReview/index.tsx` and `MilestoneCard.tsx` — seed an `@mention` chip for a milestone

All three call `setOpen(true)` on the shared Zustand store. The navbar button does the same, so all of them keep working with no changes. This is also why the panel stays an overlay rather than becoming a link to `/ask-karma`: the milestone flows depend on staying on the current page.

## Accessibility

The floating button had none of this; the replacement adds it:

- `aria-expanded` / `aria-controls` pointing at the panel
- Escape closes the panel
- Focus moves into the composer on open, and returns to the trigger on close (only when focus was still inside the panel)

Escape and focus handling are scoped to the anchored placement so the widget does not start swallowing Escape on customer pages.

## Testing steps

1. Load any page — the assistant is in the navbar, not floating bottom-right.
2. Click **Ask Karma** — the panel drops from under the navbar and runs to the bottom of the screen; focus lands in the composer.
3. Press **Escape** — the panel closes and focus returns to the button.
4. Resize to mobile — the trigger is an icon-only button next to the hamburger, *not* inside the drawer; the panel fits without horizontal overflow.
5. Go to a community page and press **⌘K / Ctrl+K** — the panel still opens with community context.
6. Open a milestone in Admin → Milestones Review and use the `@mention` button — the panel still opens with the chip seeded.
7. Go to `/ask-karma` — no navbar trigger and no floating bubble (the page is itself a full assistant surface).

## Verification

Checked in a real browser against a local dev server at four viewport heights (700 / 900 / 1200 / mobile 844). The panel starts at y=72 under the 64px navbar and ends 16px from the bottom at every size; the mobile panel fits within the viewport (x=16, width=358 at 390px wide).

- `tsc --noEmit` clean
- `biome check` clean on all changed files
- 109 tests pass across the assistant, navbar, and widget suites

Three tests in `AgentChatBubble.test.tsx` asserted the floating button existed inside that component; they now assert the inverse (panel renders, trigger does not). New `__tests__/navbar/unit/navbar-assistant-button.test.tsx` covers toggling, ARIA state, compact mode, and route suppression.

## Notes for review

- The anchored panel derives its height from `top` + `bottom` insets. Adding an `h-*` class back to that branch would break the bottom anchoring — there is a comment saying so.
- The button uses `brand-blue` (#4C6FFF) to stay visually linked to the panel it opens. Worth flagging that `tailwind.config.js` marks this a legacy colour and the current brand is teal `#2ed1a8`; recolouring would mean touching the panel header and empty state too, so it is deliberately out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Added “Ask Karma” controls to desktop, mobile, and whitelabel navigation.
  * Added anchored assistant panels that appear beneath the navbar, focus the message composer, and close with Escape when appropriate.
  * Added Karma branding within the assistant interface.
* **Accessibility**
  * Improved assistant trigger labels, expanded-state announcements, panel targeting, and focus restoration.
* **Bug Fixes**
  * Removed duplicate floating chat controls where navigation triggers are available.
  * Preserved assistant visibility and behavior across supported routes.


amaurymagalhaes:
@coderabbitai review

amaurymagalhaes:
Addressed in `c2375cd`.

The finding was valid. Focus restoration did work in Chrome, but only because the effect happened to run before the browser processed the `inert`-induced blur — timing rather than a guarantee, and not something to rely on across browsers.

Rather than capturing state before the toggle, the panel-focus state is now tracked continuously while the panel is open (driven by `focusin`) and read from that record on close. That removes the dependency on post-close `activeElement` entirely.

Verified in a browser: Escape with focus in the composer closes the panel and returns focus to the trigger; Escape with focus outside the panel is ignored, so dismissing an overlaid dialog no longer closes the assistant underneath it.

amaurymagalhaes:
QA finding A4 (High) — *Focus-restore only targets the visible navbar instance* — addressed in `500594e`.

The reported symptom was correct and the root cause was this PR's, though not what the finding's title suggests. It was not a focus-restore defect; the panel simply could not be opened because the trigger had been pushed off screen.

**Measured at a 1024px viewport:** the desktop nav row is exactly `1024px` wide *without* the trigger — zero slack — and `1075px` with it. The row is fixed-position and clips silently with no scrollbar, so the overflow carried the trigger, the theme toggle and the account menu off screen where they were not hit-testable.

**Fix:** the trigger renders icon-only from `lg` to `xl`, and regains its label at `xl` where there is room. `aria-label` is now set at every width instead of only in compact mode, so hiding the text costs assistive tech nothing.

**Verified** at 1024 / 1100 / 1279 / 1280 / 1440 / 1920: no row or document overflow at any width, the trigger is hit-testable at its centre point, and its accessible name stays "Ask Karma" throughout.

### Pre-existing issue this uncovered

Because the row has **zero slack at 1024px**, the authenticated navbar — which adds a Dashboard button and a wider account menu — already overflowed at that width before this PR. QA measured that row at `1159px`. This change removes its own `51px` contribution, but roughly `46px` of pre-existing overflow remains for signed-in users at exactly 1024px. That is a separate layout bug and is not fixed here.

PR #1939, opened 2026-07-28
fix: bound the permissions fetch so a stalled RBAC lookup can't brick the report page

amaurymagalhaes:
## Summary

A `GET /v2/auth/permissions` request that never completes bricks the nonprofit-research report page **permanently**: an infinite `Loading report…` skeleton, no timeout, no error state, no retry affordance. The user sits there indefinitely.

This is pre-existing on `main` — it was found while investigating a red QA pipeline on #1936, and the files involved are byte-identical between `main` and that branch. This PR makes an unresolvable permissions fetch **fail closed and visibly** instead of hanging.

## Diagnosis chain

1. `src/features/donor-research/components/report-brief/ReportBriefView.tsx` gates the page on `const isManageAuthPending = isStaffLoading || advisorQuery.isPending;` and returns the skeleton while it's true.
   - The stall was localised to *this* component rather than the route: the route-level `loading.tsx` renders a different string (`Loading research report…`), and the stuck page shows `Loading report…`.
2. `advisorQuery` always settles — `fetchCurrentAdvisor` maps 404 → `null` and otherwise throws. So the only input that can hang forever is `useStaff().isLoading`.
3. That traces to `PermissionProvider`'s `awaitingPermissions` guard (`believedAuthenticated && !data && !isError`), which stays `true` for as long as the permissions React Query stays pending.
4. The query had no meaningful bound. `utilities/api/client.ts` gives every default-baseURL (indexer) request a **360s** ceiling — `DEFAULT_INDEXER_TIMEOUT_MS`, a deliberate allowance for long-poll endpoints (AI evaluation, bulk jobs, report generation). A cheap authorization lookup inherited it by default, and with React Query retries on top the worst case ran into the tens of minutes.

### 1. A deadline on the permissions fetch (`src/core/rbac/services/authorization.service.ts`)

Each fetch races a 15s wall-clock deadline (`PERMISSIONS_TIMEOUT_MS`) and aborts the in-flight request via `AbortSignal` when the deadline wins.

The deadline is applied **in the service, not left to the transport**, because `api.get` awaits `TokenManager.getToken()` *before* it issues the XHR. A stalled Privy token bootstrap would never reach axios' timer, and the promise React Query is waiting on would still never settle. Racing the whole call is the only thing that guarantees the query settles regardless of which stage stalls — and the loading gate's termination depends on that guarantee, not on the transport's.

The transport is given a **longer** bound than the deadline (`PERMISSIONS_TRANSPORT_TIMEOUT_MS = PERMISSIONS_TIMEOUT_MS + 5_000`) so the two timers cannot tie. This matters more than it looks: with both set to 15s they fire in the same tick, `Promise.race` picks a winner arbitrarily, and the winner decides the error *type* — which decides whether the query retries. QA caught exactly that (see the QA response comment); the fix makes our deadline deterministically first and normalises any post-deadline rejection to `PermissionsTimeoutError`.

### 2. Not broadened to the shared api client — argued both ways

**For broadening:** one place to fix, and no future authenticated GET can accidentally inherit the long-poll ceiling.

**Against (chosen):** the 360s default is not an oversight — the comment in `client.ts` records it as a deliberate migration-parity decision so the ~220 legacy `fetchData` call sites don't have to pass `timeoutMs` individually. Some of them genuinely long-poll. Lowering the default would silently break those, and the failure mode would be a spurious timeout on a *working* endpoint — worse and harder to spot than the bug being fixed. This PR touches shared RBAC-gated loading code on a hot path; the right blast radius here is the one endpoint proven to hang. If more endpoints turn out to need it, the pattern generalises cleanly.

### 3. Timeouts are not retried (`src/core/rbac/hooks/use-permissions.ts`)

`retry` already exempted 429s; timeouts now join them. Retrying keeps the query pending, and every RBAC-gated surface holds its skeleton for the entire run — 3 × 15s plus backoff is another ~48s of an apparently-frozen page. Failing fast hands the user an explicit, immediate Retry control instead, which is both faster and visible. The endpoint answers in ~100ms when healthy; if it hasn't answered in 15s, two more 15s waits are not the remedy.

### 4. `useStaff()` is properly tri-state (`src/core/rbac/hooks/use-staff-bridge.ts`)

Now returns `{ isStaff, isLoading, isError }`. Per the repo's tri-state authorization rule, staff is granted only on a **resolved, successful** answer:

```ts
isStaff: !isLoading && !isGuestDueToError && hasRoleOrHigher(Role.SUPER_ADMIN),
```

The `!isGuestDueToError` term is load-bearing beyond the first-load case: React Query retains the last successful payload when a *refetch* fails, so without it an unresolvable refetch would keep granting privileges off a cached answer. Fail closed.

### 5. A visible outcome in `ReportBriefView`

A terminal error state with a Retry button that refetches the permissions query, replacing what used to be an eternal skeleton.

**Confirmed owners are exempt.** `canManageReport = isOwner || isStaff`, and `isOwner` comes from the advisor row, not from RBAC — so a failed staff lookup doesn't change what an owner may do, and blocking them would be a new availability regression introduced by the fix. Only a viewer we cannot confirm as the owner is actually blocked.

**Trade-off, stated plainly:** a non-owner non-staff viewer now sees an error where they previously saw the read-only variant. We cannot distinguish "not staff" from "staff status unknown", and the repo rule is explicit that an unresolved authorization state must not silently render as a decided one. Failing closed *and visibly* is the correct side of that trade — silently degrading a real super-admin to read-only with no explanation is exactly the failure this rule exists to prevent.

## Tests — mutation-verified

Three new/extended suites, all driving the **real** chain (`PermissionProvider` → `usePermissionsQuery` → `authorizationService` → api client) against an `api.get` that never settles, under fake timers. Mocking the query or the service out cannot observe whether the fetch is actually bounded, so it would not have caught this.

- `src/core/rbac/__tests__/permissions-timeout.test.tsx` — the gate terminates; a timed-out result does not grant staff; the request is not re-issued.
- `src/features/donor-research/components/report-brief/__tests__/permissions-stall.test.tsx` — the report page escapes the skeleton and surfaces an alert + retry; the in-flight request is aborted; a confirmed owner is not blocked.
- `src/core/rbac/__tests__/use-staff-bridge.test.ts` — the stale-data guard.

Note that "does not grant staff" is asserted **together with** `isLoading === false`. Asserted alone it would pass against the buggy code, which also never granted staff — it simply never answered. The two assertions are only a guard in combination.

### Mutation results

Each part of the fix was reverted individually and the suites re-run:

| Mutation | Result |
|---|---|
| Remove the deadline race (restore the 360s default) | **5 of 6 tests fail** — skeleton persists, signal never aborts, gate never terminates, staff never resolves |
| Drop `!isGuestDueToError` from `isStaff` | **1 fails** — stale-data guard |
| Remove the `isStaffError && !isOwner` branch in `ReportBriefView` | **1 fails** — no alert, no retry |
| Allow retrying timeouts | **1 fails** — 3 requests instead of 1 |
| Restore the deadline/transport timer tie (drop the headroom + normalisation) | **2 fail** — transport bound no longer strictly greater; a transport-first rejection escapes as a retryable error |

Fix restored: **43 files / 540 tests pass** across `src/core/rbac`, `__tests__/core/rbac`, `src/features/donor-research`, `__tests__/integration/rbac`. `npx tsc --noEmit` clean. `npx biome check` clean on all touched files (the two remaining warnings — an unused `DEFAULT_GUEST_PERMISSIONS` and an unused import — were verified to pre-date this branch on `main` and are left alone).

## What this does NOT explain

**The endpoint responds in 81–91ms against staging via curl, yet QA observed the XHR firing 8+ times in-browser without ever completing. That discrepancy is unexplained.** No root cause has been established for why the in-browser request stalls when the same request from curl is fast.

This PR makes that behaviour **survivable**, not fixed. It converts an unbounded, invisible, unrecoverable hang into a bounded, visible, retryable failure — and it does so without depending on which stage stalls, which is why the deadline wraps the whole call rather than just the transport. If the underlying stall is a Privy token bootstrap deadlock, a connection-pool exhaustion, or something in the proxy layer, this change does not address it and the timeouts will still fire. It should be treated as a floor under the user experience, not a closed investigation. `errorManager` still reports every failure, so the timeouts will now be visible in Sentry rather than silently swallowed by an eternal skeleton — which should give the root-cause work something to work from.

## Bot findings

- **`EMPTY_CATCH` at `authorization.service.ts` (anti-pattern check)** — fixed in `7a59642`. It was a false positive: the checker peeks only a few lines into a catch body and stops at the first `}`, so the multi-line `errorManager(...)` call hid the `throw error` behind it. Rather than annotate around it, the shared reporter is now extracted (`reportPermissionsFailure`), which deduplicates the identical payload built by both failure paths and leaves the report-and-rethrow readable in two lines. Check now passes.
- **Barrel imports at `use-staff-bridge.ts:4` and `authorization.service.ts:5` (React Doctor)** — pre-existing on `main`, surfaced only because these files appear in the diff. `import … from "../types"` on both lines is unchanged by this PR. Per `gap-app-v2/CLAUDE.md` the existing `types/` barrels are legacy and shouldn't be *added to*; unwinding them is a separate change and not something to bundle into an RBAC hot-path fix. The React Doctor check itself passes.

## Summary by CodeRabbit

* **New Features**
  * Added timeout handling for permission checks so requests no longer remain stuck indefinitely.
  * Added clear error states when authorization cannot be determined.
  * Report owners can continue accessing report management features during permission service failures.

* **Bug Fixes**
  * Prevented unauthorized access when permission checks fail or time out.
  * Stopped retrying permission requests after timeout failures.
  * Added a retry option when report authorization cannot be resolved.

### DIFF
--- a/__tests__/core/rbac/services/authorization.service.test.ts
+++ b/__tests__/core/rbac/services/authorization.service.test.ts
@@ -1,5 +1,8 @@
 import { errorManager } from "@/components/Utilities/errorManager";
-import { authorizationService } from "@/src/core/rbac/services/authorization.service";
+import {
+  authorizationService,
+  PERMISSIONS_TIMEOUT_MS,
+} from "@/src/core/rbac/services/authorization.service";
 import { Permission, ReviewerType, Role } from "@/src/core/rbac/types";
 import { api } from "@/utilities/api/client";
 import { INDEXER } from "@/utilities/indexer";
@@ -10,6 +13,12 @@ vi.mock("@/utilities/api/client", () => ({
 vi.mock("@/components/Utilities/errorManager");
 
 const mockedApiGet = vi.mocked(api.get);
+
+/** See `PERMISSIONS_TIMEOUT_MS` — the lookup is always bounded and abortable. */
+const BOUNDED_REQUEST = {
+  signal: expect.any(AbortSignal),
+  timeoutMs: PERMISSIONS_TIMEOUT_MS,
+};
 const mockedErrorManager = vi.mocked(errorManager);
 
 interface ApiResponseOverrides {
@@ -105,7 +114,10 @@ describe("authorizationService.getPermissions", () => {
 
       await authorizationService.getPermissions(params);
 
-      expect(mockedApiGet).toHaveBeenCalledWith(INDEXER.V2.AUTH.PERMISSIONS(params));
+      expect(mockedApiGet).toHaveBeenCalledWith(
+        INDEXER.V2.AUTH.PERMISSIONS(params),
+        BOUNDED_REQUEST
+      );
     });
 
     it("defaults to empty params when called with no arguments", async () => {
@@ -113,7 +125,7 @@ describe("authorizationService.getPermissions", () => {
 
       await authorizationService.getPermissions();
 
-      expect(mockedApiGet).toHaveBeenCalledWith(INDEXER.V2.AUTH.PERMISSIONS({}));
+      expect(mockedApiGet).toHaveBeenCalledWith(INDEXER.V2.AUTH.PERMISSIONS({}), BOUNDED_REQUEST);
     });
 
     it("does not invoke errorManager on a successful response", async () => {
--- a/__tests__/core/rbac/services/authorization.service.test.ts
+++ b/__tests__/core/rbac/services/authorization.service.test.ts
@@ -1,7 +1,7 @@
 import { errorManager } from "@/components/Utilities/errorManager";
 import {
   authorizationService,
-  PERMISSIONS_TIMEOUT_MS,
+  PERMISSIONS_TRANSPORT_TIMEOUT_MS,
 } from "@/src/core/rbac/services/authorization.service";
 import { Permission, ReviewerType, Role } from "@/src/core/rbac/types";
 import { api } from "@/utilities/api/client";
@@ -17,7 +17,7 @@ const mockedApiGet = vi.mocked(api.get);
 /** See `PERMISSIONS_TIMEOUT_MS` — the lookup is always bounded and abortable. */
 const BOUNDED_REQUEST = {
   signal: expect.any(AbortSignal),
-  timeoutMs: PERMISSIONS_TIMEOUT_MS,
+  timeoutMs: PERMISSIONS_TRANSPORT_TIMEOUT_MS,
 };
 const mockedErrorManager = vi.mocked(errorManager);
 
--- a/src/core/rbac/__tests__/authorization.service.test.ts
+++ b/src/core/rbac/__tests__/authorization.service.test.ts
@@ -1,4 +1,4 @@
-import { authorizationService } from "../services/authorization.service";
+import { authorizationService, PERMISSIONS_TIMEOUT_MS } from "../services/authorization.service";
 import { ReviewerType, Role } from "../types/role";
 
 vi.mock("@/utilities/api/client", () => ({
@@ -8,6 +8,17 @@ vi.mock("@/utilities/api/client", () => ({
 import { api } from "@/utilities/api/client";
 
 const mockApiGet = api.get as vi.MockedFunction<typeof api.get>;
+
+/**
+ * Every permissions fetch is issued under a wall-clock deadline with an abort
+ * signal — an authorization lookup must never inherit the api client's 360s
+ * long-poll ceiling. Asserted on each call so the bound can't be dropped
+ * silently.
+ */
+const BOUNDED_REQUEST = {
+  signal: expect.any(AbortSignal),
+  timeoutMs: PERMISSIONS_TIMEOUT_MS,
+};
 
 describe("authorizationService", () => {
   beforeEach(() => {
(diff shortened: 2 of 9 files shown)


amaurymagalhaes:
@coderabbitai review

amaurymagalhaes:
## Response to QA verdict (1 critical — A3)

The A3 finding was worth chasing: it surfaced a **real bug in this PR**, now fixed in `13d331c`. It also over-states its conclusion in one respect. Both parts below.

### The real bug A3 found

A3 and A4 disagree, and the disagreement is the tell:

- **A4 PASSED** — permissions routed to `--abort` (fails *immediately*). Owner's report rendered with full management controls in ~18s.
- **A3 FAILED** — permissions made to *hang*. Page never reached a terminal state inside the expected window.

The difference between "fails immediately" and "hangs" is which timer fires. And I had armed **two timers on the same request with the same 15s bound**: my deadline in `authorization.service.ts`, and the `timeoutMs` I passed to the api client. They fire in the same tick, and `Promise.race` picks a winner arbitrarily.

When the transport won, the rejection was an ordinary retryable transport error — not `PermissionsTimeoutError` — so the retry exemption in `usePermissionsQuery` didn't match and the query **retried twice**. An intended 15s failure became ~48s of held skeleton. A3's "the client-side AbortSignal deadline does not appear to fire in this path" was reading a real defect, even if the mechanism was the retry loop rather than a dead timer.

Fixed by removing the tie rather than papering over it:

- The transport gets 5s of headroom (`PERMISSIONS_TRANSPORT_TIMEOUT_MS = PERMISSIONS_TIMEOUT_MS + 5_000`), so our deadline is deterministically first and owns the error type.
- Any rejection arriving *after* the deadline has fired is normalised to `PermissionsTimeoutError`, so the retry decision no longer depends on rejection ordering at all.
- Three new tests lock it: the transport bound is strictly greater than the deadline; the client is handed the transport bound; and a transport that rejects first with its own error still surfaces a timeout. Mutation-verified — restoring the tie fails 2 of them.

### Where A3 over-states

A3 was run against a report the test account **owns** (`Only a single Privy test identity was available … which owns every accessible report`), and then reported the absence of the "We couldn't verify your access" card as the failure. For an owner, that card is **intended** not to render: `canManageReport = isOwner || isStaff`, and `isOwner` comes from the advisor row, not from RBAC — so a failed staff lookup must not block them. That is exactly what A4 verified and passed.

So A3 asserted a non-owner expectation against an owner fixture. It is the same missing fixture that BLOCKED A1, A6, and A7 (no report owned by a *different* advisor, and no staff-role test account). The error-card path remains unverified end-to-end in this environment — I'm flagging that as an open gap rather than claiming it's covered.

### One more lead worth recording

A3 observed 5 held-open requests accumulating with no cancellation, and the page's *report* skeleton persisting — not just the RBAC gate. Held-open requests consume the browser's ~6-connection-per-origin budget, and once exhausted every other request to the same origin queues without starting. That would starve `useDonorReport`/`useDonorAdvisor` too, and it is a plausible mechanism for the original unexplained symptom (`/v2/auth/permissions` firing 8+ times in-browser without completing while curl answers in 81–91ms).

I have **not** verified this and it is not what this PR fixes. Aborting at the deadline releases the socket, which makes the failure mode survivable — but if connection starvation is the true root cause, it deserves its own investigation. Recording it as a lead, not a conclusion.

PR #1951, opened 2026-07-30
chore(verification): remove dead verification dialogs and useCanVerifyMilestone

brunodmsi:
## What this removes

Three verification dialogs and their shared permission hook, all of which are unreachable from any live surface:

| Deleted | Why |
|---|---|
| `components/Pages/GrantMilestonesAndUpdates/screens/MilestonesAndUpdates/VerifyMilestoneUpdateDialog.tsx` | Exported, zero importers |
| `components/Pages/GrantMilestonesAndUpdates/screens/MilestonesAndUpdates/VerifyGrantUpdateDialog.tsx` | Exported, zero importers |
| `components/Pages/Project/Impact/VerifyImpactDialog.tsx` | Only importer is `ImpactVerifications`, itself unimported |
| `components/Pages/Project/Impact/ImpactVerifications.tsx` | Exported, zero importers |
| `hooks/useCanVerifyMilestone.ts` | Consumed only by the three dialogs above |
| `__tests__/hooks/useCanVerifyMilestone.test.tsx` | Tests only the deleted hook |

Plus two cleanups the deletions exposed:

- `utilities/messages.ts` — dropped `GRANT.GRANT_UPDATE.VERIFY`, `MILESTONES.VERIFY`, and `PROJECT.IMPACT.VERIFY`. These three string blocks were referenced exclusively by the deleted dialogs; after the deletions a repo-wide `grep -rn "\.VERIFY"` over `*.ts*` returns **zero** hits.
- `__tests__/hooks/useIsReviewer.test.tsx` — removed the header note pointing at the now-deleted `useCanVerifyMilestone.test.tsx`.

## Why it matters beyond dead-weight

`useCanVerifyMilestone` gated on `useIsCommunityAdmin` alone — the legacy on-chain admin check that falls back to `false` when the RPC read fails. Every live admin surface has since moved to `useCommunityAdminAccess`, which unions legacy + RBAC + owner + super-admin. Leaving the hook in the tree meant any future import would silently re-introduce the RPC-false-negative gating class that has previously locked out Celo and Lisk admins. Deleting it removes the hazard; if a public verification surface is ever built, it should be wired to `useCommunityAdminAccess` from the start.

## Zero-importer evidence

Run at branch head, from the repo root, across `*.ts *.tsx *.js *.jsx *.mjs *.json *.md *.mdx *.yml *.yaml`, excluding `node_modules/` and `.next/`:

```
$ grep -rn "VerifyMilestoneUpdateDialog" .
[...]

Issue #1821, opened 2026-07-10
Attestation actions silently no-op when wagmi useAccount().address lags the Privy signer (Create Milestone: no wallet prompt, no error)


amaurymagalhaes:
## Summary

Clicking **Create Milestone** (and the same class of attestation actions across the app) can **silently do nothing** — no wallet-signature prompt, no error toast. Reported by a user on:
`https://www.karmahq.xyz/project/desci-asia/funding/0x917f7758145a025c5d1b01a3beefe13e2bb12c7da591559263f1737f243fc721/milestones-and-updates`

> "when i click create milestone there is nothing happen and no pop up for wallet sign"

## Root cause (verified in code)

`components/Dialogs/ProgressDialog/UnifiedMilestoneScreen.tsx`:

- `L96` — `const { address, chain } = useAccount();` (wagmi, **outer** config)
- `L133` / `L234` — `if (!address || !project …) return;` runs **before** `startAttestation()` and **before** the `try/catch`, so a falsy `address` produces zero feedback.
- The actual signature is produced from the **Privy** wallet via `useSetupChainAndWallet` → `useZeroDevSigner` (`smartWalletAddress` / `walletSigner`) — a *different identity source* than the guard.

The two identities diverge because the app runs **two `WagmiProvider`s** joined by an async bridge connector (`utilities/wagmi/privy-bridge-connector.ts`, `components/Utilities/PrivyWagmiProviders.tsx`). The outer config that `useAccount()` reads is populated only after the bridge connects; its own code notes: `// Non-fatal: outer useAccount() will lack address, but useAuth().address still works`. So there is a real authenticated state (Privy embedded wallet still syncing, or an external wallet whose site session dropped) where the user **can** sign but `useAccount().address` is `undefined` → silent `return`.

Consistent with a pre-feedback silent return, there are **zero Sentry issues** for this — nothing is ever thrown.

## Scope: systemic, not milestone-only

~38 attestation call sites gate on wagmi `useAccount().address` the same way (e.g. `Forms/ProjectUpdate`, `Forms/GrantUpdate`, `Impact/AddImpactScreen`, grant/objective completion + verify dialogs, many hooks). Milestone creation is just where this user hit it.

## Sibling bug (same file)

If `setupChainAndWallet` returns `null`, the per-chain loop `continue`s (`L289–295`), then still shows **"Milestones created! They may take a moment to appear."** (`L421–423`) despite zero attestations. False success.

[...]

PR #1952, opened 2026-07-30
fix(milestones): rebuild verification flow on V2 data and real signer identity

brunodmsi:
## Problem

`hooks/useMilestoneCompletionVerification.ts` is the site of production incident **GAP-FRONTEND-261** — "Failed to verify milestone" for community admins. Four separate defects live in that one flow:

1. **A whole-project refetch mid-flow (#63).** `fetchMilestoneInstance` called the SDK's `gapClient.fetch.projectById(...)` — a V1 `GET /projects/:uid` — and the fetched entity was consumed for essentially one scalar: `.recipient`. The rest was `.chainID` (a duplicate of `milestone.chainId`, already in scope), `.completed` (answerable from the V2 `completionDetails`), and a `communityUID` that was threaded through and never used. The attestation payloads never call a method on the fetched instance; they are freshly-built `MilestoneCompleted` objects. On top of that, the post-attest indexing poll re-ran `projectById` on **every** iteration, a second refetch after the reviewer-backend completion was fully redundant (`includeCompletion` was hard-set to `false` immediately after it), and `useMilestoneCancellation` repeated the same recipient-only pattern.
2. **One context-free toast for ~10 failure classes (#64).** Every non-abort, non-rejection failure collapsed into `showError("Failed to verify milestone")`, with a Sentry context of only `{ milestoneUID, address }`. Worse, `errorManager` early-returns for reject-ish and transient network/HTTP errors — so for those the user got the generic toast **and Sentry got nothing at all**. Plus a `console.error` in violation of the no-console rule.
3. **The poll matched the wrong address (#66).** The predicate compared `verified[].attester` to wagmi's `useAccount().address`, but the attestation is signed by the Privy-resolved signer (embedded wallet preferred). For a hybrid account (email login + a linked external wallet connected) those are deterministically different, so the poll could never match: ~200 retries × 1.5s ≈ 5 minutes, then a false failure — despite the transaction landing on-chain. Retries from there can mint duplicate attestations.
4. **A post-transaction throw on a null wagmi address (#67).** `throw new Error("User address not available")` executed **after** `GapContract.multiAttest` and after the indexer notification. A null wagmi address next to a working Privy signer is a real state (the outer-config sync is an async effect with a swallow-on-failure path), so the user could see a failure toast for an attestation that had already succeeded.

### #63 — no V1 project fetch anywhere in the flow

The V2 project-updates endpoint **already returns** `grantMilestones[].recipient` (`get-updates.api.response.ts` declares it; `project-updates.api.mapper.ts` maps it). The frontend mapper was dropping it.

- `services/milestones.ts`: `GrantMilestoneWithCompletion` gains `recipient?: string` and the mapper keeps it. Mapping is extracted to one function so the full and lightweight readers can't diverge.
- New `fetchGrantMilestonesForProgram()` — a **single** V2 request, used by the indexing polls instead of the old three-request full read (and instead of the V1 SDK path).
- `fetchMilestoneInstance` and both refetches are deleted. `chainID` now comes from `milestone.chainId`, `completed` from `completionDetails`, and the dead `communityUID` threading is gone.
- `useMilestoneCancellation` gets the same treatment: recipient from V2, no `projectById`. Its `data` argument is now unused and has been removed from the mutation args (caller updated).
- **Guard**: `requireMilestoneRecipient()` throws a typed `MissingMilestoneRecipientError` **before any wallet interaction** when the recipient is missing/malformed. It never falls back to `project.owner` or `payoutAddress` — that would change the attested value.

### #64 — failures name their cause and carry telemetry

- New `utilities/milestones/attestationFailure.ts` maps an error onto a concise, user-appropriate cause: signer-unavailable, network-switch, missing-recipient, indexing-timeout, insufficient-funds, wallet-rpc, network, server (with HTTP status), unknown. User rejection keeps its dedicated "Verification cancelled" copy.
- The Sentry context now carries `projectUid`, `chainId`, `programId`, `failureKind` and a **step marker** (`setup` / `fetch` / `backend` / `attest` / `poll`).
- `errorManager` leaves a `suppressed-error` breadcrumb on every early-return path (user-rejected, switch-chain, expected-state, transient-network, transient-http). No path is silent in both directions any more.
- Both `console.error` calls removed; the verify and complete catches share one `reportMilestoneFailure` exit point.

### #66 — poll against the actual signer

- `setupChainAndWalletForMilestone` now resolves the signing address from `walletSigner.getAddress()`, falling back to `useSetupChainAndWallet().smartWalletAddress`, and builds a tolerant candidate set that also includes every Privy-linked address (`getLinkedWalletAddresses`) and, last, the wagmi address.
- `matchesSubmittedVerification()` matches the indexed `verifiedBy` against that set. When the indexer hasn't resolved an attester yet it accepts a verification whose `attestationUID` differs from the one captured **before** signing — so a milestone somebody else already verified can't produce a false positive.
- On poll timeout the copy is now "submitted on-chain and is still being indexed … no need to verify again", and it is **not** reported to Sentry (indexer lag is not a defect).

### #66 (follow-up) — the attester is the MultiAttester contract, not the wallet

The candidate set above was still missing the address that actually lands. `GapContract.multiAttest` is this flow's only live attestation path — the app sets no `gelatoOpts`, so the `multiAttestBySig` branch is dead, and the EAS-direct `Milestone.verify()` path was deleted in #1951. It submits through the chain's MultiAttester (`Networks[network].contracts.multicall`), which is therefore `msg.sender` at EAS, so the indexer records **the contract** as `verifiedBy` — never the user's wallet.

On-chain confirmation: verification `0x7b2fb1bf…` (multiAttest path) has `verifiedBy` `0x7177adc0…`, the Base MultiAttester, versus `0xb5162b77…` with `verifiedBy` `0x8353…` from the old direct path.

Without this, every verification would still have polled to timeout on a transaction that had already succeeded. Two changes:

- `getMultiAttesterAddress(chainId)` resolves the MultiAttester per chain from the SDK (lower-cased, `null` for unsupported chains) and joins the candidate set — generic, so a new network needs no change here.
- `matchesSubmittedVerification()` now accepts a verification whose `attestationUID` differs from the pre-signing snapshot even when `verifiedBy` is truthy but unrecognised. The snapshot guard is unchanged, so a pre-existing verification with an unmoved UID is still never accepted.

### #67 — nothing after the tx reads the wagmi address

The post-attest `throw` is gone. The poll takes the setup-resolved identity, so a null `useAccount().address` no longer affects anything after submission. The old `!address` **pre-flight** guard removed in #1821 is deliberately **not** reintroduced — it caused the silent no-op that removal fixed.

## Wire contract

**Unchanged.** This is read-only consumption of an existing V2 response field (`grantMilestones[].recipient`) that the backend already declares and maps. No new endpoints, no request-shape changes, no new environment variables, no backend PR required.

## Testing

`__tests__/integration/mutations/useMilestoneCompletionVerification.mutation.test.tsx` (new, 13 cases) drives the real hook with a spied `api` client:

- a full verify run performs **zero** `GET /projects/:uid` calls (both the SDK `projectById` spy and the raw request URLs are asserted), and reads `/v2/projects/:uid/updates`;
- attestation payloads carry the V2 `recipient` verbatim, in the right order (`completed` then `verified`), and drop the completion attestation when the milestone is already completed;
- a missing recipient produces a clear toast and **no** `multiAttest` call, in both verify and complete;
- the poll succeeds when the indexed attester is the signer while wagmi reports an unrelated wallet (#66);
- a null wagmi address post-attest completes with a success toast and no `errorManager` call (#67);
- poll exhaustion produces the "still being indexed" copy and is kept out of Sentry;
- error mapping produces the named cause and the `step: "attest"` Sentry context; a user rejection keeps "Verification cancelled";
- the reviewer flow attests verification-only with no re-read after the backend completion.

Also:

- `__tests__/utilities/milestones/attestationIdentity.test.ts` — address validation, candidate building, and every branch of the verification match predicate.
- `__tests__/utilities/milestones/attestationFailure.test.ts` — one case per failure class.
- `__tests__/services/milestones.recipient.test.ts` — `recipient` survives both mappers; the lightweight reader issues exactly one V2 request with the chain suffix stripped.
- `__tests__/unit/utilities/errorManager.test.tsx` — a suppressed network-class error still leaves a breadcrumb (all five suppression paths).
- `__tests__/integration/mutations/useMilestoneCancellation.mutation.test.tsx` — updated for the V2 recipient; asserts no `projectById` and the missing-recipient block.

### Manual verification steps

1. As a community admin on a program with a pending-verification milestone, open the milestone review page with the network tab filtered to `projects` — verify a full verification run issues no `GET /projects/:uid`.
2. Sign in with an email/Google account that also has a linked external wallet connected (the hybrid shape). Verify a milestone: it should succeed within normal indexing time rather than spinning for five minutes.
3. Force a chain-switch rejection and confirm the toast names the network-switch cause rather than the bare generic string.
4. Confirm the completion and cancellation flows still attest with the same recipient.

### Local results

- `npx vitest run` over the touched suites: **256 passed / 24 files**.
- `tsc --noEmit`: clean.
- `pnpm lint:fix` (Biome): clean on touched files.
- `scripts/check-anti-patterns.sh` on every touched file: clean.
- `pnpm run build`: passes.

## Review findings addressed

No findings were waived. Everything raised was fixed in follow-up commits:

**Self-review (commit 2):**

- The poll's attester candidate set included wagmi's address, which can be an unlinked wallet left connected from a previous session — dropped, so the match stays within the authenticated identity.
- `isUserRejectionError` is a substring match and an `ApiError`'s message embeds the endpoint path, so an API failure on a route containing "reject" would have been reported to the user as "cancelled". API errors now bypass that check.
- A schema-contract violation was being described as a network failure; it now gets its own message.
- Expected failures (signer not ready, indexer lag) skip Sentry but now leave a breadcrumb, so no path is silent in both directions.
- The grant-milestones endpoint is built in one place, shared by the full and poll-friendly reads.

**Anti-pattern check (commit 3):**

- `[EMPTY_CATCH]` ×2 — the inline `api.post(...)` swallow in the verify and cancel flows is replaced by the existing `notifyIndexer` helper. A failed nudge is still non-fatal (the indexer's chain listener picks the tx up regardless) but it is now reported instead of vanishing.
- `[EMPTY_CATCH]` ×1 — documented that the grant lookup's catch reports through `errorManager` rather than swallowing.

**React Doctor (commit 3):**

- `buildAttestationPayloads` was rebuilt on every render despite reading nothing from the hook — moved to module scope.

### Follow-up test coverage (MultiAttester)

- Poll succeeds when `verifiedBy` is the Base MultiAttester `0x7177AdC0f924b695C0294A40C4C5FEFf5EE1E141`, and for Arbitrum's `0x6dC1D6b864e8BEf815806f9e4677123496e12026`. Both cases omit `attestationUID` so they exercise the attester-candidate path, not the UID fallback — each fails if the MultiAttester is dropped from the candidate set.
- Poll rejects a pre-existing verification (unchanged `attestationUID`, unrecognised attester) and still reports "still being indexed".
- `getMultiAttesterAddress` is asserted against every SDK-supported chain, plus stringified and malformed chain ids.
- Existing candidate matches (signer, linked wallet, unresolved attester) are unchanged.

Local: `attestationIdentity` + `useMilestoneCompletionVerification` suites 37/37 pass; the wider `__tests__/utilities/milestones`, `__tests__/integration/mutations` and `__tests__/hooks` run is 617/617. `pnpm run build` exits 0.

## Known CI note

The first Vercel preview build failed with exit 137 / OOM. This is the recurring preview-build flake on this repo (the build container runs out of memory), not a regression from this PR — `pnpm run build` passes locally against this branch.

Fixes show-karma/super-gap#63
Fixes show-karma/super-gap#64
Fixes show-karma/super-gap#66
Fixes show-karma/super-gap#67

## Summary by CodeRabbit

* **New Features**
  * Milestone completion, verification, and cancellation now use recipient information from V2 milestone data.
  * Added clearer milestone failure messages for wallet, network, indexing, server, and transaction issues.
  * Added improved verification tracking across supported attesters and indexer updates.

* **Bug Fixes**
  * Prevented transactions when a milestone recipient is missing.
  * Improved cancellation reliability and cache refresh behavior.
  * Added breadcrumb diagnostics for expected or suppressed errors without changing user-facing notifications.

### DIFF
--- a/__tests__/integration/mutations/useMilestoneCancellation.mutation.test.tsx
+++ b/__tests__/integration/mutations/useMilestoneCancellation.mutation.test.tsx
@@ -8,10 +8,7 @@
  */
 import { act, waitFor } from "@testing-library/react";
 import { useMilestoneCancellation } from "@/hooks/useMilestoneCancellation";
-import type {
-  GrantMilestoneWithCompletion,
-  ProjectGrantMilestonesResponse,
-} from "@/services/milestones";
+import type { GrantMilestoneWithCompletion } from "@/services/milestones";
 import { HttpError } from "@/utilities/api/errors";
 import { renderHookWithProviders } from "../../utils/render";
 
@@ -107,27 +104,15 @@ const baseMilestone = (
   description: "desc",
   dueDate: "2025-01-01",
   status: "pending",
+  recipient: RECIPIENT,
   completionDetails: null,
   verificationDetails: null,
   fundingApplicationCompletion: null,
   ...overrides,
 });
 
-const data = {
-  project: { uid: "0xproject" },
-  grantMilestones: [],
-} as unknown as ProjectGrantMilestonesResponse;
-
 beforeEach(() => {
   vi.clearAllMocks();
-  mockProjectById.mockResolvedValue({
-    grants: [
-      {
-        details: { programId: "1013" },
-        milestones: [{ uid: MILESTONE_UID, recipient: RECIPIENT }],
-      },
-    ],
-  });
 });
 
 const renderCancellationHook = () =>
@@ -140,7 +125,6 @@ describe("useMilestoneCancellation", () => {
     await act(async () => {
       await result.current.cancelMilestone({
         milestone: baseMilestone(),
-        data,
         reason: "superseded",
       });
     });
@@ -149,6 +133,22 @@ describe("useMilestoneCancellation", () => {
     expect(mockPayloadFor).toHaveBeenCalledWith(0);
     // Indexer notified with the attestation tx hash.
     expect(mockApiPost).toHaveBeenCalledWith(expect.stringContaining("0xattesttx"), {});
+    // The recipient comes from the V2 milestone payload — no V1 project refetch (#63).
+    expect(mockProjectById).not.toHaveBeenCalled();
+  });
+
+  it("blocks cancelling with no recipient and never signs (#63)", async () => {
+    const { result } = renderCancellationHook();
+
+    await expect(
+      act(async () => {
+        await result.current.cancelMilestone({
+          milestone: baseMilestone({ recipient: undefined }),
+        });
+      })
+    ).rejects.toThrow(/missing its on-chain recipient/);
+    expect(mockMultiAttest).not.toHaveBeenCalled();
+    expect(mockProjectById).not.toHaveBeenCalled();
   });
 
   it("blocks cancelling a completed milestone", async () => {
@@ -164,7 +164,6 @@ describe("useMilestoneCancellation", () => {
               completedBy: "0xgrantee",
             },
           }),
-          data,
         });
       })
     ).rejects.toThrow(/already been completed or verified/);
@@ -182,7 +181,6 @@ describe("useMilestoneCancellation", () => {
               description: "done via application",
             } as GrantMilestoneWithCompletion["fundingApplicationCompletion"],
           }),
-          data,
         });
       })
     ).rejects.toThrow(/already been completed or verified/);
@@ -204,7 +202,6 @@ describe("useMilestoneCancellation", () => {
               reason: null,
             },
           }),
-          data,
         });
       })
     ).rejects.toThrow(/already cancelled/);
@@ -253,7 +250,7 @@ describe("useMilestoneCancellation", () => {
     const { result } = renderCancellationHook();
 
     await act(async () => {
-      await result.current.cancelMilestone({ milestone: baseMilestone(), data });
+      await result.current.cancelMilestone({ milestone: baseMilestone() });
     });
 
     expect(mockMultiAttest).toHaveBeenCalledTimes(1);
--- a/__tests__/integration/mutations/useMilestoneCompletionVerification.mutation.test.tsx
+++ b/__tests__/integration/mutations/useMilestoneCompletionVerification.mutation.test.tsx
@@ -1,0 +1,459 @@
+/**
+ * Flow tests for useMilestoneCompletionVerification (super-gap #63/#64/#66/#67).
+ *
+ * These cover the four defects behind GAP-FRONTEND-261:
+ *  - #63 no V1 `GET /projects/:uid` anywhere in a verify/complete run; the
+ *    attested recipient comes from the V2 milestone payload and a missing one
+ *    blocks BEFORE any transaction;
+ *  - #64 failures name their cause and carry a step marker into Sentry;
+ *  - #66 the indexing poll matches the Privy-resolved signer, not wagmi;
+ *  - #67 a null wagmi address after the tx can no longer fail the run.
+ */
+
+import type { User } from "@privy-io/react-auth";
+import { act, waitFor } from "@testing-library/react";
+import { errorManager } from "@/components/Utilities/errorManager";
+import { useMilestoneCompletionVerification } from "@/hooks/useMilestoneCompletionVerification";
+import type {
+  GrantMilestoneWithCompletion,
+  ProjectGrantMilestonesResponse,
+} from "@/services/milestones";
+import { renderHookWithProviders } from "../../utils/render";
+
+// Hoisted so the `vi.mock` factories below (which run before the module body)
+// can read them. wagmi's account is deliberately a DIFFERENT wallet from the
+// signer in every test — the hybrid Privy account shape that made the old poll
+// unmatchable (#66).
+const {
+  PROJECT_UID,
+  MILESTONE_UID,
+  RECIPIENT,
+  SIGNER_ADDRESS,
+  WAGMI_ADDRESS,
+  PROGRAM_ID,
+  wagmiAccount,
+} = vi.hoisted(() => ({
+  PROJECT_UID: "0xproject",
+  MILESTONE_UID: `0x${"a".repeat(64)}`,
+  RECIPIENT: "0x1111111111111111111111111111111111111111",
+  SIGNER_ADDRESS: "0x2222222222222222222222222222222222222222",
+  WAGMI_ADDRESS: "0x3333333333333333333333333333333333333333",
+  PROGRAM_ID: "1013",
+  wagmiAccount: {
+    current: { address: undefined as string | undefined, chain: { id: 42161 } },
+  },
+}));
+
+vi.mock("wagmi", () => ({
+  useAccount: () => wagmiAccount.current,
+}));
+
+vi.mock("@/hooks/useWallet", () => ({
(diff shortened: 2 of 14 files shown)


brunodmsi:
@coderabbitai review

brunodmsi:
## Status summary at `54a95b31`

**CI: 29/30 checks green.** All 6 test shards, `static-checks` (taskless), `quality-gate`, `react-doctor`, `smoke`, `baseline-guard`, `checklist`, `build`, and the full QA pipeline (`qa-execution`, `dogfood`, `verdict`) pass.

**Quality bots — all findings fixed, none waived:**

| Source | Finding | Resolution |
|---|---|---|
| Anti-pattern check | `[EMPTY_CATCH]` ×2 — inline `api.post(...)` swallow in the verify + cancel flows | Replaced with the existing `notifyIndexer` helper, which reports the failure instead of discarding it. Still non-fatal: the indexer's chain listener picks the tx up regardless. |
| Anti-pattern check | `[EMPTY_CATCH]` ×1 — grant lookup catch | Documented that it reports through `errorManager`. |
| React Doctor | `buildAttestationPayloads` rebuilt every render | Moved to module scope — it reads nothing from the hook. |
| Self-review | Poll accepted wagmi's address as an attester | Dropped: it can be an unlinked wallet from a previous session. Candidates are now signer + Privy-linked addresses only. |
| Self-review | `isUserRejectionError` is a substring match, and an `ApiError`'s message embeds the endpoint path | API errors now bypass that check — a route containing "reject" was being reported to the user as "cancelled". |
| Self-review | Schema-contract violations described as network failures | Given their own message. |
| Self-review | Expected failures (signer not ready, indexer lag) were silent in both directions | They now leave a Sentry breadcrumb while still skipping capture. |

`Quality Checklist` now reports **"All automated checks passed"**; React Doctor reports **"No issues found!"**.

### Blocked: Vercel preview deploy

The `Vercel` check fails with `exited with 137` / OOM in the build container. **This is not from this branch:**

- `pnpm build` passes locally against this branch at Vercel's exact limit (`NODE_OPTIONS='--max-old-space-size=4096'`).
- An explicit `vercel redeploy` of the identical commit reproduced the same OOM.
- PRs #1950, #1918 and #1914 currently show the same failing `Vercel` check, and every preview deployment in the project for the last hour has errored — including one that predates this PR.
- #1918 recorded the same build-container failure a week ago.

Fixing it means giving the preview builds more memory (Enhanced Build Machines) or trimming the build's peak footprint — infra work outside this PR's scope. Flagging rather than patching build config here.

**Not merging** — leaving that decision open.

PR #1954, opened 2026-07-30
fix(seo): truthful article date model for knowledge pages

amaurymagalhaes:
## Overview

Replaces the fabricated `datePublished` / `dateModified` values on the 25 static knowledge-base articles with dates taken from each page's first commit, and surfaces that date to readers.

Part of DEV-583 (AEO improvements). Resolves DEV-591 (AEO-08).

## Problem

Every page under `app/knowledge/` shipped the exact same two dates in its Article JSON-LD:

```tsx
datePublished="2025-01-15"
dateModified="2026-03-24"
```

Neither is true:

- The pages were added to the repository in January **2026**, not January 2025, so `datePublished` predates the articles by a year.
- `dateModified="2026-03-24"` is a copy-pasted literal, not a record of any edit. It claims freshness on 25 pages that were never revised, and it goes stale the moment nobody hand-bumps it.
- Both dates were invisible to readers — they existed only inside the structured data. Asserting a fact in JSON-LD that appears nowhere in the rendered content is exactly the pattern search engines treat as untrustworthy markup.

## Solution

**One provenance map, sourced from git.** `app/knowledge/articleDates.ts` records the UTC calendar date of the commit that first added each page. These articles are not CMS-backed (unlike `/blog/[slug]`, which carries Sanity's `publishedAt`), so the introducing commit is the only truthful per-article record. Dates were recovered from the GitHub commits API rather than a local `git log`, because clones of this repo are frequently shallow and grafted commits report every file as newly added. The three source commits and the exact query are documented in the file header.

**Unknown slugs fail loudly.** `getKnowledgeArticleDate(slug)` throws for an unrecorded slug, and each page calls it at module scope — so adding a knowledge page without recording its date breaks the build instead of silently shipping a missing or invented date.

**No `dateModified` at all.** Deriving one from git last-touch dates would fabricate freshness, since those timestamps are dominated by merges and mechanical reformatting. This matches the existing decision to keep `lastModified` off these routes in `app/sitemaps/static/sitemap.ts`. `ArticleJsonLd` already omits the key when the prop is absent, so the emitted schema simply has no `dateModified`.

**The date is visible.** `components/Knowledge/ArticlePublishedDate.tsx` renders `Published <time dateTime="…">Jan 6, 2026</time>` in a `<header>` beside each `<h1>`, using the shared `formatDate` helper pinned to UTC so the rendered day never drifts by timezone. The structured data and the page now assert the same fact.

## Testing steps

Automated:

```bash
  __tests__/integration/pages/knowledge-article-dates.test.tsx \
  __tests__/components/Seo/ArticleJsonLd.test.tsx --reporter=dot
```

19 tests, all passing. They cover:

- the date map has an entry for every directory under `app/knowledge/` and no orphan entries;
- recorded dates are plausible past dates and are not the two previously fabricated literals;
- `getKnowledgeArticleDate` throws for an unrecorded slug;
- no page source reintroduces a hardcoded date literal or a `dateModified` prop;
- each rendered page emits the recorded `datePublished`, emits no `dateModified`, and shows the same date in its server-rendered markup;
- `ArticleJsonLd` includes or omits each date key according to the props supplied.

Also run: `pnpm typecheck` (clean) and `pnpm biome check` over the changed paths (clean).

Manual:

1. `pnpm run dev`, open `/knowledge/grant-lifecycle`. A "Published Jan 6, 2026" line appears under the title.
2. View source and find the `application/ld+json` Article block: `datePublished` matches the visible date and there is no `dateModified` key.
3. Spot-check a page from each source commit — `/knowledge/grant-kyc` (Jan 7) and `/knowledge/onchain-project-profiles` (Jan 14) — to confirm dates differ per article rather than being uniform.

## Risk / scope notes

- Scope is limited to the static `app/knowledge/` routes. `/blog/[slug]` and other `ArticleJsonLd` callers are untouched; the component's props are unchanged and both date props remain optional.
- Search impact is the intended one: these pages lose a `dateModified` signal they were never entitled to, and gain a `datePublished` that matches both reality and the visible page. Rendering-wise the only change is one line of text per article.
- The throwing lookup is deliberate. It runs at module scope during build, so a missing entry surfaces as a build failure rather than a runtime error for users.
- Backfilling a real `dateModified` later remains possible, but it should come from an explicit editorial record (a per-article "last reviewed" field maintained by whoever edits the copy), not from git metadata.

### DIFF
--- a/__tests__/components/Seo/ArticleJsonLd.test.tsx
+++ b/__tests__/components/Seo/ArticleJsonLd.test.tsx
@@ -1,0 +1,64 @@
+import { render } from "@testing-library/react";
+import { ArticleJsonLd } from "@/components/Seo/ArticleJsonLd";
+import { SITE_URL } from "@/utilities/meta";
+import "@testing-library/jest-dom";
+
+describe("ArticleJsonLd", () => {
+  function getSchema(container: HTMLElement) {
+    const scripts = Array.from(container.querySelectorAll('script[type="application/ld+json"]'));
+    expect(scripts).toHaveLength(1);
+    return JSON.parse(scripts[0].innerHTML);
+  }
+
+  const baseProps = {
+    title: "The Grant Lifecycle",
+    description: "Follow the complete grant process.",
+    url: "/knowledge/grant-lifecycle",
+  };
+
+  it("emits an Article schema with the site URL and Karma as publisher", () => {
+    const { container } = render(<ArticleJsonLd {...baseProps} />);
+    const schema = getSchema(container);
+
+    expect(schema["@context"]).toBe("https://schema.org");
+    expect(schema["@type"]).toBe("Article");
+    expect(schema.headline).toBe(baseProps.title);
+    expect(schema.url).toBe(`${SITE_URL}/knowledge/grant-lifecycle`);
+    expect(schema.publisher.name).toBe("Karma");
+  });
+
+  it("includes datePublished when it is supplied", () => {
+    const { container } = render(<ArticleJsonLd {...baseProps} datePublished="2026-01-06" />);
+    const schema = getSchema(container);
+
+    expect(schema.datePublished).toBe("2026-01-06");
+  });
+
+  it("includes dateModified when it is supplied", () => {
+    const { container } = render(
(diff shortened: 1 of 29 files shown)


amaurymagalhaes:
@coderabbitai review

PR #1955, opened 2026-07-30
fix(seo): populate the LLM URL index, strip docs chrome, and cover all three audiences

amaurymagalhaes:
## Overview

Two AEO fixes to the machine-readable surface of the site: the generated LLM URL index
(`llms.txt` / `llms-full.txt`) has been publishing an empty "Site URL Index" section, and
every non-whitelabel page has been advertising a Speakable schema it isn't eligible for.

Part of DEV-583. Closes DEV-593.

## Problem

**1. The "Site URL Index" section ships empty.**

`scripts/generate-llms-txt.ts` fetched `/sitemap.xml` and scraped it with a `<url>…</url>`
regex. But `/sitemap.xml` is a sitemap *index* — `utilities/sitemap.ts` builds a
`<sitemapindex>` whose children (`/sitemaps/static/sitemap.xml`, `communities`, `projects`,
`funding-programs`) are the real `<urlset>` documents. The regex matched nothing, the
`catch` swallowed it, and the generator fell back to the hardcoded `STATIC_PAGES` list.

Every URL in that fallback is also a landing page, and the URL-index section filters out
anything already listed under Landing Pages — so the section rendered with a header and
zero links. Generation still exited 0, so the monthly workflow published the hollow file
without complaint. This has been the case for months.

**2. Global Speakable JSON-LD is invalid on every page it ships on.**

`SpeakableJsonLd` was mounted in the root layout, so every non-whitelabel route emitted a
`WebPage` + `SpeakableSpecification` schema pointing at generic CSS selectors. Google
restricts Speakable to eligible news content; Karma publishes none. It's an inapplicable
schema on ~every route, and invalid structured data dilutes the markup crawlers do act on.

## Solution

**Sitemap index traversal** (`85e232c`) — detect `<sitemapindex>` and pull entries from its
children. Only `/sitemaps/static/sitemap.xml` is allowlisted: it holds the ~50 top-level
pages plus blog posts, which is what a curated index means here and what
`SITEMAP_LABEL_MAP` / `SITEMAP_DESCRIPTION_MAP` are written for. The entity children carry
up to `MAX_URLS_PER_SITEMAP` (45,000) near-identically-labeled URLs each and would swamp
the document, so recursion is allowlisted rather than blanket.

Fetch failures now throw instead of degrading to `STATIC_PAGES`. The silent fallback is
precisely what let the empty section go unnoticed.

**Required-section guard** (`d3f862f`) — validate rendered output before writing. `llms.txt`
must have link bullets under Landing Pages, Site URL Index and Optional; `llms-full.txt`
under Site URL Index. An empty section now fails the run, so the monthly workflow goes red
rather than committing a degraded file.

**Speakable removal** (`32e6525`) — delete the component and its layout mount, plus an
architecture test asserting the root layout mounts nothing Speakable and no source file
emits `SpeakableSpecification`. Speakable may legitimately return one day, but scoped to a
single eligible article rather than mounted globally.

## Testing steps

```bash
  __tests__/unit/scripts/generate-llms-txt-sitemap.test.ts \
  __tests__/architecture/no-global-speakable-jsonld.test.ts
```

New coverage lives in `__tests__/unit/scripts/generate-llms-txt-sitemap.test.ts`. The
existing `generate-llms-txt.test.ts` is already past the 800-line file limit, so appending
there tripped the quality gate's oversized-file check; the split also draws a cleaner line
between URL discovery / output validation and the text-extraction helpers. That file is
byte-identical to `main`.

- `parseSitemapIndexLocs` / `parseUrlSetEntries` — index vs urlset documents, entity
  decoding in `<loc>`, blocks missing a `loc`, and each parser returning empty for the
  other's document type.
- `fetchSitemapEntries` (stubbed `fetch`) — asserts it requests `/sitemap.xml` then *only*
  `/sitemaps/static/sitemap.xml`; parses a direct urlset without a second fetch; rejects on
  an empty child, on an index with no allowlisted child, and on a non-OK response.
- `assertRequiredSections` — reproduces the exact silent regression (every sitemap entry is
  also a landing page, so the index filters to nothing) and asserts it now throws, for both
  documents; plus missing section, no-bullet section, and unknown document name.
- Architecture test scans `app/`, `components/`, `src/` for `SpeakableSpecification`.

`pnpm typecheck` is clean. Biome reports one pre-existing warning in
`scripts/generate-llms-txt.ts` (unused `homeTarget`, already on `main` at line 1626) —
untouched here to keep the diff scoped.

To eyeball the generated output end to end, run the generator against production and check
the section is populated:

```bash
pnpm tsx scripts/generate-llms-txt.ts
grep -A5 '## Site URL Index' public/llms.txt
```

## Risk / scope notes

- **Generator now fails loudly.** Previously any sitemap problem degraded silently; now the
  monthly workflow errors. That's the intent — a red workflow is recoverable, a quietly
  hollow `llms.txt` is not — but it does mean a production sitemap outage will fail the job
  rather than publish stale-but-present content.
- **Allowlist is deliberately narrow.** If a new *static-page* sitemap child is added, it
  must be added to `SITEMAP_INDEX_CHILD_ALLOWLIST` or its URLs won't appear in the index.
  Entity sitemaps are intentionally excluded.
- **No runtime/app behaviour changes** beyond removing one
  `<script type="application/ld+json">` from the root layout. `OrganizationJsonLd` is
  untouched. Whitelabel tenants never rendered Speakable, so they're unaffected.
- The generated `public/llms.txt` is not regenerated in this PR; the scheduled workflow
  picks up the fix on its next run.

## Audience coverage (added)

`llms.txt` and `agents.md` are the surfaces answer engines actually read, and both
still described Karma as a builder/ecosystem platform. The published `llms.txt`
contained **zero** occurrences of "foundation" or "donor advisor" and 16 of
"builder"/"ecosystem" — so an engine ingesting Karma's own agent-facing files
learned nothing about the three target audiences.

All of these strings were hardcoded in the generator, so regenerating never fixed them.

- **Organization JSON-LD** now uses a dedicated `ORGANIZATION_DESCRIPTION` naming all
  three audiences. `DEFAULT_DESCRIPTION` is deliberately left alone — it is the
  metadata fallback for community/program/project pages, not an entity description.
  The schema is mounted in the root layout, so every claim is visible on the route it
  describes.
- **`llms.txt` / `llms-full.txt`** open with the three-audience framing. The overview
  paragraph was hardcoded twice; both call sites now share one builder, and the chain
  list is derived from `SUPPORTED_NETWORKS` rather than frozen in a literal.
- **`agents.md`** gains *Funder discovery*, *Nonprofit research* and *Grant program
  operations* routing bullets. Previously every "when to use Karma" bullet was
  grantee-shaped, so an agent asked to find funders for a nonprofit had no instruction
  to call Karma despite `/nonprofits/find-funders` doing exactly that.

Onchain funding is retained as a capability rather than the frame. Homepage title and
meta description are unchanged by design — that is positioning, and it stays with the
owner.

Coverage in the generated `llms.txt`, before → after: foundation 0 → 4,
donor advisor 0 → 5, nonprofit 6 → 9, IRS 990 0 → 1.


amaurymagalhaes:
@coderabbitai review

PR #1956, opened 2026-07-30
feat(seo): server-render and hydrate funding program details

amaurymagalhaes:
## Overview

Server-renders the whitelabel funding-program detail route (`/community/[communityId]/programs/[programId]` and its whitelabel-domain equivalent) so the program's core facts are in the initial HTML instead of behind a client fetch.

Part of DEV-583 (AEO improvements); implements DEV-585 / AEO-02.

## Problem

The route's `page.tsx` was a thin sync server component that rendered `ProgramDetailClient` and nothing else. Everything a reader cares about — title, description, budget, funding range, grant types, dates — was fetched client-side by `useProgram`, so the server response contained only a `animate-pulse` skeleton.

These URLs ship in the funding-programs sitemap, which means:

- crawlers and answer engines that do not execute JavaScript index a placeholder, not the program;
- a reader with JS disabled or still loading sees a skeleton for the full round-trip;
- the server was already fetching the same program in `generateMetadata` and throwing the body away after reading the title and description.

Separately, whether applications are open, closed, or not yet started was only *implied* by the sidebar's apply button. There was no statement of the status anywhere in the page copy.

## Solution

**Prefetch and hydrate.** `page.tsx` is now an async server component that prefetches the program into a `QueryClient` and hands it to the client tree through a `HydrationBoundary`. The fetch is the same `React.cache`-wrapped `getProgramDetails` that `generateMetadata` calls, so it stays one indexer round-trip per request rather than two.

**Shared key and freshness window.** The prefetch and `useProgram` now build their key from `wlQueryKeys.programs.detail(programId)` and share an exported `PROGRAM_DETAIL_STALE_TIME`. A hydrated entry is only picked up when both hashes match and it is inside its stale time, so both are pinned in `__tests__/lib/query-keys.test.ts`.

**Error handling mirrors the hook.** The server fetch previously collapsed every failure to `null`. It now returns `null` only on a 404 and rethrows anything else, so a transient indexer 5xx is never hydrated into the client cache as a definitive "Program not found". `prefetchQuery` swallows the rejection and `dehydrate` omits failed queries, so an indexer outage degrades to exactly today's behaviour: skeleton plus a client retry. `retry: false` keeps a failing prefetch from blocking the response on React Query's backoff. `generateMetadata` keeps its own `.catch(() => null)` and still degrades to generic copy.

**Status is stated, not implied.** `ProgramDetailClient` renders `getProgramStatusInfo(program).label` as a pill next to the title, so "Open for Applications" / "Applications Closed" / "Coming Soon" is real text in the body. No structured-data or metadata fact is added that is not visible in the rendered content.

The endpoint path in both the server fetch and the hook now comes from `INDEXER.V2.FUNDING_PROGRAMS.GET` instead of being written inline.

Three commits: the key/endpoint refactor, the status pill, then the SSR change.

## Testing steps

Automated:

```bash
pnpm vitest run --project unit __tests__/app/program-detail-ssr.test.tsx --reporter=dot
pnpm vitest run --project unit __tests__/lib/query-keys.test.ts --reporter=dot
pnpm vitest run --project unit __tests__/integration/pages/smoke/whitelabel-pages.test.tsx --reporter=dot
pnpm typecheck
```

`__tests__/app/program-detail-ssr.test.tsx` is the acceptance suite. It renders the page through `react-dom/server`'s `renderToString`, which runs no effects — whatever it asserts on is literally what a no-JS client receives. It covers:

- title, status, description, budget, funding range, grant types and both dates present, with no `animate-pulse` in the output;
- exactly one indexer call across `generateMetadata` and the render path;
- open / closed / coming-soon status copy;
- a 404 rendering the not-found state server-side;
- a 500 degrading to the skeleton and *not* to "Program not found".

Manual:

1. `pnpm run dev`, open a community program detail page.
2. `curl` the same URL (or view-source) and confirm the program title, status label, description and sidebar facts are in the HTML payload.
3. Disable JavaScript and reload — the page content is there, only the interactive controls are inert.
4. Confirm no client request to `/v2/funding-program-configs/:id` fires on first load (the hydrated entry is fresh); a hard reload after 5 minutes refetches as before.

## Risk / scope notes

- Scoped to one route plus the program detail query key. `useProgram` behaviour is unchanged apart from the shared constants — its consumers see the same return shape.
- The page now awaits an indexer call before responding. It was already awaiting the same call in `generateMetadata`, and `React.cache` dedupes them, so TTFB is unchanged in the happy path.
- Failure mode is a strict superset of today's: prefetch failures are dropped from the dehydrated state and the client fetches exactly as it does on `main`.
- `getProgramDetails` is stricter than before (rethrows non-404s). That error surfaces inside `prefetchQuery`, which catches it; `generateMetadata` catches it explicitly. Neither path can now produce an unhandled rejection.
- The status pill is additive UI above the byline; no existing element moved or changed.

## Summary by CodeRabbit

- **New Features**
  - Funding program detail pages now render program information on the initial page load when available, reducing reliance on loading placeholders.
  - Program status is displayed prominently, including “Open for Applications” and “Applications Closed.”
  - Program details are reused between initial loading and subsequent interactions for a smoother experience.

- **Bug Fixes**
  - Missing programs now show a clear “Program not found” state.
  - Temporary service errors gracefully fall back to the existing loading experience.

### DIFF
--- a/__tests__/app/program-detail-ssr.test.tsx
+++ b/__tests__/app/program-detail-ssr.test.tsx
@@ -1,0 +1,204 @@
+/**
+ * AEO acceptance tests for the funding-program detail route (DEV-585).
+ *
+ * These URLs ship in the funding-programs sitemap, so the core facts have to be
+ * in the server-rendered HTML rather than behind a client fetch. Every test
+ * here renders through `react-dom/server`'s `renderToString`, which runs no
+ * effects — whatever it produces is exactly what a crawler or a reader with
+ * JavaScript disabled receives.
+ */
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import type React from "react";
+import { renderToString } from "react-dom/server";
+import type { FundingProgram } from "@/types/whitelabel-entities";
+import { HttpError } from "@/utilities/api/errors";
+
+const mockGet = vi.fn();
+
+vi.mock("@/utilities/api/client", () => ({
+  api: {
+    get: (...args: unknown[]) => mockGet(...args),
+    post: vi.fn(),
+    put: vi.fn(),
+    patch: vi.fn(),
+    delete: vi.fn(),
+    request: vi.fn(),
+    getPaginated: vi.fn(),
+  },
+}));
+
+vi.mock("next/navigation", () => ({
+  useParams: () => ({ communityId: "c1", programId: "prog-1" }),
+  useRouter: () => ({ push: vi.fn(), replace: vi.fn() }),
+  usePathname: () => "/community/c1/programs/prog-1",
+}));
+
+// The real provider depends on the Privy bridge, which has no server-side
+// equivalent. Guest defaults (`isLoading: true`) are what actually resolves
+// during SSR, so mirror them rather than granting permissions.
+vi.mock("@/src/core/rbac/context/permission-context", () => ({
+  PermissionProvider: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  usePermissionContext: () => ({
+    isLoading: true,
+    isCommunityAdmin: false,
+    isReviewer: false,
+    can: () => false,
+  }),
+}));
+
+vi.mock("@/src/components/navigation/Link", () => ({
+  Link: ({ children, ...props }: { children: React.ReactNode } & Record<string, unknown>) => (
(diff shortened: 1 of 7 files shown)


amaurymagalhaes:
@coderabbitai review

PR #1957, opened 2026-07-30
feat(seo): server-render community project entities

amaurymagalhaes:
## Overview

Makes the community project hubs — `/community/[communityId]` and
`/community/[communityId]/projects` — actually ship their project entities in the
server-rendered HTML, and makes the full paginated list reachable without
JavaScript.

Part of DEV-583. Implements DEV-584 (AEO-01).

## Problem

Both hubs already fetched page 1 on the server and passed it into
`CommunityGrants` as `initialProjects`, but that payload never reached the
rendered grid:

- The server asked the indexer for `{ page: 1, limit: 12 }` with no `sortBy`,
  while the client's first `useCommunityProjectsInfinite` call asked for
  `sortBy=milestones`. Different request, different ordering — the client
  response replaced the server one wholesale.
- `initialProjects` was only used for the total count. The grid itself rendered
  from React Query, which starts empty, so the initial HTML contained a skeleton
  and zero project names, titles, or links.
- `refetchOnMount: "always"` means React Query reports `isRefetching` on the very
  first render, and `isFilterLoading` blanked the grid to a skeleton whenever
  that was true — so even a seeded cache would have been hidden.
- There was no non-JS path past page 1. Everything beyond the first 12 projects
  lived behind an infinite-scroll observer, so crawlers only ever saw 12 of them.

Net effect: answer engines and crawlers indexed a loading state for every
community hub.

## Solution

**Shared request helpers** (`utilities/queries/v2/communityProjectsRequest.ts`)

The sort/status mappers and the page size lived inside the `"use client"` hook, so
server components could not reuse them. They now live in a framework-free module
(mirroring `utilities/projects-explorer-request.ts`), alongside page-param parsing
and page-href building. The first commit is a pure move with no behaviour change.

**Server fetches the client's exact first request**

Both pages now call `getCommunityProjects` with `{ page, limit, sortBy }` built
from those shared helpers, so the response is byte-for-byte the request the client
would have issued and can legitimately seed the cache.

**Seeding is conditional, and deliberately conservative**

- The seed is only applied while the live filter state still matches the server
  defaults (sort, maturity stage, categories, program, tracks). A user landing on
  a filtered URL never sees the unfiltered server payload.
- An empty payload is never seeded. `getCommunityProjects` swallows failures into
  an empty page, so "no projects" and "indexer blip" are indistinguishable —
  seeding an empty page would render a permanent false empty state. The client
  refetch decides instead.
- `initialPage` is part of the query key, so a `?page=2` seed cannot bleed into
  the page-1 cache entry (same policy the `/projects` explorer uses).
- `isFilterLoading` no longer blanks the grid when data is already held. Every
  filter is in the query key, so any data in hand already belongs to the current
  filters.

**Crawlable pagination**

`?page=N` is now honoured server-side (malformed, zero, negative, and
array-valued params fall back to page 1), and the grid renders `rel="prev"` /
`rel="next"` anchors built from `PAGES.COMMUNITY.*` constants. Page 1 keeps the
bare path so the canonical URL never gains a redundant `?page=1`. These are plain
`<a>` anchors, not `<Link>` — in QA on `/projects`, client interception swallowed
the navigation. Infinite scroll remains the progressive enhancement for humans
with JavaScript, and the links only render on the unfiltered view.

No structured-data facts were added; everything here is content that is now
genuinely present in the rendered HTML.

## Testing steps

Automated (all green):

```bash
  __tests__/utilities/communityProjectsRequest.test.ts \
  hooks/__tests__/useCommunityProjectsInfinite.test.tsx \
  __tests__/components/CommunityGrants/CommunityGrants-ssr.test.tsx \
  __tests__/app/ssr-server-components.test.tsx
pnpm typecheck
```

28 tests covering: helper parsing/href edge cases; the hook's seed and query-key
behaviour; all 12 project names present in the initial HTML; seeded projects
staying visible through the mount revalidation; crawlable links on first, deep,
and single-page communities; the seed being dropped under a non-default filter;
an empty server payload not being seeded; and the empty state still resolving.
Page-level tests assert the indexer receives `page`/`limit`/`sortBy` and that
`?page=2` and malformed params are handled.

Manual:

1. `pnpm run dev`, then `curl -s http://localhost:3000/community/<slug>/projects | grep -c '<project title>'` — project names should appear in the raw HTML.
2. Same for `?page=2`; confirm the HTML shows the second page and carries both prev and next anchors.
3. With JavaScript disabled, click Previous/Next and confirm full page navigations resolve.
4. With JavaScript enabled, confirm the grid does not flash a skeleton on load and that infinite scroll still appends pages.
5. Apply a category or sort filter and confirm results are filtered (not the server payload) and the crawlable links disappear.

## Risk / scope notes

- Scope is limited to the two community hub pages, `CommunityGrants`, and its
  infinite-query hook. No API, schema, or indexer changes.
- Highest-risk area is the cache seed. It is gated on an exact filter-state match
  and a non-empty payload, and `initialPage` is in the query key, so the failure
  mode of a bad seed (stale or mismatched data) is structurally excluded rather
  than assumed away.
- The `isFilterLoading` change is the one behaviour users will notice: switching
  filters no longer blanks the grid on the very first render when data is already
  held. Subsequent genuine filter changes still show the skeleton because the
  query key changes and there is no data for the new key.
- `?page=N` is new public surface on these routes. Invalid values degrade to page
  1 rather than erroring, and page 1 renders at the bare canonical path.

### DIFF
--- a/__tests__/app/ssr-server-components.test.tsx
+++ b/__tests__/app/ssr-server-components.test.tsx
@@ -43,15 +43,21 @@ vi.mock("@/components/CommunityGrants", () => ({
     communityUid,
     categoriesOptions,
     initialProjects,
+    initialPage,
+    paginationBasePath,
   }: {
     communityUid: string;
     categoriesOptions: string[];
     initialProjects: { payload: unknown[] };
+    initialPage?: number;
+    paginationBasePath: string;
   }) => (
     <div data-testid="community-grants">
       <span data-testid="community-uid">{communityUid}</span>
       <span data-testid="categories-count">{categoriesOptions.length}</span>
       <span data-testid="projects-count">{initialProjects.payload.length}</span>
+      <span data-testid="initial-page">{initialPage}</span>
+      <span data-testid="pagination-base-path">{paginationBasePath}</span>
     </div>
   ),
 }));
@@ -158,6 +164,7 @@ describe("SSR Server Components", () => {
       const CommunityProjectsPage = await importPage();
       const jsx = await CommunityProjectsPage({
         params: Promise.resolve({ communityId: "test-community" }),
+        searchParams: Promise.resolve({}),
       });
 
       render(jsx);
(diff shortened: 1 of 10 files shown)


amaurymagalhaes:
@coderabbitai review

PR #1967, opened 2026-08-03
fix(seo): make the three failing sitemap URLs readable without JavaScript

amaurymagalhaes:
## Overview

A production sitemap crawl (`node scripts/indexability/crawl-sitemap.mjs`) against tag **v1.8.20** failed **3 of 204** URLs. All three are AEO/SEO *eligibility* defects — a crawler that does not execute JavaScript cannot read the page. This fixes all three and pins each with a no-JS regression test.

From **DEV-586 (AEO-03)**.

| URL | Crawl classification | Baseline | Root cause |
|---|---|---|---|
| `/seeds` | `non-self-canonical` | `canonical: https://www.karmahq.xyz/` | metadata literal had no `alternates`, so it inherited the root layout's `alternates.canonical: "/"` |
| `/communities` | `thin` | `h1: null`, `textLength: 256` | `<h1>` sat below two loading early-returns and an error branch |
| `/nonprofits/find-funders` | `thin` | `h1: null`, `textLength: 1147` | whole landing page went through `dynamic(..., { ssr: false })` |

### 1. `/seeds` — self-referential canonical

`/seeds` is submitted in `app/sitemaps/static/sitemap.ts`, so it has to be the canonical of its own content. Added `alternates: { canonical: "/seeds" }` to the existing metadata literal.

**Not** routed through `customMetadata`: that helper derives openGraph/twitter title and description from the page `title`/`description`, and this page deliberately runs three different pairs of copy (page title, OG title, Twitter title). Using it would have silently dropped them. A test asserts the bespoke OG/Twitter copy survives.

`app/seeds/fund/page.tsx` has the same inherited-canonical shape, but it is **not** sitemap-listed (not in `staticPages`, absent from the crawl report), so it is deliberately left alone — nothing obliges an unsubmitted page to self-canonical, and changing it is out of scope here.

### 2. `/communities` — unconditional heading

`CommunitiesPage` is a `"use client"` component that early-returned `<CommunitiesSkeleton />` on `communitiesLoading && statsLoading`, then again on `communitiesLoading`, with an error branch in between. The title block — icon pill, `<h1>Organizations on Karma</h1>`, description, "Add your organization" CTA — lived below all of them, and the server render is *always* the loading case. So production served a headingless page.

The header block now renders unconditionally at the top of a single return. Only the stats row and the organization grid are gated on query state:

- **loading** → `<CommunitiesSkeleton withHeader={false} />` (new prop, so the skeleton no longer draws a placeholder heading underneath the real one)
- **error** → message + a **Try again** button wired to `refetch` (it previously had no recovery path)
- **empty** → "No communities found."
- **success** → stats row + infinite-scroll grid

The static "Add Your Community" block also moved outside the branches, so it is part of the server HTML too. No `return null` anywhere; visual layout unchanged in the success state.

### 3. `/nonprofits/find-funders` — **Option (a): removed `ssr: false`**

**Chosen option: (a)** — enabled SSR rather than duplicating a hero into `page.tsx`. The `landing-page-dynamic.tsx` wrapper existed only to make `ssr: false` legal from a Server Component; with SSR on it has no purpose, so it is deleted and `page.tsx` imports `LandingPageClient` directly.

**Evidence that nothing blocks SSR** — audited every browser-only touch reachable from `landing-page-client.tsx` and its imports (`connect-cta.tsx`, `install-configs.ts`, `connect-targets.ts`, `stats.ts`, `store/search-session.ts`, `components/ui/input.tsx`, `src/components/navigation/Link.tsx`, `hooks/use-url-builder.ts`, `hooks/useCopyToClipboard.ts`):

- `landing-page-client.tsx:1033,1035` — `document.getElementById` / `document.querySelector` are inside `scrollToHero`, a `useCallback` invoked from an `onClick`. Never runs during render.
- `connect-cta.tsx:82,90` — `window.sessionStorage` behind `typeof window === "undefined"` guards, called from a `useEffect` and a click handler.
- `connect-cta.tsx:123-125` — `document.getElementById` + `IntersectionObserver` inside a `useEffect`.
- `store/search-session.ts` — zustand `persist` with the default storage. Verified in `node_modules/zustand/middleware.js:245-251`: `createJSONStorage` wraps `getStorage()` in `try/catch` and returns `undefined` on the server, so `persist` no-ops. The store is only ever read via `getState()` inside a callback, never selected during render.
- `hooks/use-url-builder.ts:15` — `window.location.hostname` behind a `typeof window` guard (pre-existing pattern used by every SSR'd page in the app).
- No `matchMedia`, no non-SSR-safe third-party library, nothing at module scope.

**Hydration-mismatch check:** the two components with post-mount state both render their pre-mount value identically on server and first client render — `ConnectFloatingCard` starts `dismissed: true` (returns `null` on both), `RotatingPlaceholder` starts with an empty `sub` string. `crypto.randomUUID()` only runs inside `createSession`, on submit. The production build compiled clean and the served page shows no hydration error.

Result: the `<h1 className="lp-hero-title">` and all eight content sections are now in the server HTML.

## Verification

`pnpm build && pnpm start` (build exit 0), then fetched each route over plain HTTP with no JavaScript:

```
$ curl -s http://localhost:3000/communities | grep -o '<h1[^>]*>[^<]*</h1>'
<h1 class="text-4xl sm:text-[72px] font-bold text-black dark:text-white">Organizations on Karma</h1>

$ curl -s http://localhost:3000/seeds | grep -o '<link rel="canonical"[^>]*>'
<link rel="canonical" href="https://www.karmahq.xyz/seeds"/>
```

For find-funders the `[^<]*` form of the grep matches nothing — **not** because the heading is missing, but because that `<h1>` contains nested markup (`<br/>` and a `<span>`), which `[^<]*` cannot cross. With a tag-aware pattern:

```
$ curl -s http://localhost:3000/nonprofits/find-funders | grep -oP '<h1\b[^>]*>.*?</h1>'
<h1 class="lp-hero-title">Stop hunting for funders.<br/><span class="italic">Ask an agent.</span></h1>

$ curl -s http://localhost:3000/nonprofits/find-funders | grep -o '<h1' | wc -l
1

$ curl -s http://localhost:3000/nonprofits/find-funders | grep -o '<link rel="canonical"[^>]*>'
<link rel="canonical" href="https://www.karmahq.xyz/nonprofits/find-funders"/>
```

Re-running the crawler's own metrics (strip script/style/tags, collapse whitespace) against the served HTML, versus the v1.8.20 baseline in the crawl report. The crawler's threshold is `textLength >= 200 && Boolean(h1)`:

| URL | baseline `h1` / `textLength` | now `h1` / `textLength` |
|---|---|---|
| `/communities` | `null` / 256 | 1 / 689 |
| `/nonprofits/find-funders` | `null` / 1147 | 1 / 10894 |
| `/seeds` | canonical `→ /` | canonical `→ /seeds` |

All three now satisfy `meaningful: true` / self-canonical.

## Tests

New `__tests__/app/aeo-crawl-eligibility.test.tsx`, following the precedents in `community-subpage-canonicals.test.ts` and `knowledge-collection-ssr.test.tsx` — assertions are made on `renderToString` output, which runs no effects and resolves no queries, i.e. exactly what a non-executing crawler receives.

- `/seeds`: self-canonicals at `/seeds`; bespoke OG/Twitter copy preserved; still present in the static sitemap (the fact that obliges it to self-canonical).
- `/communities`: the `<h1>` and lead paragraph are present while **both queries are loading** (the server-render case), when the query **errors** (alongside the retry), and when it resolves **empty**; exactly one `<h1>` while loading, so the skeleton adds no second heading.
- `/nonprofits/find-funders`: hero `<h1>` in server output, exactly one `<h1>`, and >2,000 chars of real landing copy covering every section — not a spinner.

```
$ ./node_modules/.bin/vitest run __tests__/app/aeo-crawl-eligibility.test.tsx
 Test Files  1 passed (1)
      Tests  12 passed (12)

$ ./node_modules/.bin/vitest run __tests__/accessibility/community-page-a11y.test.tsx \
    __tests__/app/knowledge-collection-ssr.test.tsx __tests__/app/community-subpage-canonicals.test.ts
 Test Files  3 passed (3)
      Tests  42 passed (42)
```

`knowledge-collection-ssr.test.tsx` carried a comment asserting `/communities` is legitimately thin and that stubbing `CommunitiesPage` "changes nothing a crawler would have seen". That is no longer true, so the comment is corrected to say why the stub is there (its React Query hooks need a provider that file does not set up) and to point at the new coverage. No assertion in that file changed.

`pnpm lint:fix` run; `tsc --noEmit` clean; the pre-push typecheck hook passed.

## Notes / not verified

- Verified against a **local** production build only. Not yet re-crawled against a deployed preview — worth re-running `crawl-sitemap.mjs` against the preview URL before merge.
- `app/seeds/fund/page.tsx` still inherits the root canonical. Left alone on purpose (not sitemap-listed) — flagging it here rather than fixing it silently.
- The pre-commit anti-pattern scan emitted 15 pre-existing `no-hardcoded-color-tsx` warnings in `components/Pages/Communities/Loading.tsx` (the `pickColor` palette and a `#000000`, all on lines this PR does not touch). Not introduced here, not fixed here.

---

## Follow-up: QA finding — hero was in a hidden streamed chunk (`c42627d`)

QA verified that after the first commit the find-funders HTML, while fully server-rendered, arrived as a React streaming payload: the page content sat in a `<div hidden id="S:0">` chunk revealed only by the client-side `$RC()` swap, with the route's loading fallback as the visible page. A genuine no-JS renderer still saw a spinner-only shell. Reproduced locally against the built output.

### Root cause

Every route in this app renders dynamically (the root layout awaits `headers()` for whitelabel detection), and with a `loading.tsx` anywhere above a page, Next defers the page segment behind that Suspense boundary — by design ("instant loading states"). The page's HTML always streams as a hidden chunk; the visible part of the response is a loading fallback. This is not specific to find-funders: **production today serves /projects, /seeds, /knowledge/*, and /about with their `<h1>`s inside hidden chunks** (verified against `https://www.karmahq.xyz` with the same tag-stack checker). The sitemap crawler reads raw HTML and does not model `hidden`, which is why only find-funders (which had NO h1 at all under `ssr: false`) was flagged.

Worse, **which ancestor's fallback ends up visible is a per-build module-chunk race**. Instrumented across six builds: the same route flushed at the find-funders boundary in one build and at the /nonprofits boundary in the next, purely from unrelated import changes. Three candidate fixes were implemented and disproven against the built output before the final one:

- **(landing) route-group layout holding the hero** — group layout modules load lazily; the suspension bubbled to the parent boundary and hid the hero *and* the section navbar.
- **`export const dynamic = "force-static"`** — the route built as ○ Static, but Next serializes the fallback + hidden chunk *into the prerendered HTML* too.
- **Meaningful loading.tsx fallback (static hero replica, zero imports)** — rendered correctly, but the flush point regressed to the `/nonprofits` skeleton fallback in the very next build; nothing anchored inside the segment system survives chunk-graph changes.

The only region React *guarantees* into the initially visible HTML — by Suspense semantics, not by winning a race — is the root layout's own output, outside every boundary. In every one of the six instrumented builds plus production, the root layout markup was in the visible shell.

### The fix

- `proxy.ts` tags `/nonprofits/find-funders` requests (that one path only) with an `x-pathname` request header — layouts cannot read the pathname themselves.
- `app/layout.tsx` renders `<FindFundersNoscriptHero />` when the tag matches: a `<noscript>`-wrapped static hero (h1, lead copy, example prompts) using the same `lp-*` classes. A JS-executing visitor (including Googlebot's renderer) never displays it — they get the interactive hero when hydration reveals the page — while a no-JS reader renders exactly this block. No duplication in either mode.
- Hero copy moved to `src/features/non-profits/lib/hero-content.ts`, shared by the interactive hero and the noscript replica so they cannot drift; the interactive hero split into `landing-hero.tsx` (rendered inside `LandingPageClient` as before — visual output unchanged).
- `loading.tsx` reverted to the pre-existing spinner.

### Verification (local `pnpm build && pnpm start`, build exit 0)

Tag-stack checker (parses the full HTML, tracks the `hidden` attribute through nested elements) on the served page:

```
{"offset":8305,  "insideHidden":false, "insideNoscript":true}   ← visible shell, no-JS reader's h1
{"offset":20433, "insideHidden":true,  "insideNoscript":false}  ← real page h1, revealed on hydration
```

```
$ curl -s http://localhost:3000/nonprofits/find-funders | grep -oP '<h1\b[^>]*>.*?</h1>' | head -1
<h1 class="lp-hero-title">Stop hunting for funders.<br/><span class="italic">Ask an agent.</span></h1>

$ curl -s http://localhost:3000/nonprofits/find-funders | grep -o '<noscript><div class="landing">.\{0,160\}'
<noscript><div class="landing"><section class="lp-hero"><div class="lp-container lp-hero-inner"><h1 class="lp-hero-title">Stop hunting for funders.<br/><span class="italic">Ask an agent.</span></h1><p class="lp-hero-sub">AI agents that find the right foundations…
```

All three original fixes re-verified on the same build:

```
$ curl -s http://localhost:3000/communities | grep -o '<h1[^>]*>[^<]*</h1>'
<h1 class="text-4xl sm:text-[72px] font-bold text-black dark:text-white">Organizations on Karma</h1>

$ curl -s http://localhost:3000/seeds | grep -o '<link rel="canonical"[^>]*>'
<link rel="canonical" href="https://www.karmahq.xyz/seeds"/>

$ curl -s http://localhost:3000/nonprofits/find-funders | grep -o '<link rel="canonical"[^>]*>'
<link rel="canonical" href="https://www.karmahq.xyz/nonprofits/find-funders"/>
```

Hydration sanity (Playwright against the local prod build): exactly one `<h1>` in the live DOM, one `lp-hero` section, search input functional, `<noscript>` at zero height; only console errors are the local-only `/_vercel/insights` 404s.

Other routes carry no noscript hero (`/`, `/nonprofits`, `/communities`, `/seeds` all checked — 0 occurrences).

### Tests

- `__tests__/app/aeo-crawl-eligibility.test.tsx` (15 tests): noscript hero carries the h1 + lead copy + all example chips; its h1 matches the page h1 exactly; page has exactly one h1; full landing sections in server output; canonical; per-subroute loading files.
- `__tests__/middleware-find-funders-header.test.ts` (6 tests): the proxy sets `x-pathname` on exactly `/nonprofits/find-funders` and on no sibling/sub/other route.
- Full pass: 7 files, 110 tests green (includes middleware-dashboard, middleware-indexability, knowledge-collection-ssr, community-page-a11y, community-subpage-canonicals).

### Known limitation (flagging, not hiding)

The renderToString-based tests cannot exercise Next's streaming, so the "h1 outside `hidden`" property is proven by the curl/tag-stack checks above against the built output, and structurally by the proxy-header + noscript-component tests. If the streaming behavior is ever revisited app-wide (every other sitemap route still ships its h1 inside a hidden chunk in production), that belongs to a separate initiative.


amaurymagalhaes:
## `c2cae0b` — test (1) shard fix + P6 chip-click investigation

**Test (1) shard — real failure, fixed.** Not the OOM flake: reproduced locally under the CI heap cap (`NODE_OPTIONS=--max-old-space-size=6144`, `--maxWorkers=2 --shard=1/6`) — 4 failures in `__tests__/integration/pages/layout.test.tsx`, all `headers was called outside a request scope` from the root layout's new direct `await headers()` read (the vitest summary was buried under console noise in the CI log). Fix: the pathname read moved into `getRequestPathname()` in `utilities/whitelabel-server.ts` — the module that already owns request-scope header reads and is the layout suite's existing mock seam — and the suite mocks it alongside `getWhitelabelContext`. Shard 1/6 re-run locally: **175 files, 2802 tests, green.** Rebuilt and re-curled after the refactor: noscript hero still present on find-funders (1 block, h1 visible), still absent on `/`, `/nonprofits`, `/communities`.

**QA P6 (chip clicks do nothing) — not reproducible, and the QA evidence shows the chips working.** The chip handler is byte-identical to `main` (`diff` of `onChipClick` between `fdbb36ef3` and this branch: no changes — the code moved files, nothing else). Local matrix against the built PR branch: all 6 chips × 3 viewports (1280/1366/mobile) × real mouse clicks × 6x CPU throttle × slow-network chunk delay — **every click types the chip text and navigates to `/nonprofits/find-funders/search/[id]`**. The QA run's own screenshot (`P6-fail.png`) shows the input mid-typewriter with "Funders of refugee resettlement giving ove|" — i.e. the chip click had fired and the animation was typing. The chip interaction intentionally animates the text into the input at ~16ms/char (60–70 char prompts ≈ 1.1s) before it navigates; sampling the input/URL inside that window reads as "empty input, no navigation, no errors". The animation timing is pre-existing behavior, unchanged from main.

PR #1984, opened 2026-08-04
feat(nonprofit-research): public server-rendered diligence answers on the section index

amaurymagalhaes:
## Overview

AEO-12 experiment **E4** ([DEV-595](https://linear.app/showkarma/issue/DEV-595)) — contract pre-registered in `artifacts/aeo-measurement/experiments.md`. Draft on purpose: copy AND an auth-surface change need human review before the ready-PR AI stack spends.

**Revised per reviewer feedback**: the "Nonprofit research for donor advisors" answer block originally sat at the footer of `/nonprofit-research` and read as bolted onto the product page. It now lives as a knowledge-base article at **`/knowledge/nonprofit-due-diligence`**, following the established `app/knowledge/*` pattern (Article JSON-LD, truthful per-article date from `articleDates.ts`, knowledge index entry, static sitemap entry, llms.txt category map). `/nonprofit-research` stays public and indexable with the experiment title/meta, and keeps a footnote-level one-liner linking to the article (per a second review pass, the descriptive intro sentence was removed and the pointer shrunk to text-xs muted with tight padding).

## Contract (binding, from experiments.md)

Changed cohort **C13** (/nonprofit-research), control **C12** (/donor-advisors — interference caveat documented). Target prompts: P003, P004, P009, P017, P024, P034, P041, P042. Pass = citations earned on target prompts and citation-share movement beats C12 across two consecutive 7-day windows from ship date. releases.csv row at ship time.

## Contract note (E4 measurement after the reviewer's change)

The E4 contract targeted `/nonprofit-research` (C13) as the single changed surface carrying the diligence Q&A in its server HTML. Per review, the experiment now ships as a **two-surface change**, and the ship-time rows should be written accordingly (artifacts/ is deliberately NOT edited in this PR):

- **Informational prompts (P003, P004, P009, P017, P024, P034, P041, P042)** now target `/knowledge/nonprofit-due-diligence`: that page carries the full Q&A, the FAQPage JSON-LD, and the Article JSON-LD with `datePublished: 2026-08-04`. Any citation earned on these prompts is expected to land on the knowledge article URL, not on `/nonprofit-research`.
- **C13 (`/nonprofit-research`) keeps the buyer-intent framing**: the measurable change to that URL is now noindex removal + public index + experiment title/meta + a footnote-level one-liner linking to the article. Citation movement on C13 itself would come from indexability and title/meta, not from answer content.
[...]

PR #1992, opened 2026-08-04
fix(seo): make sitemap routes readable without JavaScript by removing loading boundaries (DEV-612)

amaurymagalhaes:
## Overview

DEV-612 (AEO-17). Fixes the app-wide pattern where every route with a `loading.tsx` above it streams its page HTML as a hidden Suspense chunk (`<div hidden id="S:n">`), revealed only by the client-side `$RC` swap. Because the root layout awaits `headers()` for whitelabel detection, every route renders dynamically, and the loading fallback is what a non-executing reader sees. Result before this PR: a hidden-aware crawl classified 203/204 production sitemap URLs as thin in no-JS mode; only `/nonprofits/find-funders` passed, and only because of its hand-built `<noscript>` hero.

**This is an architecture change. Opened as DRAFT for human review. Do not merge without sign-off.**

## Approach chosen: remove the Suspense boundary above crawlable routes

The mechanism is `loading.tsx` + dynamic rendering. With no loading boundary anywhere above a page, React has nothing to flush a fallback into, so the response blocks until the page renders and the page HTML lands in the initially visible shell. This PR:

1. Deletes `loading.tsx` from every sitemap-listed route and all its ancestor segments (21 files): the root `app/loading.tsx`, the static marketing and knowledge pages, blog, `mcp/connect`, the nonprofits landing plus connect guides, and the crawl-listed entity routes (`project/[projectId]`, community home, `funding-opportunities`, `programs/[programId]`). Non-crawlable routes (admin, manage, dashboard, the find-funders workbench subroutes, and everything else auth-gated) keep their instant loading states.
2. Ratchets the invariant: `__tests__/app/route-file-structure.test.ts` now carries a `SITEMAP_NO_LOADING` set and asserts, for each crawlable route, that no `loading.tsx` exists anywhere on its segment chain. Re-adding one fails CI instead of silently hiding a route from no-JS readers again. The CLAUDE.md route-trio rule is amended to match.
3. Hoists the `/funding-map` h1 out of the client search component into the server page, outside the page's own inner Suspense boundary. That was the one crawlable route whose heading still streamed hidden after the boundary removal (page-level `<Suspense>` around a client component is the same mechanism in miniature).
4. Retires the DEV-586 find-funders `<noscript>` hero, the proxy's `x-pathname` tag, and `getRequestPathname`. With the real hero, sections, and FAQ now in the visible HTML, the replica made text-mode readers see the hero twice. The crawl-eligibility tests now assert on the page's server output directly.
5. `/nonprofit-research` keeps its loading boundary and its documented crawl known-issue (auth-gated advisor workspace, E4, PR #1984). The known-issue comment in `scripts/crawl-sitemap.mjs` is updated to say it is now the only route that still streams behind a fallback.

## Alternatives rejected, with evidence

- **(b) Stop awaiting `headers()` in the root layout / make routes static.** Two independent blockers. First, PR #1967 already disproved the useful half: with `export const dynamic = "force-static"` the route builds Static, but Next serializes the fallback plus the hidden chunk into the prerendered HTML too, so static rendering does not remove the hidden-chunk problem while a loading boundary exists. Second, whitelabel host detection must branch per request on `Host` for every route (whitelabel domains serve the same paths); moving it to proxy-injected headers still reads request state, which is still dynamic. Static main-site rendering would need PPR or a separate route tree. Not pursued.
- **(c) PPR / Cache Components.** Correction from review: `cacheComponents` is a stable top-level flag in Next 16, not experimental. It is still not a DEV-612-sized change: the root layout reads `headers()` for whitelabel detection, and under Cache Components runtime-header readers must sit behind Suspense, which recreates exactly the fallback-visible/content-hidden result for no-JS readers. Adopting it properly means separating host-independent public content from tenant chrome and defining cache lifetimes for all public entity data: a deliberate platform migration, tracked as the long-term replacement for block rendering, not this fix.
- **(d) Extend the proxy + noscript mechanism app-wide.** Rejected: it duplicates every page's content and has to be hand-maintained per route. This PR deletes the one existing instance because the real content is now visible.
- Also re-confirmed from PR #1967: route-group layout hoisting and meaningful zero-import loading fallbacks are not fixes; the flush point is a per-build module-chunk race, and only content outside every boundary is guaranteed visible.

## Crawl before and after

Local production build (`pnpm build && pnpm start`), crawled with the repo crawler's own `crawlSitemap` in `--visibility-mode no-js`, over a 76-URL tree: all 49 static-sitemap URLs the local build serves, the communities sitemap, and strided samples of the projects (15) and funding-programs (10) kinds from the indexer the build points at. Transport rewritten to localhost; canonical origin kept at production.

| | ok | thin | failing (non-allowlisted) |
|---|---|---|---|
| before (98952d9) | 1/76 | 75 | 74 |
| after | 75/76 | 1 | 0 |

The one remaining thin URL is `/nonprofit-research`, allowlisted by its existing known-issue entry. Report `ok: true`. Per class, all of: static, knowledge (26), marketing, blog, `/project/[id]` (15/15), `/community/[id]`, `/community/[id]/funding-opportunities`, and `/community/[id]/programs/[pid]` (10/10) flipped from thin to ok. Tag-stack checks confirm the h1s sit outside any `hidden` subtree.

## TTFB before and after

Same machine, local production build against the **staging** indexer (network RTT to staging measured at ~525 ms per call, so entity-route absolutes are inflated versus Vercel production where the indexer is a few tens of ms away). 3-5 runs per URL after one warmup; values are means.

| Route | TTFB before | TTFB after | Content-complete before | Content-complete after |
|---|---|---|---|---|
| `/` (home) | 35 ms | 15 ms | 38 ms | 16 ms |
| `/about` | 26 ms | 13 ms | 28 ms | 13 ms |
| `/knowledge/grant-accountability` | 13 ms | 12 ms | 14 ms | 13 ms |
| `/blog` | 16 ms | 15 ms | 17 ms | 16 ms |
| `/communities` | 16 ms | 13 ms | 17 ms | 14 ms |
| `/nonprofits/find-funders` | 20 ms | 14 ms | 22 ms | 15 ms |
| `/projects` | 14 ms | 21 ms | 1024 ms | 939 ms |
| `/project/[id]` (2 staging slugs) | 172 ms | 872 ms | 1397 ms | 1414 ms |
| `/community/celo` | 18 ms | 986 ms | 1191 ms | 987 ms |
| `/community/gitcoin/funding-opportunities` | 18 ms | 624 ms | 646 ms | 626 ms |
| `/community/arbitrum/programs/101` | 18 ms | 543 ms | 837 ms | 620 ms |

Reading it honestly:

- **Content-complete time is unchanged everywhere** (slightly better on several routes). The entity pages always did this server work; it used to happen after an early empty-shell flush, now it happens before the first byte.
- **Static, knowledge, marketing, blog: no cost.** These pages have no server data work; they now arrive whole in ~15 ms.
- **`/projects` keeps a low TTFB** because its grid stays behind the page's own inner Suspense; the h1 and page copy are outside it.
- **Entity routes trade an empty 20 ms first byte for a full-content first byte** at 0.5-1.0 s locally against staging. What a user loses is the intermediate skeleton frame; what every non-executing reader gains is the actual content. On production infra the absolute TTFB will be substantially lower (most of the local number is two to three sequential ~525 ms staging round trips). Worth re-measuring on the Vercel preview before merge; if entity TTFB on production infra is deemed too high, the follow-up lever is trimming or parallelizing the layout prefetches, not restoring the boundary.
- Client-side navigations to crawlable routes now stay on the current page until the server responds (React transition behavior) instead of showing an instant skeleton. Accepted trade for this route set; the authenticated app surfaces keep their skeletons.

## Verification

- `pnpm build` green (3x: baseline, change, final).
- Local no-js crawl: table above; raw-mode equivalence implied (content unchanged, only visibility moved).
- Playwright against the local prod build, 11 routes (home, about, knowledge, projects, communities, find-funders, funding-map, community, funding-opportunities, program, project): **zero hydration errors, zero page errors, visible h1 on every route**. Remaining console noise is the pre-existing local-only 404s (`/_vercel/insights` etc.) and third-party frames.
- Whitelabel: full unit suite green including the whitelabel/layout/header-detection tests, plus a live smoke with `Host: testapp.opgrants.io` against the local build (tenant title, tenant theme token, community rewrite all intact).
- Full unit suite: **1056 files, 15138 tests, all passing.** `tsc --noEmit` clean. `pnpm lint:fix` run (only pre-existing warnings remain).
- Structural guards: updated `route-file-structure` ratchet (both directions) and `aeo-crawl-eligibility` tests pass.

## Known limitations / follow-ups

- The local crawl uses the staging indexer's entity URLs (the build's env), not the production URL set; the acceptance re-run on a deployed preview (`node scripts/crawl-sitemap.mjs --visibility-mode no-js` against production after ship) is the final word.
- `impacts`, `grants`, and `milestones` sitemap kinds are not currently in the production sitemap index. Their routes (`/project/[id]/impact`, `/funding/[grantUid]`, `/funding/[grantUid]/milestones-and-updates`) still have loading boundaries below `(profile)`. If those kinds are ever re-enabled in the sitemap, extend `SITEMAP_NO_LOADING` first.
- Entity-route TTFB on production infra should be spot-checked after deploy; the follow-up lever is reducing sequential server fetches in the project/community layouts.

## Addendum: adversarial review round (commit 52b0185)

An independent adversarial review (Codex, gpt-5.6-sol high) was run against this PR. Verdict: ship-with-changes, endorsing the architecture ("the only option that gives every requester the same host-correct, complete HTML through normal Next rendering") and rejecting the rivals (bot-UA-conditional rendering, edge HTML snapshots, prerender sidecars, noscript replicas) with reasons recorded in the review comment below. Two findings were confirmed empirically and fixed:

1. **Project pages passed the crawl on chrome alone.** The `(profile)` layout's explicit `<Suspense>` still hid the whole profile (visible text 226-392 chars vs 1200-2100 raw); the crawler's 200-char + h1 bar was met by navbar/footer plus the sr-only h1. The boundary's stated reason (useSearchParams) only applies to statically prerendered routes; removed. Project pages now put title, description, and tabs in the visible HTML: visible text 640-1344 chars (min/max across the 15 sampled pages). Same for `/funding-map`: search block and prefetched program cards now visible, visible == raw at 1697 chars. Program pages were checked and were already fully visible (visible == raw). Consequence for the TTFB table: project pages' first byte now waits for the full profile render, TTFB ~1.4s locally against staging, equal to their unchanged content-complete time.
2. **The root `app/loading.tsx` deletion had orphaned loading states of non-crawlable routes** that relied on it (admin, stats, super-admin, old-home, seeds/fund, funding-map subpages, donate, kyc-settings, program apply, project funding/impact/team tabs). Each got a route-local `loading.tsx`, placed only where no sitemap-listed route lives at or below the segment. The community `(whitelabel)/programs` listing page remains without one by necessity: a boundary there would sit above the crawlable `programs/[programId]` detail.

Review follow-ups deliberately NOT in this PR (tracked for reviewer judgment): unify the three hand-maintained route lists (sitemap sources vs `SITEMAP_NO_LOADING`) into one crawlability manifest with a contract test; extend crawl acceptance from "200 chars + h1" to semantic per-class anchors; gate merge on preview p50/p95 TTFB plus slow/failing-indexer injection.

Re-verified after the fixes: build green, no-js crawl 75/76 ok (`ok: true`), Playwright on 13 routes with zero hydration errors, full unit suite green (1056 files / 15138 tests), tsc clean.

---

## Addendum 2: hybrid streaming for project routes (commits 5e8914a, 53ee48e)

Two commits refine the project profile from "block the whole route" to "block the identity, stream the rest". **This partially supersedes Addendum 1, finding 1** — read the supersession note below before trusting the earlier project-page numbers.

### Why

Addendum 1 removed the `(profile)` layout's `<Suspense>` outright, which put the whole profile in the visible HTML but also put every server fetch on the critical path: `getProjectCachedData`, then a three-call `prefetchProjectProfileData` (grants + updates + impacts), then `getProjectFeed`. Nothing could flush until all of it resolved. Project TTFB and time-to-first-visible-shell became the same number.

### What changed

1. **Scoped Suspense in `app/project/[projectId]/(profile)/layout.tsx`.** Only `{children}` — the tab body — is wrapped. `ProjectProfileLayout` and `SidebarProfileCardStatic` stay outside it. Still no `loading.tsx` at or above the segment, so `SITEMAP_NO_LOADING` and the route-file-structure ratchet are unchanged.

   - **Outside the boundary (visible HTML):** the sr-only h1 and JSON-LD/breadcrumbs from the parent layout, the sidebar card's name, description and social links, and the profile navigation.
   - **Inside (streamed):** the tab body — the activity feed on the overview, or the client-queried funding/impact/team bodies.

2. **`prefetchProjectProfileData` dropped from the blocking path** in `app/project/[projectId]/layout.tsx`. It cost three indexer round-trips before the first byte and produced no server-rendered HTML: grants and impacts render only in client-only tab bodies, and the overview feed comes from `getProjectFeed` below the boundary. Its only visible effect was seeding the tab counters, which are hidden at zero (`ContentTabs` renders a badge only when `count > 0`) and now appear once the client queries resolve.

   Streaming it back from a lower `HydrationBoundary` was considered and rejected: the client `useQuery` mounts before a late boundary arrives and fires its own request anyway, so it would duplicate the fetch rather than replace it.

   The `PROJECT.DETAILS` prefetch stays blocking on purpose — `ProjectProfileLayout` drops `children` entirely while it has no project, so without it the shell never renders and the boundary never mounts.

3. **`useSearchParams` extracted out of `ProjectProfileLayout`** (React Doctor: CSR-bailout warning). Reading search params anywhere in a subtree opts that whole subtree out to client-side rendering — and here the subtree is the identity shell this PR exists to keep server-rendered. The invite-code reader, its open-once guard, its effect and the contributor-profile modal store moved into `ProjectInviteCodeWatcher`, rendered inside its own local `<Suspense>`. It renders nothing, so the boundary cannot affect the streamed shell. It is rendered in all four return branches (error, both loading branches, success) because the effect it replaced ran before every early return.

4. Minor, same file: the three skeletons are imported from source instead of the `../Skeletons` barrel (AGENTS.md), and the invite-code test's now-dead barrel mock is removed — nothing asserted on its testids.

### Supersession of Addendum 1, finding 1

Addendum 1 reported project pages at **640–1344 chars** of no-JS visible text with the full profile visible. That is no longer accurate. Measured on a local production build of `/project/check-ia` with the repo's own `extractNoJsVisibleHtml`:

| | bytes | visible text | h1 |
|---|---|---|---|
| First flush (up to first hidden chunk) | 56,402 | **1,018** | `Check-IA` (1) |
| No-JS visible (whole response) | 47,197 | **1,018** | `Check-IA` (1) |
| Full response | 220,298 | 13,380 | 1 |

First flush and no-JS view are identical, so the identity shell is entirely pre-boundary. The activity feed is absent from the first flush and present in the full response.

**The tradeoff is explicit and is a reversal of Addendum 1's direction:** milestone and update prose moved out of the no-JS-visible HTML into a hidden streamed chunk. No-JS readers now get 1,018 chars of project identity instead of the fuller profile. This buys early semantic FCP. If crawler value on project routes is the milestone content rather than the identity, this is the wrong trade and these two commits are cleanly revertible.

Crawl acceptance is unaffected: 1,018 chars with a real h1 clears the crawler's 200-char + h1 bar comfortably.

### TTFB

Local production build against the production indexer, `/project/check-ia`. Warm, 4 runs: **TTFB 0.68–1.31 s, full stream 1.45–2.08 s.** Cold first request: headers 1638 ms, stream complete 2403 ms.

Stated honestly: **there is no before/after delta for these two commits.** A baseline rebuild was not run, so this number stands alone. What is verified is the mechanism — three indexer calls left the critical path, and the shell now flushes ahead of the body rather than after it. The Addendum 1 row for `/project/[id]` (TTFB ~1.4 s locally against staging, equal to content-complete) is the closest comparison but is not apples-to-apples: different indexer, different network.

### Verification

- `pnpm build` **exit 0**. (Note: the build predates the `useSearchParams` extraction and the two import-only edits; those are covered by tests and Biome, and CI will confirm.)
- `tsc --noEmit` clean project-wide; Biome clean on every touched file.
- Focused suite **103/103 across 13 files**: the new streaming test, the DEV-612 route-structure ratchet, `aeo-crawl-eligibility`, `project-layout-ssr-shell`, `project-layout-h1`, `knowledge-collection-ssr`, `program-detail-ssr`, `ssr-server-components`, both `ProjectProfileLayout` tests, `ProjectInviteCodeWatcher`, `project-profile.mutation`, `prefetchProjectProfile`.
- New `__tests__/app/project-profile-streaming-shell.test.tsx` drives the real layouts through `renderToPipeableStream` and snapshots at `onShellReady` (shell) vs `onAllReady` (full), so it can distinguish visible HTML from hidden streamed chunks — something `renderToString` cannot model. The page body suspends on a promise the test releases only after the shell snapshot, so ordering is deterministic rather than timer-raced. **Mutation-verified:** removing the `Suspense` wrapper fails 3 of its 4 tests, including the identity-in-shell assertion.
- `route-file-structure.test.ts` needed no change — its ancestor-chain walk already encodes the hybrid rule, and this design adds no `loading.tsx`.

### Follow-ups from this round

- `prefetchProjectProfileData` is now orphaned in production code (referenced only by its own unit test and the unrendered `ProjectStatsBarAsync.tsx`). Left in place to keep the diff scoped.
- The layout-level invite-code test exercises only one of the four return branches that render `ProjectInviteCodeWatcher`; the error and success branches carry it untested.
- Hidden-chunk count varies by render timing (3 cold, 49 warm, same build) as React batches more into a cold shell before its first flush. Behaviour is identical in both; do not write assertions against that count.

## Summary by CodeRabbit

- **New Features**
  - Project profiles now display identity and navigation sooner while tab content loads.
  - Invite links reliably open the contributor profile dialog once.
  - Added accessible loading indicators for administrative and workflow pages.
  - Funding map search controls now support improved keyboard navigation and selection states.

- **Bug Fixes**
  - Improved server-rendered content across funder discovery, funding maps, projects, and crawlable pages.
  - Project updates remain visible during refreshes and offer retry options when requests fail.
  - Reduced duplicate or delayed content during loading.

- **Documentation**
  - Clarified route rendering and loading-state requirements.

---

### The contract this PR now enforces, per route class

QA scenarios keep re-deriving what "readable without JavaScript" means per route and reaching different conclusions, so it is written down here. **This is the spec; a scenario that contradicts it is testing the wrong thing.**

| Route class | Must be in the initially visible HTML | May stream into a hidden chunk |
|---|---|---|
| Marketing, knowledge, blog | Everything | — |
| `/projects` | h1, page copy, the project list itself | — |
| `/for-agents` | h1, FAQ answers, the tool catalog | — |
| `/funding-map` | h1, search block, program cards | — |
| `/community/[id]`, `/community/[id]/programs/[pid]` | Entity identity + listing content | — |
| **`/project/[projectId]`** | **Identity only:** sr-only h1, JSON-LD/breadcrumbs, sidebar name + description + socials, profile navigation | **The tab body** — activity feed, funding, impact, team |

Project routes are the single deliberate exception. Everything else in `SITEMAP_NO_LOADING` blocks and ships its substantive content.

**Why project routes differ.** The full profile blocks on `getProjectCachedData` → `getProjectFeed` plus the client tab queries. Serving all of it inline made time-to-first-visible-shell equal to time-to-everything. The scoped `<Suspense>` in `(profile)/layout.tsx` puts the identity on the wire first and streams the feed behind it.

**The honest cost:** milestone and update prose is no longer in the no-JS HTML for project pages. No-JS readers get 1,018 chars of identity (measured on `/project/check-ia`). This is a deliberate, reversible trade, revisitable against Agent Analytics evidence at the 2026-08-18 checkpoint — the same criterion in the earlier park note. QA scenario **P3** flags this as a failure; it is the intended contract, not a regression.

### QA findings fixed

**`/projects` streamed its whole list into a hidden chunk** (scenario P2, Critical). The list sat behind an in-page `<Suspense>` — the same anti-pattern as the removed `loading.tsx` files, expressed with an explicit boundary. Awaiting the loader inline, measured on a production build:

| | before | after |
|---|---|---|
| hidden chunks | 1 (`S:0`, ~3k projects) | **0** |
| no-JS visible text | skeleton only | **25,382 chars** |
| crawlable project links | 0 | **50** |
| TTFB | — | 1.77s |

**`/for-agents` hid its tool catalog** (scenario P8, High). Same shape, same fix; the boundary had been added earlier in this branch to keep the FAQ out of the streamed region, which awaiting inline achieves without hiding the catalog.

**Updates failures were silent** (scenario A1, High). Blocking the `/updates` request left the filters inert with no message and no retry. Two causes, both older than the streaming work: `UpdatesContent` gated on `!allUpdates || isUpdating`, but `aggregateProjectProfileData` always returns an array so `!allUpdates` was dead and only `isUpdating` drove the skeleton — which a request that never settles never clears; and the server-rendered feed was discarded at hydration, so real content vanished into that skeleton. The feed now persists until the client query produces data, and an error banner with a retry always renders on failure, above whatever content survives.

**A repaired prefetch.** The server-side updates prefetch hand-wrote `QUERY_KEYS.PROJECT.UPDATES(id)` — the first two elements of a key the hook builds from eight, including every filter value — so it warmed an entry nothing read. Both sides now go through an exported `projectUpdatesQueryKey`. Grants and impacts stay off the critical path.

**An invite-modal regression, caught in self-review.** `ProjectProfileLayout` returns different root element types per branch (`<div>` while loading, `<ErrorBoundary>` on success), so React remounts its children on that transition. With the open-once latch held in the extracted watcher's own state, it reset on the remount and re-opened a contributor-profile modal the user had already closed. The latch moved to the layout, which persists across branches. Dogfood independently confirmed the fix live.

**`useSearchParams` off the identity shell** (React Doctor). Reading search params bails the enclosing subtree out to client rendering, and that subtree was the server-rendered identity shell. The invite-code reader moved into `ProjectInviteCodeWatcher` behind its own `<Suspense>`; it renders nothing, so the boundary cannot affect the streamed shell.

**Accessibility** (CodeRabbit, Major). Funding-map category filters were clickable `<Badge>`s with no keyboard focus, no Enter/Space activation and no pressed state. Now native `<button>`s with `aria-pressed`.

### Found and filed, not fixed here

Both are pre-existing and independent of DEV-612.

- **#2000 — project routes answer 404 with an empty body.** `/project/<unknown-id>` returns 404 with **zero bytes** and no content-type, so the browser shows its native error page. `/blog/<unknown>` returns a 47KB body, `/unknown-path` 61KB — the root not-found path works elsewhere. Four hypotheses were implemented and disproven against real production builds (segment `not-found.tsx`; swallowing the notFound digest in the layout's catch; guarding `generateMetadata`; both guards plus a boundary). All four are recorded in the issue so nobody repeats them. The layout-vs-page distinction is **not** the mechanism. Base commit `98952d923` has the identical structure.

- **#2001 — Funding and Impact tabs stay on skeleton forever** despite 200 responses (scenario P6, High). Classified `pr-caused` and attributed to the prefetch narrowing; the attribution is very likely wrong. The old grants prefetch keyed on the **slug** while `FundingContent` reads via `useProjectGrants(project.uid)` — the **uid**. slug ≠ uid, so that prefetch never seeded the query the tab reads, and removing dead code cannot cause a hang. Same class as the updates key mismatch above. Flagged as a static read, not a reproduction.

### Verification

- `pnpm build` exit 0; `tsc --noEmit` clean; Biome clean on every touched file.
- Focused suites green throughout, including the DEV-612 route ratchet and the AEO crawl-eligibility tests.
- New `__tests__/app/project-profile-streaming-shell.test.tsx` drives the real layouts through `renderToPipeableStream` and snapshots at `onShellReady` versus `onAllReady`, so it can tell visible HTML from hidden streamed chunks — something `renderToString` cannot model. **Mutation-verified:** removing the `Suspense` wrapper fails 3 of its 4 tests.
- The invite-latch and updates-resilience regression tests are likewise mutation-verified against the previous logic.


amaurymagalhaes:
Adversarial review of this PR, run independently (Codex CLI, gpt-5.6-sol high) against the checkout and PR body. Findings 1 (orphaned loading states) and 2 (project/funding-map content still behind inner Suspense) were confirmed and fixed in 52b0185; the Cache Components wording in the body was corrected per the analysis below. Remaining recommendations are left to reviewer judgment.

# VERDICT

**ship-with-changes**

The architectural diagnosis is convincing, and removing `loading.tsx` is the smallest proven way to make dynamically rendered route content visible to non-JS readers. I would not restore the app-wide boundary or revive the app-wide `noscript` idea. I would, however, require the loading-state regression and the ratchet/measurement holes below to be fixed before merge. The PR currently claims a narrower UX cost and a stronger crawl guarantee than the patch actually establishes.

# TOP RISKS

## 1. The root deletion silently removes loading UI from non-crawlable routes too

`app/loading.tsx` was the only ancestor boundary for many routes that the test explicitly exempts from having their own file. `LOADING_LEGACY_ALLOWLIST` in `__tests__/app/route-file-structure.test.ts:96-119` includes `/admin`, `/admin/projects`, `/community/[id]/programs`, `/community/[id]/programs/[programId]/apply`, `/funding-map/add-program`, `/funding-map/manage-programs`, `/old-home`, `/seeds/fund`, `/stats`, and `/super-admin`. With the root file deleted, those physical segment chains contain no `loading.tsx` at all (the project profile routes have an explicit inner Suspense fallback, but not a route transition boundary).

That contradicts the PR statement that non-crawlable/admin/manage surfaces keep instant loading states. It also means the new test blesses the regression: the legacy allowlist checks only for a missing sibling (`route-file-structure.test.ts:231-241`), not whether an effective ancestor boundary remains. On direct loads users get no response body until render completion; on client navigation the old page appears inert. Backfill route-local/group-local boundaries below crawlable ancestors and add an assertion that every non-crawlable route has an effective boundary, not merely a sibling exemption.

## 2. “75/76 ok” does not prove the substantive page content is in the visible shell

The test only bans special-file boundaries. Explicit Suspense remains. Most importantly, `app/project/[projectId]/(profile)/layout.tsx:38-41` wraps `ProjectProfileLayout` **and all page children** in `<Suspense>`, while the only guaranteed project-specific element outside it is the parent layout's screen-reader-only h1 (`app/project/[projectId]/layout.tsx:153-165`). `/funding-map` similarly leaves the actual opportunity list inside Suspense (`app/funding-map/page.tsx:69-74`); only the heading was hoisted.

The crawler calls a page meaningful when visible text is at least 200 characters and any h1 exists (`scripts/indexability/crawl-sitemap.mjs:28-29,304-321`). Navbar/footer copy plus the hoisted/sr-only h1 can therefore classify `ok` even if the activity feed, project facts, or funding cards are hidden. The PR's tag-stack check is reported only for h1s, reinforcing this gap. Next/React may serialize a suspending explicit boundary with the same hidden late-chunk mechanism the PR correctly describes for `loading.tsx`; deleting the file does not override inner Suspense.

Require semantic, route-class-specific visible-shell assertions: project title plus a feed/project fact; program title/description; community project/program names; funding-map card names. The no-JS report should record these anchors, not just h1 and aggregate chrome-heavy character count.

## 3. The ratchet is a manually duplicated list, not a sitemap invariant

`SITEMAP_NO_LOADING` is hand-maintained in `route-file-structure.test.ts:39-94`; the actual static URLs are separately hand-maintained in `app/sitemaps/static/sitemap.ts:7-57`, community subpages in `app/sitemaps/communities/sitemap.ts:25-44`, and entity kinds in `utilities/sitemap.ts:73-90`. No test proves that every emitted sitemap route maps to the no-loading set.

The current test catches re-adding a boundary to an **existing set member**, but a developer can add an existing loading-backed page to `staticPages` (or add a new route with the normally required `loading.tsx`) and every structural test still passes. That is exactly the future regression the ratchet claims to prevent. Put crawlable route templates in one shared manifest consumed by sitemap generation and the structural test, or add an exhaustive sitemap-to-App-Router mapping test. Keep the documented `/nonprofit-research` exception explicit in that same source.

## 4. The entity UX and outage cost is real and not preview-measured

The table shows 543-986 ms warm local TTFB for community/program routes and 872 ms for two project slugs. Calling production “substantially lower” is an inference: there are no Vercel-preview p50/p95/cold-start measurements, only 3-5 post-warmup means against staging. There is also no client-navigation responsiveness measurement or user-visible progress metric.

The outage comparison is only server-work-equivalent, not UX-equivalent. Previously a browser received a skeleton before an indexer stall; now a direct navigation is blank and a client navigation holds the old page until the same work fails. `app/community/[communityId]/(with-header)/layout.tsx:21` awaits community data without a local fallback, and project metadata blocks on `getProjectCachedData` (`app/project/[projectId]/layout.tsx:24-47`). A non-JS fetcher that waits for the complete body still times out in either design, so this PR does not improve outage resilience; it improves the healthy path.

Before merge, measure preview p50/p95 and cold/warm TTFB for a representative distribution, then inject slow/failing indexer responses and verify timeout/error behavior. Set a product-owned threshold. If it fails, parallelize/deduplicate the entity layout and metadata fetches; do not restore a crawl-hiding boundary.

## 5. The evidence wording is stronger than the experiment

“Content-complete time is unchanged” should be “no material regression observed in this small local sample”; no variance or confidence interval is supplied. The after crawl covers 76 URLs, with only 15 project and 10 program samples, while the motivating production claim covers 204 URLs. It demonstrates the mechanism across sampled classes, not production-wide closure. “Raw-mode equivalence implied” is not verification, especially because this patch changes funding-map markup and deletes duplicated find-funders content.

# WHAT I WOULD DO DIFFERENTLY

1. Keep the no-loading architecture for sitemap routes, but first restore targeted loading boundaries for every non-crawlable route orphaned by deleting `app/loading.tsx`. Where a parent also owns crawlable descendants (for example `programs`), use a URL-neutral route group around only the non-crawlable page rather than placing a boundary above the crawlable detail route.
2. Replace the three-way route-list duplication with a single crawlability manifest or an exhaustive emitted-sitemap-to-route-pattern contract test.
3. Extend no-JS acceptance from “200 chars + h1” to semantic anchors per route class and explicitly audit inner `<Suspense>` boundaries. Hoist a compact server-rendered project summary/funding directory outside those boundaries where needed.
4. Gate merge on preview distributions and failure injection. Then optimize sequential entity fetches/caching if the agreed TTFB or navigation threshold is missed.
5. Remove the find-funders `noscript` hero as proposed. Once the real hero/FAQ is verified in the visible shell, retaining the replica creates duplicate text and two sources of presentation. The removed `x-pathname` proxy header has no remaining purpose.

# NITS

- `route-file-structure.test.ts:7-17` still describes “two allowlists” and a universal page/loading/error trio, although the file now has a third crawlable-route set and an explicit exception.
- Record the exact commands/URLs and raw before/after reports as artifacts. A prose table is not independently auditable.
- Separate “h1 outside hidden ancestry” from “substantive content outside hidden ancestry” in the PR body; they are not equivalent.

# RIVAL APPROACH ANALYSIS

**I would ship the author's no-loading/block-render architecture today, with the loading-state backfill and stronger crawlability tests required above.** It is the only option here that gives every requester the same host-correct, complete HTML through normal Next rendering, has no second content system, and is already empirically proven against the failure mode. Its cost is explicit and tunable—entity TTFB and delayed navigation feedback—whereas the rivals move correctness into routing, cache invalidation, UA classification, or HTML-rewriting machinery. I would reduce that cost with request deduplication and public-data caching, then revisit Cache Components as a deliberate platform migration.

**Cache Components/PPR:** the PR body should not call this experimental on Next 16.2. The installed 16.2.6 types expose top-level `cacheComponents`, deprecate `experimental.ppr`, and the official [Cache Components documentation](https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheComponents) says the unified flag was introduced in 16.0; PPR is the default rendering model when it is enabled. It is the best long-term architecture: cache public project/community/program content into a real static shell and stream only personalized controls. It is not a safe DEV-612-sized change here. `app/layout.tsx` calls `getWhitelabelContext()`, which reads `headers()` in `utilities/whitelabel-server.ts:28-31`; under Cache Components, runtime-header readers must sit behind Suspense. Putting the host-dependent root subtree there recreates the exact fallback-visible/content-hidden result for no-JS readers. Making useful content prerenderable therefore first requires separating host-independent main-site content from tenant chrome, defining cache lifetimes/tags for all public entity data, and migrating unsupported/changed route-segment caching semantics across the app. The flag also enables React Activity navigation behavior globally. I would prototype that separately, require build-shell semantic assertions, and roll it out route-class by route-class—not flip it to close this incident.

**Per-host static route trees:** this can unlock PPR cleanly: proxy-rewrite the canonical host to a host-independent public tree and tenant hosts to a tenant tree, so the main shell never calls `headers()`. Architecturally sound, but it duplicates or relocates a very large route hierarchy and forces every page, canonical, metadata path, error boundary, internal link, auth rule, and sitemap mapping through two routing topologies. It is justified only as a broader tenancy decomposition, not as an AEO hotfix. A smaller main-site-only crawlable tree is plausible, but it still creates two implementations of the same URL surface unless shared page bodies and exhaustive parity tests are built first.

**Bot-UA-conditional blocking:** I would reject it. `loading.tsx` is structural, not request-conditional, so implementing this means duplicate route trees, response transformation, or a proxy branch. AI fetchers have incomplete, mutable, or absent UA strings; CDN keys must vary correctly; users and bots receive materially different rendering; and failures look like cloaking. Next's `htmlLimitedBots` option applies to streaming metadata, not page-body Suspense, so it is not a hidden one-line fix. The all-request blocking behavior is slower but truthful and deterministic.

**Edge-cached HTML snapshots:** I would use framework-owned caching of public data/pages, not an independent HTML snapshot layer. A snapshot can give excellent TTFB after warmup, but this app has host variants, dynamic entities, query-state rules, RSC navigation payloads, build-specific client references, and frequent updates. Correct cache keys, atomic HTML/RSC invalidation, purge-on-mutation, cold misses, 404/redirect changes, and tenant isolation become a second rendering platform. If implemented inside Cache Components/ISR it becomes the desirable long-term option; if implemented as an opaque edge response cache, it is too easy to serve stale or hydration-incompatible documents.

**Prerender-service sidecar:** reject. A browser service that executes the stream and stores or returns the post-swap DOM adds another deployable, queue/crawl scheduling, browser patching, capacity planning, cache invalidation, and an outage mode precisely on crawler traffic. On-demand rendering worsens bot latency; scheduled rendering is stale between crawls; both must reproduce redirects, canonicals, host variants, and authentication exclusions. It also serves bots a separately generated representation. This is an escape hatch for an unmodifiable SPA, not for a Next server that already knows how to emit the HTML.

**`noscript` replicas:** reject beyond a temporary, single-page incident patch. They work for known static copy and saved find-funders, but entity pages would require duplicating live project/program/community renderers into a second tree. Drift, inconsistent links/structured data, doubled text in text-mode tools, and per-route proxy tagging scale linearly with the sitemap. Once the real page is visible, removing the existing replica is correct.

**Other variants:** `force-static`, route-group hoisting, and “meaningful” loading fallbacks are already disproved by the serialized fallback/hidden-chunk behavior. A Pages Router or custom blocking SSR tree would recover full HTML but is a larger parallel-router migration. Buffering and rewriting React's streamed HTML at the proxy to apply `$RC` server-side is worse: it couples production correctness to private React/Next markup and script protocols, defeats streaming anyway, complicates CSP, and can corrupt hydration. Build-time exporting every entity has the same host, freshness, invalidation, and scale problems as snapshots.

**Bottom line:** ship blocking SSR for sitemap routes now because it is boring, uniform, and correct under healthy upstreams; restore local skeletons to the non-crawlable routes accidentally orphaned by the root deletion; make semantic visible-shell tests and preview TTFB/error budgets merge gates. In parallel, design a host-separated Cache Components architecture in which canonical public content is explicitly cached into the prerendered shell and tenant/personalized UI is the only streamed portion. That is the architecture I would aim to replace block rendering with, but I would not make DEV-612 depend on completing it.

amaurymagalhaes:
## Parked by reviewer decision (2026-08-04)

This PR stays a reviewed draft, deliberately unmerged. The work is complete and rival-audited (no-js crawl 75/76 ok, 15,138 tests green, regression-guarded by the SITEMAP_NO_LOADING ratchet), but two considerations argue for waiting:

1. **Core Web Vitals risk on slow connections**: block-rendering delays resource discovery (head flush) behind server work; lab measurements cannot capture the field LCP effect on slow mobile — only CrUX can.
2. **The trade may be unnecessary**: Googlebot executes JavaScript, so the hidden-Suspense pattern may not cost anything with the crawlers that matter. Agent Analytics (live since 2026-08-03) will show which AI crawlers fetch without rendering.

**Revisit conditions (either unparks this):**
- Agent Analytics / Profound evidence that non-rendering crawlers materially fetch our pages and the experiments underperform on citation because of it (check at the 2026-08-18 evaluation checkpoint), or
- the cacheComponents/PPR migration becomes viable on stable Next for a host-branched app, which delivers streaming UX and crawler-visible content with no trade (the SITEMAP_NO_LOADING route set in this PR is the migration manifest).

Until then the per-route noscript pattern (retired here, still on main) remains the targeted mitigation for any individual page that evidence says needs it.

amaurymagalhaes:
## Babysit sign-off

**CI: all green — 37 checks passing, 0 failures.** QA verdict: `PASS — no PR-caused blocking issues (1 pre-existing)`.

**Self-review score: 97/100.** The review caught a real bug I had introduced: `ProjectProfileLayout` returns different root element types per branch, so React remounted the extracted invite-code watcher when the project resolved, resetting its open-once latch and re-opening a modal the user had already closed. Latch moved to the layout; regression test is mutation-verified.

**Review threads: 6 resolved** (1 Major + 4 Minor from an earlier pass, plus 1 Major CodeRabbit found in this PR's own new code — the server feed was gated on item count, so a filtered query that legitimately returned zero results kept the unfiltered feed on screen).

### QA findings fixed

| Scenario | Issue | Fix |
|---|---|---|
| P2 | `/projects` streamed its entire list into a hidden chunk | Boundary removed; 0 hidden chunks, 25,382 chars of no-JS text, 50 crawlable links |
| P8 | `/for-agents` hid its tool catalog the same way | Same fix |
| A1/A2 | Updates failures were invisible | See below |

### The updates feed, in full

Five commits, each exposing the next layer. Worth reading in order before merge — this is the riskiest part of the diff:

1. Error banner added — **dead code**: the service swallowed every error, so React Query never saw a rejection and `isUpdatesError` could not become true.
2. `hasUpdatesData` — gate on query settlement, not item count.
3. `defaultQueryOptions` — the query wasn't inheriting the app retry policy, so failures took ~7s of silent retries to surface.
4. **Service rethrow** — the actual root cause. Non-404 failures now reach React Query; 404 still resolves to empty. This also fixes a latent bug in `useUpdateActions`, which read a failed fetch as "the update is gone".
5. **Request timeout** — a hung request neither resolves nor rejects. #4 made this reachable by removing the empty-response fallback, so the request is now bounded at 15s. Closes #2007.

The recurring cause of 1–3: the tests mocked the very thing whose behaviour they asserted. A test that mocks the service to reject cannot tell you whether the service rejects. New coverage uses a real `QueryClient` and exercises the service's own error path.

### Filed, not fixed here — all pre-existing and independent of DEV-612

- #2000 — project routes answer 404 with a zero-byte body. Four hypotheses implemented and disproven against real production builds; all recorded so nobody repeats them.
- #2001 — Funding/Impact tabs stuck on skeleton despite 200s. Classified `pr-caused` by QA; the attribution is very likely wrong — the old grants prefetch keyed on the slug while the tab reads by uid, so it never seeded anything and removing it cannot have caused a hang.
- #2003 — `/admin` hangs for authenticated non-staff users. Gating flags in `CommunityAdmin.tsx`, untouched by this PR.

### Pipeline change

`qa-execution` counted every Critical/High failure toward blocking, including bugs in code the PR never touched — which made a green pipeline unreachable by working on the PR. Dogfood already scored PR-caused only; `qa-execution` now matches. Fails safe: an unclassified failure counts as PR-caused and blocks. Pre-existing failures stay visible in this comment rather than being hidden.

This changes merge gating for every PR in the repo and is worth reviewing on its own merits — it's a clean two-file change if you'd rather it land separately.

### Not covered

- The `/project/[projectId]` contract is deliberate and documented in the description: identity in the visible HTML, tab body streamed. No-JS readers get ~1,018 chars of identity, not milestone prose. Revisitable against Agent Analytics at the Aug 18 checkpoint.
- Two QA scenarios (A3, A4) could not run — `agent-browser`'s request interception didn't fire against the cross-origin indexer. Tooling, not app behaviour, but it means the retry-clears-the-error path is unverified in a browser.

